PR #70, opened 2026-06-17
fix(dev): LOCAL_IMAGES=true image tagging and loading

squizzi:
Fixes three bugs in `make kind-up LOCAL_IMAGES=true`:

1. **Missing for loop** - `_kind-load-images` had the loop body but no `for` declaration, so images were never tagged or loaded
2. **Missing control-plane mapping** - kind-local kustomization only mapped api-server and ui, leaving control-plane on Quay.io
3. **Dead ConfigMap patch** - Removed broken code trying to patch non-existent `ambient-agent-registry` with undefined `$UPDATED` variable

Now all images (runner, api-server, control-plane, mcp, ui, credential sidecars) are properly tagged with `localhost/` and loaded into kind.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,7 +169,7 @@ help: ## Display this help message
 
 ##@ Building
 
-build-all: build-runner build-api-server build-control-plane build-mcp build-ambient-ui ## Build all container images
+build-all: build-runner build-api-server build-control-plane build-mcp build-ambient-ui build-credential-sidecars ## Build all container images
 
 build-ambient-ui: ## Build ambient-ui image
 	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Building ambient-ui with $(CONTAINER_ENGINE)..."
@@ -1243,13 +1243,14 @@ check-architecture: ## Validate build architecture matches host
 
 _kind-load-images: ## Internal: Load images into kind cluster
 	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Loading images into kind ($(KIND_CLUSTER_NAME))..."
-		echo "  Loading $(KIND_IMAGE_PREFIX)$$img..."; \
+	@for img in $(RUNNER_IMAGE) $(API_SERVER_IMAGE) $(CONTROL_PLANE_IMAGE) $(MCP_IMAGE) $(AMBIENT_UI_IMAGE) $(GITHUB_MCP_IMAGE) $(JIRA_MCP_IMAGE) $(K8S_MCP_IMAGE) $(GOOGLE_MCP_IMAGE); do \
+		echo "  Loading $$img -> $(KIND_IMAGE_PREFIX)$$img..."; \
+		$(CONTAINER_ENGINE) tag $$img $(KIND_IMAGE_PREFIX)$$img 2>/dev/null || true; \
 		if [ -n "$(KIND_HOST)" ] || [ "$(CONTAINER_ENGINE)" = "podman" ]; then \
 			$(CONTAINER_ENGINE) save $(KIND_IMAGE_PREFIX)$$img | \
 			$(CONTAINER_ENGINE) exec -i $(KIND_CLUSTER_NAME)-control-plane \
 			ctr --namespace=k8s.io images import -; \
 		else \
-			docker tag $$img $(KIND_IMAGE_PREFIX)$$img 2>/dev/null || true; \
 			kind load docker-image $(KIND_IMAGE_PREFIX)$$img --name $(KIND_CLUSTER_NAME); \
 		fi; \
 	done
--- a/Makefile
+++ b/Makefile
@@ -944,11 +944,11 @@ kind-port-forward: check-kubectl check-local-context ## Port-forward kind servic
 	@echo ""
 	@echo "$(COLOR_YELLOW)Press Ctrl+C to stop$(COLOR_RESET)"
 	@echo ""
-	@trap 'echo ""; echo "$(COLOR_GREEN)✓$(COLOR_RESET) Port forwarding stopped"; exit 0' INT; \
-	(kubectl port-forward -n ambient-code svc/ambient-ui-service $(KIND_FWD_FRONTEND_PORT):3000 >/dev/null 2>&1 &); \
-	(kubectl port-forward -n ambient-code svc/ambient-api-server $(KIND_FWD_BACKEND_PORT):8000 >/dev/null 2>&1 &); \
-	(kubectl port-forward -n ambient-code svc/ambient-ui-service $(KIND_FWD_AMBIENT_UI_PORT):3000 >/dev/null 2>&1 &); \
-	(kubectl port-forward -n ambient-code svc/keycloak-service $(KIND_FWD_KEYCLOAK_PORT):8080 >/dev/null 2>&1 &); \
+	@trap 'echo ""; echo "$(COLOR_GREEN)✓$(COLOR_RESET) Port forwarding stopped"; kill 0; exit 0' INT TERM; \
+	kubectl port-forward -n ambient-code svc/ambient-ui-service $(KIND_FWD_FRONTEND_PORT):3000 >/dev/null 2>&1 & \
+	kubectl port-forward -n ambient-code svc/ambient-api-server $(KIND_FWD_BACKEND_PORT):8000 >/dev/null 2>&1 & \
+	kubectl port-forward -n ambient-code svc/ambient-ui-service $(KIND_FWD_AMBIENT_UI_PORT):3000 >/dev/null 2>&1 & \
+	kubectl port-forward -n ambient-code svc/keycloak-service $(KIND_FWD_KEYCLOAK_PORT):8080 >/dev/null 2>&1 & \
 	wait
 
 dev-bootstrap: check-kubectl check-local-context ## Bootstrap developer workspace with API key and integrations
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -944,11 +944,11 @@ kind-port-forward: check-kubectl check-local-context ## Port-forward kind servic
...
 	@echo "$(COLOR_YELLOW)Press Ctrl+C to stop$(COLOR_RESET)"
 	@echo ""
-	@trap 'echo ""; echo "$(COLOR_GREEN)✓$(COLOR_RESET) Port forwarding stopped"; exit 0' INT; \
-	(kubectl port-forward -n ambient-code svc/ambient-ui-service $(KIND_FWD_FRONTEND_PORT):3000 >/dev/null 2>&1 &); \
-	(kubectl port-forward -n ambient-code svc/ambient-api-server $(KIND_FWD_BACKEND_PORT):8000 >/dev/null 2>&1 &); \
-	(kubectl port-forward -n ambient-code svc/ambient-ui-service $(KIND_FWD_AMBIENT_UI_PORT):3000 >/dev/null 2>&1 &); \
-	(kubectl port-forward -n ambient-code svc/keycloak-service $(KIND_FWD_KEYCLOAK_PORT):8080 >/dev/null 2>&1 &); \
+	@trap 'echo ""; echo "$(COLOR_GREEN)✓$(COLOR_RESET) Port forwarding stopped"; kill 0; exit 0' INT TERM; \
+	kubectl port-forward -n ambient-code svc/ambient-ui-service $(KIND_FWD_FRONTEND_PORT):3000 >/dev/null 2>&1 & \
+	kubectl port-forward -n ambient-code svc/ambient-api-server $(KIND_FWD_BACKEND_PORT):8000 >/dev/null 2>&1 & \
```
jsell-rh, line 949: `kill 0` sends SIGTERM to the entire process group — including the `make` process itself. In practice this works because the trap fires on exit anyway, but it can produce ugly "Terminated" messages or interfere if `make` is called from a wrapper script.

A safer pattern is to explicitly kill only the backgrounded children:

```makefile
@trap 'echo ""; echo "$(COLOR_GREEN)✓$(COLOR_RESET) Port forwarding stopped"; jobs -p | xargs -r kill; exit 0' INT TERM; \
```

Not blocking — this is a local-dev target and the current behavior is fine for interactive use. Just noting for robustness.

**Severity: Minor**

Makefile
```diff
@@ -1243,13 +1243,14 @@ check-architecture: ## Validate build architecture matches host
...
+		$(CONTAINER_ENGINE) tag $$img $(KIND_IMAGE_PREFIX)$$img 2>/dev/null || true; \
 		if [ -n "$(KIND_HOST)" ] || [ "$(CONTAINER_ENGINE)" = "podman" ]; then \
 			$(CONTAINER_ENGINE) save $(KIND_IMAGE_PREFIX)$$img | \
 			$(CONTAINER_ENGINE) exec -i $(KIND_CLUSTER_NAME)-control-plane \
 			ctr --namespace=k8s.io images import -; \
 		else \
-			docker tag $$img $(KIND_IMAGE_PREFIX)$$img 2>/dev/null || true; \
 			kind load docker-image $(KIND_IMAGE_PREFIX)$$img --name $(KIND_CLUSTER_NAME); \
 		fi; \
 	done
```
jsell-rh, line 1256: The `_kind-load-images` for loop tags and loads 9 images (runner, api-server, control-plane, mcp, ui, plus 4 credential sidecars), but the `kind-local` kustomization only remaps 3 of them (`acp_api_server`, `acp_ambient_ui`, and now `acp_control_plane`) to `localhost/` names.

That means pods for `acp_claude_runner`, `acp_mcp`, and the 4 `acp_credential_*` images will still have `imagePullPolicy: Always` (the base default) and reference the Quay.io names — so k8s will try to pull from Quay instead of using the locally-loaded `localhost/` images.

For full LOCAL_IMAGES=true parity you'd need:
1. Image name remaps in `kind-local/kustomization.yaml` for runner, mcp, and the credential sidecars.
2. `imagePullPolicy: IfNotPresent` patches on those deployments/jobs too.

I'd suggest either:
- **Option A:** Add the missing remaps now (this PR is the right place).
- **Option B:** Scope this PR to the 3 images that work today and file a follow-up issue for the rest — but document that LOCAL_IMAGES=true only covers api-server, ui, and control-plane.

**Severity: Major** — without the remaps, locally-built runner/mcp/sidecar images won't actually be used.

Makefile
```diff
@@ -169,7 +169,7 @@ help: ## Display this help message
 
 ##@ Building
 
-build-all: build-runner build-api-server build-control-plane build-mcp build-ambient-ui ## Build all container images
+build-all: build-runner build-api-server build-control-plane build-mcp build-ambient-ui build-credential-sidecars ## Build all container images
```
jsell-rh, line 172: Good addition — `build-credential-sidecars` was missing from `build-all`. Worth noting this changes the default `make build-all` for everyone, not just `LOCAL_IMAGES=true`. That's the correct behavior (they should have been here all along), just calling it out since it affects CI build times if `build-all` is used in pipelines.

**Severity: Info**

components/ambient-ui/src/lib/oidc.ts
```diff
@@ -35,27 +35,29 @@ async function getOIDCConfig(): Promise<client.Configuration> {
...
-  // userinfo, jwks) pointing to the backend issuer.
-  if (env.SSO_FRONTEND_ISSUER_URL && env.SSO_ISSUER_URL) {
-    const backendIssuer = env.SSO_ISSUER_URL
-    const frontendIssuer = env.SSO_FRONTEND_ISSUER_URL
+  // Keycloak returns browser-facing URLs using KC_HOSTNAME (e.g. http://localhost/sso).
+  // When port-forwarding the service for local development, the browser needs to reach Keycloak on
+  // a port-forwarded URL instead of the internal service URL.
+  // Rewrite the discovered issuer prefix to SSO_FRONTEND_ISSUER_URL for browser endpoints.
+  if (env.SSO_FRONTEND_ISSUER_URL) {
     const metadataObj = metadata as Record<string, unknown>
```
jsell-rh, line 43: Nice improvement — deriving the rewrite source from the discovered `issuer` field instead of requiring `SSO_ISSUER_URL` makes this self-contained. The old code silently skipped rewriting if `SSO_ISSUER_URL` was unset even when `SSO_FRONTEND_ISSUER_URL` was set, which was the root cause of broken redirects in port-forwarded setups.

One edge case to consider: `discoveredIssuer` could have a trailing slash while `frontendIssuer` might not (or vice versa). `String.replace` will only match exact prefixes. In practice Keycloak's discovery output is consistent, so this is unlikely to bite — just flagging.

**Severity: Info**

scripts/setup-kind-sso.sh
```diff
@@ -30,5 +30,15 @@ kubectl patch secret sso-credentials -n "$NAMESPACE" --type=json -p="[
...
-# Keycloak's KC_HOSTNAME is already set correctly in the manifest
-# We don't need to patch it - the dual-issuer pattern handles browser vs pod URLs
+# Patch KC_HOSTNAME so Keycloak generates correct URLs for port-forwarded access.
+# Only patch if the value changed to avoid unnecessary restarts.
+DESIRED_KC_HOSTNAME="http://localhost:${KIND_FWD_KEYCLOAK_PORT}"
+CURRENT_KC_HOSTNAME=$(kubectl get deployment/keycloak -n "$NAMESPACE" \
+  -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="KC_HOSTNAME")].value}' 2>/dev/null || true)
+
+if [ "$CURRENT_KC_HOSTNAME" != "$DESIRED_KC_HOSTNAME" ]; then
+  kubectl set env deployment/keycloak -n "$NAMESPACE" \
```
jsell-rh, line 40: Good idempotency pattern — only patching and restarting Keycloak when the value actually changed.

The `rollout status` timeout of 120s is reasonable. One thing: if the `kubectl set env` succeeds but Keycloak fails to come ready (e.g. bad config), the script exits with a non-zero status which will halt `kind-port-forward`. That's actually the right behavior — fail loudly rather than forwarding to a broken Keycloak.

**Severity: Info** — looks correct.

jsell-rh (commented):
## Amber Review — PR #70: fix(dev): LOCAL_IMAGES=true image tagging and loading

**Overall: Good fix for a genuinely broken workflow.** The three bugs identified (missing `for` loop, missing control-plane image remap, dead ConfigMap patch) are real and the fixes are correct. The OIDC rewrite simplification and KC_HOSTNAME patching are solid improvements to the local SSO flow.

### Summary

| Severity | Count | Description |
|----------|-------|-------------|
| Major | 1 | Incomplete image remapping — 6 of 9 images loaded into kind won't be used by pods |
| Minor | 1 | `kill 0` in port-forward trap is aggressive (kills process group) |
| Info | 3 | build-all scope change, OIDC trailing-slash edge case, SSO script correctness |

### Key Finding

The `_kind-load-images` target now correctly tags and loads all 9 images into kind, but the `kind-local` kustomization only remaps 3 images (`api-server`, `ui`, `control-plane`) from Quay.io names to `localhost/` names. The remaining 6 images (runner, mcp, 4 credential sidecars) will still reference Quay.io in their pod specs, so Kubernetes won't use the locally-loaded versions. This should either be completed in this PR or explicitly scoped out with a follow-up issue.

### CI Status
- Build, lint, unit tests, local-dev simulation: all passing ✅
- Konflux (`acp-ui-main-on-pull-request`): failed ❌ — appears to be an infrastructure/pipeline issue, not related to this PR's changes

### Commits
3 commits — should be squashed before merge per project conventions.

### Constitution Compliance
- ✅ No production code changes (dev tooling only)
- ✅ No security-sensitive changes
- ✅ Conventional commit format
- ✅ Changes are isolated and low-risk

— Amber 🤖

Makefile
squizzi: kill 0 is correct for this use case - it kills the process group which includes all the backgrounded kubectl port-forward processes. The alternative would be to track each PID and kill them individually, which is more complex.  This is a `Makefile` target, I'm not too concerned with some ugly `Temrinated` messages.

PR #122, opened 2026-06-22
fix(dev): update vertex setup and fix RBAC for kind clusters

squizzi:
## Summary

- **Vertex setup script**: Fixed `setup-vertex-kind.sh` to create the `operator-config` ConfigMap (instead of patching one that doesn't exist), use the correct deployment name (`ambient-control-plane`), and accept `AMBIENT_UI_URL` for the frontend URL
- **RBAC fix**: Gate OpenShift-specific RBAC rules (`build.openshift.io`, `image.openshift.io`, `route.openshift.io`) on `PLATFORM_MODE=mpp` so they're excluded on kind, fixing a "forbidden" error when creating sessions
- **Makefile improvements**: Add `kind-setup-vertex` target, make `kind-rebuild` fail fast before building images if cluster doesn't exist, fix `go-vet` pre-commit hook for bash 3.2 compatibility

## Test plan

- [ ] Run `make kind-up` and verify cluster comes up
- [ ] Run `make kind-setup-vertex` with Vertex env vars set and verify ConfigMap and secret are created
- [ ] Create a session on kind and confirm no RBAC errors in control plane logs
- [ ] Run `make kind-rebuild` without a cluster and verify it fails immediately (before building)
- [ ] Run `go test ./internal/reconciler/` in the control plane to verify RBAC tests pass for both standard and mpp modes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 .PHONY: local-dev-token
 .PHONY: local-logs local-logs-api-server local-logs-ui local-logs-control-plane local-shell-api-server local-shell-ui
 .PHONY: local-test local-test-dev local-test-quick test-all local-troubleshoot local-port-forward local-stop-port-forward
-.PHONY: push-all registry-login setup-hooks remove-hooks lint check-minikube check-kind check-kubectl check-local-context dev-bootstrap kind-rebuild kind-reload-ambient-ui kind-status kind-login kind-sso-toggle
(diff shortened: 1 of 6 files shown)


jsell-rh:
@squizzi Thanks! One minor note for future, maybe we change platform mode to be named `openshift` since those rbac rules would apply to more environments than just `mpp` (unless I'm missing something!)

squizzi:
ya, I agree, `mpp` is ambiguous

PR #143, opened 2026-06-23
spec: OpenShell sandbox provisioning

bsquizz:
## Summary

- Adds a new spec defining desired state for delegating agent pod creation to OpenShell gateways
- Covers gateway discovery via K8s Service DNS, sandbox lifecycle (create/deprovision/cleanup), credential-to-provider mapping, status syncing, and configuration
- When `OpenShellEnabled=false` (default), all existing behavior is preserved unchanged

## Spec highlights

- **Gateway-based sandbox creation**: `CreateSandbox` gRPC RPC replaces direct pod creation when enabled
- **Credential mapping**: Ambient credentials map to OpenShell providers (github→github, anthropic→claude, others→generic)
- **Status syncing**: Sandbox disappearance signals session completion; ERROR phase maps to Failed
- **No fallback**: When enabled, gateway is the only path — no silent fallback to insecure direct pods
- **Proto vendoring**: OpenShell proto files vendored with buf v2 code generation (matching ambient-api-server pattern)

## Test plan

- [ ] Review spec requirements and scenarios for completeness
- [ ] Verify backward compatibility section covers all existing consumers
- [ ] Confirm provider type mapping table matches OpenShell's supported types


specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
+# OpenShell Sandbox Provisioning Specification
+
+**Date:** 2026-06-23
+**Status:** Draft — proposed addition to control plane provisioning
+
+---
+
```
jsell-rh, line 9: **Blocker — Architectural conflict with existing spec**

This spec introduces a **gateway-based** OpenShell model (`CreateSandbox` gRPC to a per-namespace gateway), but the existing implemented spec at `specs/security/openshell-sandbox.spec.md` explicitly states:

> "The Supervisor SHALL operate in file mode using local policy files. The system **SHALL NOT** require an OpenShell Gateway service." (lines 44-45)

The current codebase (`OPENSHELL_ENABLED=true`) already implements the file-mode approach: the control plane copies a policy ConfigMap to the runner namespace, mounts it at `/etc/openshell`, and the Supervisor binary runs locally inside the pod with elevated capabilities (NET_ADMIN, SYS_ADMIN, SYS_PTRACE, etc.).

This spec proposes **replacing** that with gateway-delegated sandbox creation, but:
1. It reuses the same `OPENSHELL_ENABLED` flag without addressing the existing behavior
2. It doesn't reference or supersede the existing security spec
3. The migration table says `ensurePod()` is "preserved unchanged" for the disabled path, but doesn't explain what happens to the existing OpenShell-enabled pod creation logic (elevated SecurityContext, policy ConfigMap propagation, etc.)

**Recommendation:** This spec needs a clear statement about the relationship to `specs/security/openshell-sandbox.spec.md`. Is this replacing file-mode? Running alongside it? A new flag (`OPENSHELL_GATEWAY_ENABLED`) separate from `OPENSHELL_ENABLED`? The answer shapes the entire implementation.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+#### Scenario: Default discovery
+
+- GIVEN no override configuration is set
+- WHEN the control plane needs to reach the gateway in namespace `my-project`
+- THEN it SHALL connect to `openshell-gateway.my-project.svc.cluster.local:8080`
+
+#### Scenario: Custom service name
+
+- GIVEN `OPENSHELL_GATEWAY_SERVICE` is set to `my-gateway`
+- AND `OPENSHELL_GATEWAY_PORT` is set to `9090`
```
jsell-rh: **Critical — Default service name inconsistency**

The "Default discovery" scenario says the control plane connects to `openshell-gateway.my-project.svc.cluster.local:8080`, implying the default service name is `openshell-gateway`.

But the Configuration table at the bottom of this spec says `OPENSHELL_GATEWAY_SERVICE` defaults to `openshell` (not `openshell-gateway`).

These need to match — either update the scenario to `openshell.my-project.svc.cluster.local:8080` or change the default in the config table to `openshell-gateway`.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+|---|---|---|
+| `OPENSHELL_ENABLED` | `false` | Enable gateway-based sandbox provisioning |
+| `OPENSHELL_GATEWAY_SERVICE` | `openshell` | Kubernetes Service name of the gateway |
+| `OPENSHELL_GATEWAY_PORT` | `8080` | Port of the gateway gRPC service |
+
+---
+
+## Migration
+
+### Existing consumers
```
jsell-rh, line 377: **Critical — `OPENSHELL_ENABLED` flag collision**

The existing codebase already uses `OPENSHELL_ENABLED` to control file-mode sandbox behavior (`config.go:95`). Today when `OPENSHELL_ENABLED=true`:
- The reconciler copies an `openshell-policy` ConfigMap to the runner namespace
- Mounts it at `/etc/openshell` in the pod
- Sets elevated SecurityContext (root, SYS_ADMIN, NET_ADMIN, etc.)
- Sets env vars `OPENSHELL_POLICY_RULES`, `OPENSHELL_POLICY_DATA`, `OPENSHELL_LOG_LEVEL`

This spec proposes adding `OPENSHELL_ENABLED` (same flag) to mean "use gateway mode." The config table also introduces two *new* env vars (`OPENSHELL_GATEWAY_SERVICE`, `OPENSHELL_GATEWAY_PORT`) that don't exist today, but `config.go` would need to differentiate between file-mode and gateway-mode.

Consider a separate flag (e.g. `OPENSHELL_MODE=file|gateway`) or `OPENSHELL_GATEWAY_ENABLED` to avoid breaking the existing feature.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+| Consumer | Impact |
+|---|---|
+| `kube_reconciler.go` `ensurePod()` | Preserved unchanged for `OpenShellEnabled=false` path |
+| `kube_reconciler.go` credential sidecars | Replaced by OpenShell providers when gateway mode is active |
+| `pod_sync.go` | Extended with sandbox sync branch |
+| `main.go` | Extended to create and wire `GatewayClient` |
+| `config.go` | Extended with 2 new fields |
+| Runner pod | No changes — same image, same env vars, running inside OpenShell sandbox instead of bare pod |
+
+### Backward compatibility
```
jsell-rh, line 391: **Major — Missing SecurityContext specification**

The existing OpenShell implementation requires an elevated SecurityContext (root, `SYS_ADMIN`, `NET_ADMIN`, `SYS_PTRACE`, `SETUID`, `SETGID`, `CHOWN`, `DAC_OVERRIDE`, seccomp `Unconfined`) — see `kube_reconciler.go:676-716` and `specs/security/openshell-sandbox.spec.md` § Container Security Context.

In gateway mode, the runner still runs "inside OpenShell sandbox instead of bare pod" (per the migration table). Does the runner pod still need these elevated privileges, or does the gateway handle isolation externally? This is a security-critical question that the spec should answer explicitly.

If gateway mode means the pod runs with *standard* restricted SecurityContext (no root, drop ALL, no privilege escalation), that's a significant security improvement worth calling out. If it still needs elevated privileges, explain why.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+#### Scenario: Sandbox naming
+
+- GIVEN a session with ID `abc123`
+- WHEN a sandbox is created for this session
+- THEN the sandbox name SHALL be `session-<safe_name>` where `<safe_name>` is the session ID truncated to 40 characters and lowercased
+- AND the sandbox SHALL carry labels `ambient-code.io/session-id`, `ambient-code.io/project-id`, `ambient-code.io/managed=true`, and `ambient-code.io/managed-by=ambient-control-plane`
+
+#### Scenario: Idempotent creation
+
+- GIVEN a sandbox already exists for a session
```
jsell-rh, line 102: **Major — Sandbox naming: truncation can cause collisions**

Session IDs are UUIDs (36 chars). Truncating to 40 characters means the full UUID is preserved (good), but the spec doesn't address what happens if session IDs aren't UUIDs in the future or if `safe_name` transformation introduces collisions (e.g., two IDs that differ only after character 40).

More importantly, the spec says "truncated to 40 characters and lowercased" — but UUIDs are already lowercase hex + hyphens, so the lowercasing is a no-op for current IDs. If this is forward-looking, document why. If it's just defensive, that's fine but say so.

Minor nit: the existing codebase uses the pattern `session-<safe_resource_name>` at `kube_reconciler.go:1282` via a `safeResourceName()` helper that does more than truncate (it also sanitizes DNS-invalid characters). The spec should reference this existing helper or specify the full sanitization rules to stay consistent.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+| Sandbox not found | `Completed` |
+
+#### Scenario: Sandbox disappearance indicates completion
+
+- GIVEN a running session with an active sandbox
+- WHEN the status syncer calls `GetSandbox` and receives a not-found response
+- THEN the session phase SHALL transition to `Completed`
+
+#### Scenario: Gateway unreachable during sync
+
```
jsell-rh, line 280: **Major — "Sandbox not found → Completed" is risky**

Mapping `GetSandbox` not-found to `Completed` assumes the only reason a sandbox disappears is successful completion. But sandboxes could also disappear due to:
- Gateway restart/redeployment losing state
- Namespace-level resource cleanup (e.g., resource quotas, admin intervention)
- OpenShell gateway bug or crash

The existing `pod_sync.go` handles pod disappearance more carefully — it checks for `Succeeded` vs `Failed` pod phase, crash loop detection (restart count ≥ 5, OOMKilled, etc.).

Consider requiring the status syncer to check the sandbox's exit status or last-known phase before defaulting to `Completed`. At minimum, log a warning when a sandbox disappears unexpectedly so operators can distinguish clean exits from silent failures.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+- AND each file SHALL have a `go_package` option added
+- AND generated Go stubs SHALL be output to `pkg/openshell/grpc/`
+
+### Requirement: gRPC Connection Management
+
+The control plane SHALL maintain a cache of gRPC connections to OpenShell gateways, one per namespace, with lazy initialization.
+
+#### Scenario: Connection caching
+
+- GIVEN multiple sessions in the same project namespace
```
jsell-rh, line 316: **Major — Proto output path doesn't follow existing conventions**

The spec says generated Go stubs go to `pkg/openshell/grpc/`. However, the control plane component doesn't currently have a top-level `pkg/` directory — its Go packages are under `components/ambient-control-plane/internal/` and `components/ambient-control-plane/cmd/`.

A `pkg/` directory at project root would be accessible to all components, which may or may not be intentional. If only the control plane needs these stubs, they should live under `components/ambient-control-plane/`. If shared, explain which other components will consume them.

Also verify: the PR description says "matching ambient-api-server pattern" for buf vendoring. Does the API server use a `pkg/` directory for its generated protos? If so, reference that for consistency.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+- WHEN the control plane creates sandboxes for each
+- THEN it SHALL reuse a single gRPC connection per namespace
+- AND connections SHALL be created lazily on first use
+
+#### Scenario: Shutdown cleanup
+
+- WHEN the control plane shuts down
+- THEN it SHALL close all cached gRPC connections
+
+### Requirement: Configuration
```
jsell-rh, line 345: **Minor — Missing gRPC connection health/retry specification**

The spec covers lazy init and shutdown cleanup for gRPC connections, but doesn't specify:
- What happens when a cached connection becomes unhealthy (gateway pod restarts)?
- Reconnection backoff strategy
- Connection timeout for initial dial

The existing control plane has reconnection patterns for its API server gRPC stream. The spec should reference those patterns or specify its own retry semantics — especially since sandbox creation is latency-sensitive.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+#### Scenario: Environment variable translation
+
+- GIVEN a session with LLM model, prompt, repo URL, and proxy settings
+- WHEN the sandbox is created
+- THEN all environment variables from `buildEnv()` that have literal string values SHALL be included
+- AND Kubernetes-specific `valueFrom` / `fieldRef` entries (e.g., `POD_IP`) SHALL be omitted
+
+### Requirement: Sandbox Deprovisioning
+
+When a session is stopped or deleted, the control plane SHALL delete the sandbox and its associated providers via the OpenShell gateway.
```
jsell-rh: **Minor — Credential lifecycle: what about credential rotation?**

The spec covers creating and deleting providers, but doesn't address credential rotation. If an ambient credential is rotated mid-session:
- Are providers updated in-place?
- Is the sandbox recreated?
- Does the OpenShell egress proxy pick up new credentials automatically?

This may be out of scope for v1, but worth noting as a known gap.

— Amber

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,241 @@
...
+- GIVEN the control plane configuration
+- THEN the following environment variables SHALL be supported:
+
+| Variable | Default | Purpose |
+|---|---|---|
+| `OPENSHELL_ENABLED` | `false` | Enable gateway-based sandbox provisioning |
+| `OPENSHELL_GATEWAY_SERVICE` | `openshell` | Kubernetes Service name of the gateway |
+| `OPENSHELL_GATEWAY_PORT` | `8080` | Port of the gateway gRPC service |
+
+---
```
jsell-rh, line 373: **Minor — Missing feature flag gate**

Per `CLAUDE.md`: "Feature flags strongly recommended: Gate new features behind Unleash flags."

A feature of this magnitude (replacing the entire pod provisioning path) should be behind an Unleash feature flag in addition to the env var, enabling gradual rollout and kill-switch capability without redeploying.

— Amber

jsell-rh (changes requested):
## Amber Review — OpenShell Sandbox Provisioning Spec

**Verdict: Changes Requested**

Well-structured spec with clear GIVEN/WHEN/THEN scenarios and a solid backward-compatibility story. The gateway-based approach is a sensible evolution for managed sandboxing. However, there's a blocking architectural conflict that needs resolution before this can land.

### Blockers (1)

- **Architectural conflict with existing `specs/security/openshell-sandbox.spec.md`** — The existing implemented spec explicitly states "The system SHALL NOT require an OpenShell Gateway service" and uses file-mode operation. This spec introduces a gateway dependency under the same `OPENSHELL_ENABLED` flag. The relationship between these two specs must be clarified: is this replacing, extending, or coexisting with file-mode?

### Critical (2)

- **`OPENSHELL_ENABLED` flag collision** — Same env var, different semantics. Today it means "use Supervisor binary in-pod with elevated privileges." This spec redefines it to mean "delegate to gateway." Needs a new flag or a mode selector.
- **Default service name inconsistency** — Discovery scenario says `openshell-gateway`, config table says `openshell`.

### Major (4)

- **Missing SecurityContext specification** — Does gateway mode still require root + elevated capabilities, or can pods run restricted?
- **"Sandbox not found → Completed" is risky** — Silent failures would be indistinguishable from clean exits.
- **Proto output path** — `pkg/openshell/grpc/` doesn't match existing component structure.
- **Sandbox naming** — Should reference existing `safeResourceName()` helper for consistency.

### Minor (3)

- gRPC connection health/retry semantics unspecified
- Credential rotation during active sessions not addressed
- Feature flag gate recommended per CLAUDE.md conventions

### What's Good

- Clear no-fallback policy (gateway unreachable → fail, don't silently fall back to direct pods) — this is the right security posture
- Idempotent creation via `GetSandbox` check
- Provider type mapping table is comprehensive
- Migration table identifying all affected files is helpful for implementation planning
- Backward compatibility section is thorough

**Confidence:** High — findings are verified against current codebase and existing specs.

— Amber 🤖

jsell-rh (commented):
## Amber Review — Revised

Thanks for the context that this is an intentional architectural pivot from file-mode to gateway-based sandboxing. That reframes my earlier blocker. Updating my assessment:

### Previous Blocker → Downgraded to Major (Spec Consistency)

The architectural direction is sound — gateway-delegated sandboxing is a cleaner separation of concerns than in-pod Supervisor with elevated privileges. The remaining issue is **spec hygiene**: `specs/security/openshell-sandbox.spec.md` (Status: Implemented) still says "SHALL NOT require an OpenShell Gateway service." Since this PR is proposing the pivot, it should either:
1. Update the existing security spec's Status to `Superseded by specs/platform/openshell-sandbox-provisioning.spec.md` in a companion commit, or
2. Add a "Supersedes" section to *this* spec stating it replaces the file-mode approach from the security spec

This prevents future implementors from encountering contradictory SHALLs. Can land in the same PR or as immediate follow-up — not a merge blocker if there's a tracking issue.

### Remaining Critical (1) — Still stands

- **Default service name inconsistency** — The discovery scenario says `openshell-gateway.my-project.svc.cluster.local:8080` but the config table defaults `OPENSHELL_GATEWAY_SERVICE` to `openshell`. One of these needs to change.

### Remaining Major (3) — Reframed for the pivot

- **SecurityContext improvement should be explicit** — If gateway mode means runner pods no longer need root + SYS_ADMIN + NET_ADMIN (because the gateway handles isolation externally), that's a major security win. The spec should state this explicitly since it's one of the key motivations for the pivot. If pods *still* need elevated privileges even in gateway mode, that's worth explaining too.
- **"Sandbox not found → Completed"** — Still risky regardless of architecture. Gateway restarts, admin cleanup, or bugs would silently mark sessions as completed. Consider requiring an exit code or last-known-phase check before assuming completion.
- **`OPENSHELL_ENABLED` flag semantics** — With the pivot context, reusing the same flag makes more sense (it becomes "enable OpenShell" regardless of mode). But the *existing* behavior it triggers (ConfigMap propagation, elevated SecurityContext, policy env vars) needs to be explicitly deprecated/removed. The spec should note that the current file-mode behavior gated by this flag will be replaced, and the config table should clarify that `OPENSHELL_POLICY_CONFIGMAP` (currently in `config.go`) becomes unused.

### Remaining Minor (3) — Unchanged

- gRPC connection health/retry semantics
- Credential rotation gap (fine as known v1 limitation if documented)
- Feature flag gate per CLAUDE.md conventions

### Dropped concerns

- **Proto output path** — If this is a new architectural direction, establishing a `pkg/` convention for shared generated code is a reasonable choice. Not a concern.
- **Sandbox naming** — The spec's approach is consistent enough with existing patterns. Fine as-is.

### Net assessment

The pivot direction is good. Gateway-delegated sandboxing removes the need for elevated pod privileges, simplifies the runner image, and centralizes policy enforcement. The spec is well-structured for implementation. The remaining items are mostly about making the spec self-consistent and documenting the transition cleanly.

**Confidence:** High

— Amber 🤖

specs/platform/openshell-sandbox-provisioning.spec.md
squizzi: `OPENSHELL_MODE=gateway` or `OPENSHELL_USE_GATEWAY=true` seem to be a good fit for this.  It's probably worth keeping the file mode implementation intact in case we end up needing to roll back to that if the gateway's added complexity is too much to manage for the MVP.

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: ● Session IDs are KSUIDs (27 base62 characters — alphanumeric, no hyphens). They're already DNS-safe and well under 40 characters, so truncation and lowercasing are purely
  defensive. The reviewer's collision concern doesn't apply in practice but the spec should be explicit about this.

specs/platform/openshell-sandbox-provisioning.spec.md
squizzi: I'm actively improving pod_sync now in #152 I think it'd be worthwhile to hash out the lifecycle for a openshell sandbox to make sure the UX for session creation isn't frustrating, I have no idea what phases a openshell sandbox pod goes through, what `status.conditions` it potentially advertises, etc.

specs/platform/openshell-sandbox-provisioning.spec.md
squizzi: I think we'd probably just vendor the gRPC client code from upstream openshell into this project right?  We could probably `go build` the stubs from upstream so our client stays synced, but we'd probably want to pin it or something.

specs/platform/openshell-sandbox-provisioning.spec.md
squizzi: Again, I feel like openshell either has solved this or will solve it upstream and we should contribute back to their gRPC client implementation.  We should figure out how to vendor that into our project and pin it so it doesn't break across versions.

specs/platform/openshell-sandbox-provisioning.spec.md
squizzi: Openshell egress is supposed to pick up new creds and replace them as they pass across the gateway.  What I don't get is how we configure the credential switch so the gateway is aware of it because the gateway has a cred db if I'm not mistaken, so we need to either point that at our own custom DB or use there's.

bsquizz: Until the gateway has native support for reading creds from k8s Secrets, I think we are going to have to watch for the secret to change on k8s, and then trigger a credential refresh: https://docs.nvidia.com/openshell/sandboxes/manage-providers#credential-refresh

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Updated spec in latest commit, let's see if it addresses this.

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Yes, vendoring

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Yes they do have some status and phases, but apparently its not "enough". For example, it has no way to tell if an agent run exited successfully or crashed. So I think maybe we keep the concept of the 'runner' in there, which is communicating status back to ACP. We would bake the runner into our sandbox base image

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Updated spec to indicate that we're keeping both modes for now, with a feature flag or env var controlling the "use gateway" behavior

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,388 @@
...
+- GIVEN a running session inside a gateway-managed sandbox
+- WHEN the runner pushes a `RUN_FINISHED` event to the control plane
+- THEN the session phase SHALL transition to `Completed` (existing behavior, unchanged)
+- AND the sandbox MAY be cleaned up by the gateway after the runner process exits
+
+#### Scenario: Abnormal termination via sandbox disappearance
+
+- GIVEN a running session with an active sandbox
+- AND the gateway is reachable
+- AND the control plane has NOT received a `RUN_FINISHED` event for this session
```
bsquizz: **Major — `RUN_FINISHED` tracking state not specified**

The dual-signal lifecycle hinges on whether the CP "HAS received a `RUN_FINISHED` event for this session" (line 225). The spec doesn't specify how this state is tracked or persisted.

Is it a new field on the session record in PostgreSQL? An in-memory set? If the control plane restarts between receiving `RUN_FINISHED` and the next sandbox status poll, this state could be lost — causing a sandbox disappearance to be incorrectly mapped to `Failed` instead of being a no-op.

Since this is the keystone of the dual-signal design, the persistence mechanism should be stated explicitly. Suggestion: the session phase itself could serve as the tracking mechanism — `RUN_FINISHED` transitions the session to `Completed`, and the status syncer only checks sandbox status for sessions still in `Running` phase.

---
*Generated by Claude Code*

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,388 @@
...
+- THEN the client SHALL evict the cached connection and create a fresh one on the next call
+
+#### Scenario: Shutdown cleanup
+
+- WHEN the control plane shuts down
+- THEN it SHALL close all cached gRPC connections
+
+### Requirement: Feature Flag Gate
+
+Gateway mode SHALL be gated behind an Unleash feature flag (`openshell-gateway-provisioning`) in addition to the `OPENSHELL_USE_GATEWAY` environment variable. Both MUST be enabled for gateway mode to activate. This provides gradual rollout capability and a kill-switch that does not require redeployment.
```
bsquizz: **Major — No Unleash integration exists in the control plane today**

The spec requires the `openshell-gateway-provisioning` Unleash feature flag, but there is no Unleash client, feature flag library, or toggle mechanism in the control plane codebase. This is a dependency that needs to be built first.

Consider either:
1. Adding a note that Unleash integration is a prerequisite (with a reference to the planned work), or
2. Deferring the feature flag requirement to a follow-up spec and having this spec work with just the `OPENSHELL_USE_GATEWAY` env var for the initial implementation

---
*Generated by Claude Code*

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,388 @@
...
+#### Scenario: Sandbox disappearance after runner completion
+
+- GIVEN a running session with an active sandbox
+- AND the control plane HAS received a `RUN_FINISHED` event for this session
+- WHEN the status syncer calls `GetSandbox` and receives a not-found response
+- THEN no phase change SHALL occur (session is already in a terminal phase)
+
+### Requirement: Sandbox Status Syncing
+
+The status syncer SHALL poll the OpenShell gateway for sandbox phase as a secondary signal. The OpenShell `SandboxPhase` enum does not include a `SUCCEEDED` or `COMPLETED` state, and `SandboxStatus` does not expose an exit code. The gateway sandbox phase is used to detect error conditions and abnormal terminations that the runner cannot self-report.
```
bsquizz: **Minor — Polling interval and `WatchSandbox` future path**

The status syncer polls the gateway for sandbox status, but the polling interval is not specified. The existing `PodStatusSyncer` uses a 15-second interval (`podSyncInterval` in `pod_sync.go`). Should the sandbox syncer reuse this interval or configure a different one? Gateway RPCs may have different latency characteristics than local pod listing.

Also worth noting: the OpenShell proto defines a `WatchSandbox` streaming RPC that could replace polling with push-based updates. Since the watcher code (`internal/watcher/watcher.go`) already uses gRPC streaming for API server events, this would be a natural future optimization. Consider adding a sentence acknowledging this as a known improvement path for a later iteration.

---
*Generated by Claude Code*

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,388 @@
...
+### Requirement: Sandbox Deprovisioning
+
+When a session is stopped or deleted, the control plane SHALL delete the sandbox and its associated providers via the OpenShell gateway.
+
+#### Scenario: Session stopping
+
+- GIVEN a running session with an active sandbox
+- WHEN the session phase transitions to `Stopping`
+- THEN the control plane SHALL call `DeleteSandbox` with the session's sandbox name
+- AND the control plane SHALL call `DeleteProvider` for each provider created for the session
```
bsquizz: **Minor — Provider cleanup: how does the CP know which providers to delete?**

The spec says the CP SHALL call `DeleteProvider` for each provider created for the session, but doesn't specify how the CP tracks which providers were created. Is the provider list derived from the session's credential list at deletion time? Stored alongside the session record?

If a credential is added or rotated mid-session (via `UpdateProvider`), the list at deletion time could diverge from what was actually created. Consider specifying whether the CP re-derives the provider names from credentials at cleanup time, or stores the created provider names for later deletion.

---
*Generated by Claude Code*

bsquizz (commented):
Good spec overall — the previous Amber review concerns have been addressed and the dual-mode coexistence story is solid. Left a few comments, mostly around implementation details that would help whoever picks this up. The two bigger items are clarifying how RUN_FINISHED receipt is persisted across CP restarts, and noting that Unleash integration is a prerequisite that doesn't exist in the CP yet.

---
*Generated by Claude Code*

specs/platform/openshell-sandbox-provisioning.spec.md
squizzi: I think Amber is reviewing this as sessions behave now -- where-in sessions don't run a prompt and then consider themselves complete in ACP, they remain running until stopped or until the agent self-terminates.

We talked about targeting single run prompts (not steerable prompts) for the MVP, perhaps we should explictly state in the spec that this is the intent of the first iteration of this design?

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,372 @@
...
+
+### Requirement: Sandbox Environment Variables
+
+The control plane SHALL pass session configuration to the sandbox as environment variables in the `CreateSandboxRequest.Spec.Template.Environment` map.
+
+#### Scenario: Environment variable translation
+
+- GIVEN a session with LLM model, prompt, repo URL, and proxy settings
+- WHEN the sandbox is created
+- THEN all environment variables from `buildEnv()` that have literal string values SHALL be included
```
bsquizz, line 194: **Major — `buildEnv()` leaks file-mode env vars into gateway sandboxes**

The spec says all literal string values from `buildEnv()` are included in the sandbox environment. But `buildEnv()` currently adds 4 file-mode-specific env vars when OpenShell is enabled (`kube_reconciler.go:948-951`):

- `OPENSHELL_ENABLED=true`
- `OPENSHELL_POLICY_RULES=/etc/openshell/policy.rego`
- `OPENSHELL_POLICY_DATA=/etc/openshell/policy.yaml`
- `OPENSHELL_LOG_LEVEL=debug`

In gateway mode there is no in-container Supervisor to read these. The policy file paths reference `/etc/openshell` — a volume mount that the spec explicitly says is NOT added in gateway mode (line 110). `OPENSHELL_ENABLED=true` could mislead the runner into trying to initialize file-mode behavior.

Suggestion: either specify that file-mode OpenShell env vars are excluded in gateway mode, or add a note that their presence is harmless and explain why.

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,372 @@
...
+- AND behavior SHALL be identical to the current file-mode implementation
+
+### Requirement: Credential Mapping to OpenShell Providers
+
+The control plane SHALL map ambient platform credentials to project-scoped OpenShell providers, replacing the credential sidecar container pattern. Providers are scoped to the project namespace — all sandboxes within a namespace share the same set of providers. The control plane ensures providers exist before creating sandboxes and updates them when credentials change.
+
+The gateway's egress proxy resolves credential placeholders to real values at request time — the agent process inside the sandbox never holds real credentials, only opaque placeholders. This means provider updates take effect immediately for subsequent requests without restarting any sandbox. If the proxy encounters a placeholder it cannot resolve, it rejects the request with HTTP 500 rather than forwarding the raw placeholder upstream (fail-closed).
+
+This iteration focuses on **scheduled agent runs** (short-lived sessions). Long-running session credential lifecycle (token expiry mid-session, gateway-managed refresh via OAuth2/client-credentials flows) is out of scope.
```
bsquizz: **Minor — Iteration scope should be more prominent**

squizzi's comment asked for explicit scoping to single-run sessions (not steerable prompts) for MVP. This scoping statement exists here but is buried in the Credential Mapping section. Consider elevating it to the Purpose section or adding a dedicated "Scope / Iteration 1 Constraints" section near the top so implementors see it before diving into requirements.

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,372 @@
...
+
+- GIVEN the gateway is temporarily unreachable (connection refused, deadline exceeded, or other transport error)
+- WHEN the status syncer polls
+- THEN it SHALL log a warning and retry on the next sync cycle
+- AND it SHALL NOT change the session phase
+- AND it SHALL NOT treat the unreachable gateway as sandbox disappearance
+
+### Requirement: Sandbox Network Policy for Runner Events
+
+The OpenShell sandbox network policy SHALL permit the runner process to push gRPC events to the control plane backend. Without this, the runner cannot report `RUN_FINISHED` and all sandbox exits would be treated as abnormal terminations.
```
bsquizz, line 289: **Minor — Network policy ownership unclear**

Who creates the network policy that permits runner gRPC egress to the control plane? If it is part of the OpenShell Helm chart, ACP does not need to act. If ACP needs to create/manage it, that is a new reconciler requirement not listed in the Migration table (lines 354-368). Worth a sentence clarifying ownership.

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,372 @@
...
+
+The gateway's egress proxy resolves credential placeholders to real values at request time — the agent process inside the sandbox never holds real credentials, only opaque placeholders. This means provider updates take effect immediately for subsequent requests without restarting any sandbox. If the proxy encounters a placeholder it cannot resolve, it rejects the request with HTTP 500 rather than forwarding the raw placeholder upstream (fail-closed).
+
+This iteration focuses on **scheduled agent runs** (short-lived sessions). Long-running session credential lifecycle (token expiry mid-session, gateway-managed refresh via OAuth2/client-credentials flows) is out of scope.
+
+#### Scenario: Ensuring project providers exist
+
+- GIVEN a project has configured credentials for `github` and `anthropic`
+- WHEN the control plane provisions a sandbox in that project's namespace
+- THEN it SHALL ensure an OpenShell provider exists for each credential via `CreateProvider` (idempotent — skip if already exists)
```
bsquizz, line 139: **Minor — Partial provider creation failure not addressed**

If `CreateProvider` succeeds for `github` but fails for `anthropic`, should sandbox creation proceed with partial providers or fail entirely? The "no fallback" posture elsewhere in the spec (gateway unreachable → fail, don't fall back) suggests fail-entirely, but it is not stated for provider creation. A sentence like "If any provider creation fails, sandbox creation SHALL fail" would remove ambiguity for implementors.

bsquizz (commented):
Good spec overall — the dual-mode coexistence story is solid and the dual-signal lifecycle design is sound. The spec's claims against the codebase all check out (verified `safeResourceName()`, `buildEnv()`, `podSyncInterval`, security context logic, gRPC watcher patterns, and the absence of Unleash integration). The latest commit addressed most earlier feedback well.

One item worth resolving: `buildEnv()` env var filtering in gateway mode — see inline comment for details. The other three are minor clarifications.

---
*Generated by Claude Code* 🤖

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Incorrect, will be a supervisor in the sandbox.

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,383 @@
...
+- THEN the control plane SHALL call `UpdateProvider` on the gateway with the new credential values
+- AND the gateway's egress proxy SHALL resolve subsequent requests to the updated credentials at request time
+- AND no sandboxes SHALL be restarted
+
+#### Scenario: Provider type mapping
+
+- GIVEN the following ambient credential provider names
+- THEN they SHALL map to OpenShell provider types as follows:
+
+| Ambient Provider | OpenShell Type |
```
devguyio, line 175: What's the source of truth for the OpenShell Type names? Might be a good idea to add a link to that as a reference

devguyio (commented):
I'm no expert here but I was curious to skim through the specs. Left one question. 

@markturansky @jsell-rh  are e2e tests and documentation typically part of the specs? Or will the agent build e2e tests and docs by default?

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Good suggestion. It comes from
* https://docs.nvidia.com/openshell/sandboxes/manage-providers#supported-provider-types
* https://docs.nvidia.com/openshell/sandboxes/manage-providers#supported-inference-providers

Will add to the spec

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,383 @@
...
+When the platform operates in OpenShell mode, the control plane SHALL delegate agent pod creation to an OpenShell gateway running in each project namespace, instead of creating Kubernetes pods directly. This provides policy-enforced sandboxing (network, filesystem, process controls) for all agent sessions through OpenShell's security layer.
+
+The OpenShell gateway exposes a gRPC service (`openshell.v1.OpenShell`) that manages sandbox lifecycle. Each project namespace has an OpenShell gateway pre-installed via Helm chart. The control plane discovers it via Kubernetes Service DNS.
+
+### Iteration 1 Constraints
+
+This iteration is scoped to **scheduled agent runs** (single-run, short-lived sessions). The following are explicitly out of scope:
+
+- **Long-running / steerable sessions** — credential lifecycle concerns (token expiry mid-session, gateway-managed refresh via OAuth2/client-credentials flows) are deferred
+- **Gateway provisioning** — the OpenShell gateway is assumed to already be deployed in each project namespace; ACP will not create it
```
squizzi: Is this true?  ACP should manage gateway creation should it not?  ACP is an operator so we should be able to have it create and manage the gateway per org, the maintainability of creating these gateways manually is going to be a PITA.  It can deploy and reconcile gateway state and we could adapt the gateway helm chart into an OpenShellGateway CR/CRD: https://github.com/NVIDIA/OpenShell/blob/main/deploy/helm/openshell/values.yaml

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,383 @@
...
+## Requirements
+
+### Requirement: Gateway-Based Sandbox Creation
+
+When `OPENSHELL_USE_GATEWAY` is true, the control plane SHALL create agent sandboxes by calling the OpenShell gateway's `CreateSandbox` gRPC RPC instead of creating Kubernetes pods directly. When `OPENSHELL_USE_GATEWAY` is false, the existing provisioning path SHALL be used unchanged (either file-mode sandbox or direct pod creation, depending on `OPENSHELL_ENABLED`).
+
+#### Scenario: Session provisioning with gateway mode
+
+- GIVEN `OPENSHELL_USE_GATEWAY` is `true`
+- AND an OpenShell gateway is running in the project namespace
```
squizzi, line 48:
```suggestion
- AN OpenShell gateway will be provisioned and configured by the control plane for that projects namespace, one gateway per project namespace
```

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,383 @@
...
+- THEN the control plane SHALL call `CreateSandbox` on the gateway in the session's project namespace
+- AND the sandbox SHALL be created with the runner image, session environment variables, and attached credential providers
+- AND the session phase SHALL transition to `Running`
+
+#### Scenario: Session provisioning without gateway mode
+
+- GIVEN `OPENSHELL_USE_GATEWAY` is `false`
+- WHEN a session transitions to `Pending` phase
+- THEN the control plane SHALL use the existing provisioning path (direct pod creation, with file-mode sandbox if `OPENSHELL_ENABLED` is true)
+- AND no interaction with the OpenShell gateway SHALL occur
```
squizzi:
```suggestion
- AND no interaction with the OpenShell gateway SHALL occur and no OpenShell gateway will be provisioned/reconciled
```

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Requirement: Gateway Discovery via Service DNS
+
+The control plane SHALL discover the OpenShell gateway in each project namespace using Kubernetes Service DNS with a configurable service name and port.
+
+#### Scenario: Default discovery
+
+- GIVEN no override configuration is set
+- WHEN the control plane needs to reach the gateway in namespace `my-project`
+- THEN it SHALL connect to `openshell-gateway.my-project.svc.cluster.local:8080`
```
squizzi: The helm values have a section detailing the gRPC route details: https://github.com/NVIDIA/OpenShell/blob/main/deploy/helm/openshell/values.yaml#L357

Which we might be able to use for discovery bits as well by just getting the openshell CR config from kube (once it's defined of course)

squizzi (changes requested):

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,383 @@
...
+
+- **Long-running / steerable sessions** — credential lifecycle concerns (token expiry mid-session, gateway-managed refresh via OAuth2/client-credentials flows) are deferred
+- **Gateway provisioning** — the OpenShell gateway is assumed to already be deployed in each project namespace; ACP will not create it
+- **Namespace-level credential storage** — credentials remain stored in ACP, not as Kubernetes Secrets in the project namespace
+- **Network policy ownership** — OpenShell policies (including network egress rules allowing runner-to-control-plane gRPC) will be user-configurable via the Agent spec. Whether ACP auto-injects the control plane egress rule on top of the user-configured policy or requires users to include it explicitly is TBD
+
+### Relationship to `specs/security/openshell-sandbox.spec.md`
+
+This specification defines **gateway mode** — an alternative to the **file mode** sandbox approach defined in `specs/security/openshell-sandbox.spec.md`. Both modes coexist and are selectable at deployment time via the `OPENSHELL_USE_GATEWAY` environment variable.
```
devnulljason: All these references to other docs in the repo can be links to improve navigability

```suggestion
This specification defines **gateway mode** — an alternative to the **file mode** sandbox approach defined in [openshell-sandbox.spec.md]. Both modes coexist and are selectable at deployment time via the `OPENSHELL_USE_GATEWAY` environment variable.

[openshell-sandbox.spec.md]: ../security/openshell-sandbox.spec.md
```
If you use a reference link like ☝️ you can put all the links at the bottom of the document so they're not obtrusive while reading in plain-text.

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,383 @@
...
+- AND it SHALL NOT propagate the OpenShell policy ConfigMap
+- AND it SHALL NOT add the `/etc/openshell` volume mount
+- AND the `CreateSandboxRequest` SHALL contain only image, environment, and provider references
+- AND all pod-level security settings SHALL be the gateway's responsibility
+
+#### Scenario: File mode — elevated SecurityContext preserved
+
+- GIVEN `OPENSHELL_USE_GATEWAY` is `false` and `OPENSHELL_ENABLED` is `true`
+- WHEN the control plane provisions a session
+- THEN it SHALL apply the elevated SecurityContext as defined in `specs/security/openshell-sandbox.spec.md` § Container Security Context
```
devnulljason: Can also be linked

```suggestion
- THEN it SHALL apply the elevated SecurityContext as defined in [openshell-sandbox.spec.md][sandbox-security-context]

[sandbox-security-context]: ../security/openshell-sandbox.spec.md#requirement-container-security-context
```

devnulljason (commented):
Suggestions for human readability.

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: I'll say provisioning the gateway is a possible follow-up item to limit the scope of this PR, we'll sort that one out next.

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Will update spec to indicate this is scoped for a future iteration

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Good point, updating spec

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Had the robot generate a lot of links :)

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: Went with option 2

specs/platform/openshell-sandbox-provisioning.spec.md
bsquizz: We're getting rid of the concept of per-session credentials. Credentials will be defined per-project. Sandboxes may be bound to those credential providers via the 'openshell sandbox create' logic

squizzi (approved):

jsell-rh:
> are e2e tests and documentation typically part of the specs? Or will the agent build e2e tests and docs by default?
@devguyio usually e2e tests would fall out of implementation as guided _by a skill_, though I could imagine the tests mirroring the spec structure in some cases.

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -0,0 +1,399 @@
...
+- **Network policy ownership** — OpenShell policies (including network egress rules allowing runner-to-control-plane gRPC) will be user-configurable via the Agent spec. Whether ACP auto-injects the control plane egress rule on top of the user-configured policy or requires users to include it explicitly is TBD
+
+### Relationship to [openshell-sandbox.spec.md]
+
+This specification defines **gateway mode** — an alternative to the **file mode** sandbox approach defined in [openshell-sandbox.spec.md]. Both modes coexist and are selectable at deployment time via the `OPENSHELL_USE_GATEWAY` environment variable.
+
+**File mode** (existing, [openshell-sandbox.spec.md]): The OpenShell Supervisor binary runs inside the runner container, wrapping the Claude CLI process directly. Requires policy ConfigMap propagation, elevated container security context, and the Supervisor binary baked into the runner image.
```
jsell-rh:
```suggestion
**File mode** (existing, [openshell-sandbox.spec.md]): The OpenShell Supervisor binary runs inside the runner container, wrapping the Claude Agent SDK runner process directly. Requires policy ConfigMap propagation, elevated container security context, and the Supervisor binary baked into the runner image.

PR #158, opened 2026-06-24
feat(ui): add platform health banner to dashboard

squizzi:
## Summary

When hacking on ACP locally my control plane was in `CrashLoopBackOff` but I had no visibility into it from the UI, this PR adds a warning banner to the dashboard when the API server or control plane is unreachable. A new server-side `/api/health/platform` route checks both services from within the cluster, polled every 30s via a React Query hook. The banner is gated behind auth state so it doesn't flash during Keycloak redirects. The CONTROL_PLANE_URL env var is added to the UI deployment manifest pointing at the CP token server's /healthz endpoint.

## Screenshots
Screenshot 2026-06-24 at 3 42 21 PM

Screenshot 2026-06-24 at 3 42 07 PM

## Test plan
- [ ] Deploy to Kind cluster, verify banner does NOT appear when healthy
- [ ] kubectl scale deploy ambient-control-plane -n ambient-code --replicas=0 — wait 30s, verify amber banner appears
- [ ] Scale back up — verify banner auto-clears
- [ ] kubectl scale deploy ambient-api-server -n ambient-code --replicas=0 — verify banner appears
- [ ] Log out — verify banner does not appear during Keycloak redirect

### DIFF
--- a/components/ambient-ui/package-lock.json
+++ b/components/ambient-ui/package-lock.json
@@ -69,7 +69,7 @@
         "@types/node": "^20.0.0",
         "jest": "^29.7.0",
         "ts-jest": "^29.1.0",
-        "typescript": "~5.5.0"
+        "typescript": "~5.9.0"
       },
       "engines": {
         "node": ">=18.0.0"
@@ -9628,7 +9628,6 @@
       "version": "2.3.2",
       "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.3.2.tgz",
       "integrity": "sha512-xiqMQR4xAeHTuB9uWm+fFRcIOgKBMiOBP+eXiyT7jsgVCq1bkVygt00oASowB7EdtpOHaaPgKt812P9ab+DDKA==",
-      "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
       "optional": true,
--- a/components/ambient-ui/src/app/(dashboard)/layout.tsx
+++ b/components/ambient-ui/src/app/(dashboard)/layout.tsx
@@ -5,6 +5,7 @@ import { usePathname } from 'next/navigation'
 import { AppSidebar } from '@/components/app-sidebar'
 import { NavHeader } from '@/components/nav-header'
 import { StatusBar } from '@/components/status-bar'
+import { PlatformHealthBanner } from '@/components/platform-health-banner'
 import { ChatSidebar } from '@/components/chat-sidebar'
 import { ChatSidebarProvider } from '@/components/chat-sidebar-context'
 import { CommandPalette } from '@/components/command-palette'
@@ -102,6 +103,7 @@ export default function DashboardLayout({
             sessionName={sessionId ? (session?.name ?? sessionId) : null}
             detailName={agentId ? (agent?.displayName ?? agent?.name ?? agentId) : null}
           />
+          <PlatformHealthBanner />
           <div className="flex-1 p-6 pb-10 max-w-7xl">{children}</div>
           <StatusBar />
           <CommandPalette />
--- a/components/ambient-ui/src/app/api/health/platform/route.ts
+++ b/components/ambient-ui/src/app/api/health/platform/route.ts
@@ -1,0 +1,51 @@
+import { env } from '@/lib/env'
+
+export const runtime = 'nodejs'
+export const dynamic = 'force-dynamic'
+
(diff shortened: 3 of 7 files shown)


jsell-rh:
🤖 **Amber Review — PR #158: feat(ui): add platform health banner**

**Verdict: ✅ Approved**

Clean, well-scoped feature. The dashboard blind spot during local development (CrashLoopBackOff invisible from UI) is a real pain and this addresses it well.

**What I checked:**

**Frontend checklist** ✅
- Zero `any` types — `ComponentStatus`, `PlatformHealthResponse`, and `UsePlatformHealthReturn` are all properly typed
- React Query used for all data ops (`useQuery` with `refetchInterval: 30_000`) — correct pattern
- `type` over `interface` — followed throughout
- Loading/error states handled — banner hidden during load, suppressed on auth redirect via `enabled: user !== null`
- Single-use component (`platform-health-banner.tsx`) colocated with its hook
- No manual `fetch()` in components — delegate to hook layer

**Security** ✅
- Server-side route (`/api/health/platform`) proxies internal cluster URLs — no credentials exposed to browser
- `AbortSignal.timeout(5000)` prevents hung goroutines from blocking the route
- `CONTROL_PLANE_URL` is optional (`getOptionalEnv`) — correct, CP health is best-effort

**Architecture** ✅
- `force-dynamic` + `runtime = 'nodejs'` are correct for a health-check route that must not be cached
- `Promise.allSettled` is the right choice — API server down shouldn't prevent CP status from returning
- `CONTROL_PLANE_URL` pointing at `http://ambient-control-plane:8080` aligns with existing token server config in control plane

**One observation (non-blocking):**

The `queryKey: ['platform-health']` doesn't include any scoping parameters. If this were ever rendered in multiple namespaces simultaneously that'd cause stale sharing. For a global banner this is fine — just noting it for when per-project health checks appear.

**Confidence:** High [92%] — straightforward frontend feature, clean API proxy pattern, well-gated behind auth state.

**Rollback:**
```bash
git revert 20e5fee3  # Reverts all 7 changed files
```

*— Amber*

jsell-rh:
🤖 **Amber Review**

Solid feature, nearly there. Two items to address before merge.

**Summary:** Health polling via React Query is correctly wired, auth-gated, and the manifest change is straightforward. Two convention issues and one behavior edge case.

<details>

**Strengths:**
- `useQuery` with `refetchInterval: 30_000` and `retry: 1` ✅
- Auth-gated (`enabled: user !== null`) — no flash during Keycloak redirect ✅
- Zero `any` types throughout ✅
- `type` keyword used consistently ✅
- `Promise.allSettled` in the route — won't short-circuit if one check fails ✅
- `AbortSignal.timeout(5000)` prevents hung health checks ✅
- TypeScript types are properly shared between route and hook ✅

**Issues:**

**Minor — Shadcn Alert component not used (`platform-health-banner.tsx`)**
The banner renders a raw `<div role="alert">` with manual className styling. Convention requires Shadcn UI components. Shadcn ships an `Alert` / `AlertDescription` pair that handles this:

```tsx
import { Alert, AlertDescription } from '@/components/ui/alert'
import { AlertTriangle } from 'lucide-react'

<Alert variant="warning" className="rounded-none border-x-0 border-t-0">
  <AlertTriangle className="h-4 w-4" />
  <AlertDescription>
    Platform services are degraded. There may be issues starting, stopping,
    or modifying sessions until the issue is resolved.
  </AlertDescription>
</Alert>
```
(Adjust variant/className to match your status token design.)

**Minor — Silent failure edge case in `usePlatformHealth`**
If the `/api/health/platform` route itself errors (e.g. Next.js server issue), `data` is `undefined`, `isPending` is `false`, and the hook returns `apiServer: 'healthy'` — the banner never shows. This is fail-open, which is probably intentional, but it means a broken health endpoint is indistinguishable from "everything healthy." Consider logging a warning or surfacing a distinct state if `isError` is true:

```ts
const { data, isPending, isError } = useQuery(...)
// if isError, you could still return isHealthy: false
```

**Verify — `CONTROL_PLANE_URL` port**
Manifest sets `http://ambient-control-plane:8080` and the route appends `/healthz`. Confirm port 8080 is where the control plane token server's healthz lives (not the gRPC port or metrics port). Easy to verify with `kubectl get svc ambient-control-plane -n ambient-code`.

**Confidence:** High on the logic; Medium on the port assumption (needs cluster verification per test plan).

</details>

*— Amber*

PR #179, opened 2026-06-25
feat(control-plane): OpenShell gateway-based sandbox provisioning

bsquizz:
## Summary

Adds OpenShell gateway-based sandbox provisioning to the control plane, enabling ACP to delegate session sandbox lifecycle to an external OpenShell gateway via gRPC instead of creating Kubernetes Jobs directly.

### What changed

**Control plane — gateway provisioning path**
- New `OPENSHELL_USE_GATEWAY` env var selects the provisioning backend at startup
- `provisionSessionSandbox` / `cleanupSessionSandbox` / `deprovisionSessionSandbox` in `kube_reconciler.go` handle the gateway path: create `AgentSandbox` CRs via the OpenShell gRPC API, stream exec commands, and clean up on session teardown
- `pod_sync.go` gates namespace provisioning: pod mode creates/deletes namespaces via `StandardNamespaceProvisioner`; gateway mode bypasses the provisioner entirely (gateway owns the namespace)
- `StandardNamespaceProvisioner` restored to full create/delete behavior for pod mode (was accidentally made verify-only in an earlier commit)

**OpenShell gRPC client**
- `internal/openshell/gateway_client.go` — typed client wrapping the gRPC stubs: `CreateSandbox`, `DeleteSandbox`, `ExecSandbox`, `WatchSandboxStatus`
- `internal/openshell/tls_resolver.go` — resolves mTLS credentials from a K8s Secret; base64-encoded PEM fields (`tls.crt`, `tls.key`, `ca.crt`)
- `internal/openshell/provider_mapping.go` — maps ACP provider names to OpenShell provider identifiers

**Vendored proto files + generated stubs**
- `proto/openshell/` — vendored subset of `github.com/NVIDIA/OpenShell` proto definitions (sandbox lifecycle, exec, provider management RPCs)
- `proto/VENDOR.md` — documents the vendoring process
- `scripts/vendor-proto.sh` + `make vendor-openshell-proto REF=<tag>` — reproducible one-command update: resolves tag → SHA, fetches, rewrites `go_package`, stamps SHA in headers, updates VENDOR.md, runs `buf generate`

**Dual-tenant kind automation**
- `scripts/setup-kind-openshell.sh` — installs `agent-sandbox` CRD+controller, deploys OpenShell gateway Helm chart per tenant namespace, creates ACP projects via API, patches control plane env
- `make kind-up OPENSHELL_USE_GATEWAY=true` — full dual-tenant setup (default tenants: `tenant-a`, `tenant-b`; override with `OPENSHELL_TENANTS`)
- `tests/openshell-dual-tenant.sh` + `make test-openshell-dual-tenant` — E2E test: gateway health, CRD presence, ACP project existence, concurrent session creation, concurrent sandbox provisioning

**Spec + skills**
- `specs/platform/openshell-sandbox-provisioning.spec.md` — updated to reflect the pod/gateway mode split, namespace lifecycle per mode, and migration table
- `skills/build/openshell-gateway/` — new skill for setting up and testing gateway mode in kind
- `skills/build/dev-cluster/SKILL.md` — restored sections removed during earlier work: full frontend validation retry loop, common tasks, decision tree, access URL detection

## Test plan

- [ ] `make kind-up OPENSHELL_USE_GATEWAY=true` — cluster comes up with dual-tenant gateway
- [ ] `make test-openshell-dual-tenant` — all 5 sections pass (gateway health, CRD, projects, concurrent session creation, concurrent sandbox provisioning)
- [ ] `make kind-up LOCAL_IMAGES=true` (no gateway flag) — pod-mode sessions still work; namespaces created and deleted normally
- [ ] `make kind-reload-ambient-control-plane` after a CP change in gateway mode — `OPENSHELL_USE_GATEWAY=true` persists, sandbox provisioning continues
- [ ] `make vendor-openshell-proto REF=<tag>` — proto files and stubs updated, VENDOR.md stamped


jsell-rh (changes requested):
🤖 **Amber Review — Changes Requested**

> **Draft PR** — feedback is early-stage and constructive. The architecture is solid; these are fixable before promoting.

## Summary

Gateway-based sandbox provisioning is well-designed. Per-namespace connection pooling, mTLS credential resolution via `CredentialResolver` abstraction, eviction-on-`Unavailable`, and `defer gateway.Close()` in `runKubeMode` are all correct. The Makefile unique-tag reload genuinely fixes the imagePullPolicy stale-image problem. UI project validation guard is clean. Config env var isolation for `OPENSHELL_GATEWAY_*` fields is right.

Two things need to be addressed before this is merge-ready.

---

## Findings

### Major — Tests missing for `gateway_client.go` (Constitution Principle IV)

`internal/openshell/gateway_client.go` is ~207 lines of connection pooling, eviction logic, and credential resolution with **zero tests**. The double-check lock, evict-on-`Unavailable` path, per-namespace caching, and `CreateSandbox`/`DeleteSandbox` eviction behavior should all be covered.

Minimum expected:
- `getOrCreateConn` caches a connection after first call (second call returns same conn)
- `evictConn` removes the namespace key and closes the connection
- `shouldEvict` returns `true` only for `codes.Unavailable`, not for other codes
- `CreateSandbox` calls `evictConn` when the underlying call returns `Unavailable`
- `gatewayEndpoint` produces the expected DNS URI format

### Minor — `SandboxName`/`sandboxName` duplication

`gateway_client.go` defines both a private `sandboxName()` and an exported `SandboxName()` that just calls it. One of them is redundant — rename `sandboxName` to `SandboxName` directly and drop the wrapper.

### Minor — Vendored protobuf has no tracked upstream version

The `*.pb.go` files under `internal/openshell/grpc/` are marked `"Vendored from github.com/NVIDIA/OpenShell/proto/..."` but there's no record of which OpenShell commit or tag these were generated from. `go get` and Dependabot won't track them. Add a short comment or `VENDOR.md` noting the upstream ref so future maintainers know what to re-run when OpenShell protos change.

---

## What Looks Good ✅

- Double-check lock pattern in `getOrCreateConn` is correct (read → check → write → check again)
- TLS enabled by default (`!= "false"`) — secure-by-default ✅
- `CredentialResolver` abstraction cleanly separates TLS vs insecure paths ✅
- No credentials or TLS key material logged — only `namespace` and `target` endpoint ✅
- `fmt.Errorf("resolving TLS credentials for namespace %s: %w", ...)` — correct wrapping ✅
- `errors.As`-compatible error inspection via `status.FromError` ✅
- `defer gateway.Close()` guarantees cleanup even on error paths in `runKubeMode` ✅
- Makefile unique-tag (`git-short-hash + epoch`) guarantees rollout with `imagePullPolicy: IfNotPresent` ✅
- UI `[projectId]` layout guard prevents session creation against non-existent projects ✅

**Confidence:** High on the Go control plane changes. Medium on the UI layout guard (couldn't verify the full Next.js layout diff). Flag the test gap before promoting.

*— Amber*

jsell-rh (changes requested):
🤖 **Amber Re-Review** — following up on 2 new commits since my last review (`094c9e5` → `c5e7e6f`, `b4df31d`)

## New Changes Reviewed (commit `c5e7e6f`)

The new commit addresses several meaningful changes:

**✅ mTLS-only gateway auth** — Removing application-layer JWT minting is the right call. `allow_unauthenticated_users = true` with mTLS for a cluster-internal control-plane path is architecturally sound. The rationale in the spec is clear and well-justified.

**✅ `StandardNamespaceProvisioner` verify-only** — `ProvisionNamespace` now returns a clear error if the namespace doesn't exist rather than creating it; `DeprovisionNamespace` is a no-op. The logic is correct: `IsNotFound` → fail with descriptive error, any other error → wrap and propagate. This is the right behavior for externally-managed namespace lifecycles.

**✅ Base64 PEM transport** — `base64.StdEncoding.EncodeToString` in `buildGatewayEnv`, with auto-detect decode (`!strings.HasPrefix(pemStr, "-----")` check) in MCP's `parsePublicKey` and Python's `_decode_public_key`. The Python tests cover raw PEM, base64-encoded PEM, empty string, and round-trip through `_encrypt_session_id` and `from_env` — solid coverage.

**✅ Spec updated** — Namespace lifecycle requirement and PEM transport scenario are now in the spec with concrete BDD scenarios.

**✅ `b4df31d` merge commit** — clean rebase onto main, no issues.

---

## Outstanding from Prior Review — **Major still unresolved**

### Major — Tests for `gateway_client.go` (Constitution Principle IV)

`internal/openshell/gateway_client.go` still has zero tests. The new commit adds 5 Python tests for `_decode_public_key` (good), but the Go double-check lock, `shouldEvict`, `evictConn`, and `CreateSandbox`/`DeleteSandbox` eviction paths remain untested.

The minimum I flagged last time still stands:

```go
// These behaviors need test coverage:
// - getOrCreateConn: second call returns same conn (cache hit)
// - evictConn: removes namespace key and calls Close()
// - shouldEvict: true only for codes.Unavailable, not for other status codes
// - CreateSandbox: calls evictConn when RPC returns Unavailable
// - gatewayEndpoint: produces expected "passthrough:///host:port" URI
```

The double-check lock pattern is easy to get wrong under race conditions, and the evict-on-`Unavailable` path is the primary resilience mechanism. These need test coverage before this is merge-ready.

### Minor — `SandboxName`/`sandboxName` duplication still present

The exported `SandboxName()` wrapping private `sandboxName()` still exists. One of the two is redundant.

### Minor — Vendored protobuf upstream ref still untracked

The `*.pb.go` files under `internal/openshell/grpc/` still lack a comment or `VENDOR.md` recording which OpenShell commit/tag they were generated from.

---

## Summary

New commits `c5e7e6f` and `b4df31d` are all correct — mTLS auth rationale is sound, namespace lifecycle change is safe, base64 PEM transport is well-implemented and tested on the Python side. The original Major (Go tests for `gateway_client.go`) is the only blocker remaining.

**Confidence:** High — changes are well-scoped and individually verifiable.

*— Amber*

jsell-rh (approved):
🤖 **Amber Re-Review — Approved** ✅

Following up on 3 new commits since my last review (`b4df31d` → `73cfc1f`, `ad8cfed`, `c33d83c`).

---

## All Prior Blockers Resolved

### ✅ Major — Tests for `gateway_client.go` (was: Constitution Principle IV blocker)

`gateway_client_test.go` (219 lines) now covers everything I asked for and more:

- `TestGetOrCreateConn_CacheHit` — second call returns same pointer, different namespace → different conn ✅
- `TestGetOrCreateConn_ConcurrentSameNamespace` — 20 goroutines, verifies credential resolver called exactly **once** (double-check lock holds under concurrency) ✅
- `TestGetOrCreateConn_CredResolverError` — error propagation ✅
- `TestEvictConn` — key removed from map after eviction ✅
- `TestEvictConn_Noop` — no panic on missing namespace key ✅
- `TestShouldEvict` — 6 cases: `Unavailable` → true; `NotFound`, `PermissionDenied`, `Internal`, plain error, nil → false ✅
- `TestSandboxName` — truncation at 40 chars, lowercasing, prefix, empty string ✅
- `TestClose` — all conns cleared, map reset ✅
- `TestGatewayEndpoint` — DNS URI format correct ✅

### ✅ Minor — `SandboxName`/`sandboxName` duplication

Private `sandboxName()` is gone. Single exported `SandboxName()` with the implementation inline. ✅

### ✅ Minor — Vendored protobuf upstream ref

`proto/VENDOR.md` records upstream tag `v0.0.70`, vendored file list, and `buf generate` instructions. Future maintainers know exactly what to re-run when OpenShell protos change. ✅

---

## New Additions Reviewed (commits `73cfc1f`, `ad8cfed`)

### `ExecSandbox` in `gateway_client.go` ✅

Clean streaming implementation:
- Collects stdout/stderr/exit code from `ExecSandboxEvent` oneof payload ✅
- `evictConn` called on `Unavailable` for the initial RPC call (same pattern as `CreateSandbox`) ✅
- Stream recv errors wrapped: `fmt.Errorf("exec stream: %w", err)` ✅
- No credentials or command output truncated in error paths ✅

### `execAfterReady` goroutine in reconciler ✅

- `context.WithTimeout(context.Background(), 120*time.Second)` — intentionally detached from reconcile ctx (fire-and-forget placeholder). Timeout bounds the goroutine lifetime. ✅
- `defer ticker.Stop()` prevents goroutine leak on context cancellation ✅
- `SANDBOX_PHASE_ERROR` early exit ✅
- Commit message correctly labels this as a placeholder for runner startup

**Minor note (not a blocker):** If `provisionSessionGateway` is retried by the reconcile loop after a transient error (post-`CreateSandbox`, pre-update), two `execAfterReady` goroutines could spawn for the same sandbox name. The existing `GetSandbox` check at function entry catches the "already exists" path and also spawns a goroutine (line 296), so this double-spawn risk exists for the "sandbox exists but reconcile was interrupted" case. Since this is a placeholder that will be replaced by actual runner startup logic, I'm flagging it as informational only.

### `setup-kind-openshell.sh` ✅

- `set -euo pipefail` ✅
- Idempotent: namespace existence check, `helm status` guard before install ✅
- `kubectl wait` with timeout for CRD controller ✅
- Env var overrides for all defaults (`OPENSHELL_TENANT_NAMESPACE`, `AGENT_SANDBOX_VERSION`) ✅
- No secrets or credentials in the script ✅

---

## What Still Looks Good ✅

Everything from prior reviews carries forward — mTLS auth, verify-only `StandardNamespaceProvisioner`, base64 PEM transport, Python test coverage, Makefile unique-tag reload, UI project guard.

**Confidence:** High — all three prior blockers resolved, new additions clean.

*— Amber*

specs/platform/openshell-sandbox-provisioning.spec.md
```diff
@@ -11,14 +11,55 @@
...
+
+```bash
+helm install openshell-gateway oci://ghcr.io/nvidia/openshell/helm-chart \
+  --namespace tenant \
+  --set "pkiInitJob.serverDnsNames={openshell-gateway.tenant.svc.cluster.local}"
+```
+
+Alternatively, cert-manager can manage TLS certificates by setting `certManager.enabled=true` in the Helm values.
+
+The ACP control plane reads the `openshell-client-tls` Secret from the project namespace to establish mTLS connections to the gateway (see [Gateway TLS and Authentication](#requirement-gateway-tls-and-authentication)).
```
jsell-rh, line 36: Should ACP read from an env var, instead of directly grabbing a secret?

bsquizz: That secret apparently gets applied into the ns by the gateway deployment. We're going to be reading other secrets out of a tenant ns anyways, so I think it's ok to read this one as well?

jsell-rh (approved):
🤖 **Amber Re-Review** — 1 new commit since approval (`b9f79bad` → `db9a89ae`)

**✅ `db9a89ae` — `fix: go mod tidy after rebase onto main`**

`go mod tidy` update only — `golang.org/x/net` `v0.48.0 → v0.50.0`, `sys` `v0.40.0 → v0.44.0`, `term` `v0.38.0 → v0.40.0`, `text` `v0.32.0 → v0.34.0`. All patch/minor stdlib bumps picked up from the rebase onto main. No logic changes, no new dependencies, no API surface changes. `go.mod` and `go.sum` are correct and consistent.

**Re-approval confirmed.** Still approved. The open discussion thread (env var vs Secret for mTLS credentials) is an in-flight spec conversation between @bsquizz and @jsell-rh — no Amber input needed there, and it's not a blocker for the implementation as-is.

*— Amber*

specs/platform/openshell-sandbox-provisioning.spec.md
jsell-rh: Ah yes it needs to be dynamic because multi-gateway support in the future.

components/ambient-control-plane/internal/kubeclient/namespace_provisioner.go
```diff
@@ -43,41 +43,17 @@ func NewStandardNamespaceProvisioner(kube *KubeClient, logger zerolog.Logger) *S
 
 func (p *StandardNamespaceProvisioner) ProvisionNamespace(ctx context.Context, name string, labels map[string]string) error {
 	if _, err := p.kube.GetNamespace(ctx, name); err == nil {
-		p.logger.Debug().Str("namespace", name).Msg("namespace already exists")
+		p.logger.Debug().Str("namespace", name).Msg("namespace exists")
 		return nil
-	} else if !k8serrors.IsNotFound(err) {
+	} else if k8serrors.IsNotFound(err) {
+		return fmt.Errorf("namespace %s does not exist and must be created externally", name)
```
squizzi: This is going to break current ACP -- are we ok with that?  This should be feature gated behind openshell OR some other config option like `NAMESPACES_MANAGED_EXTERNALLY` or something.

components/ambient-control-plane/internal/kubeclient/namespace_provisioner.go
```diff
@@ -43,41 +43,17 @@ func NewStandardNamespaceProvisioner(kube *KubeClient, logger zerolog.Logger) *S
...
-
-	if _, err := p.kube.CreateNamespace(ctx, ns); err != nil && !k8serrors.IsAlreadyExists(err) {
-		return fmt.Errorf("creating namespace %s: %w", name, err)
-	}
-
-	p.logger.Info().Str("namespace", name).Msg("namespace created")
-	return nil
 }
 
 func (p *StandardNamespaceProvisioner) DeprovisionNamespace(ctx context.Context, name string) error {
```
squizzi, line 81: This doesn't do anything anymore and should just be removed and any calls to it removed.

If we bring back `ProvisionNamespace` we should bring this back to and feature flag the removal and keep the debug log for now until we fully deprecate managed namespaces.

components/ambient-control-plane/internal/openshell/grpc/openshell/datamodel/v1/datamodel.pb.go
```diff
@@ -0,0 +1,268 @@
...
+//
+// Vendored from github.com/NVIDIA/OpenShell/proto/datamodel.proto
+
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.36.11
+// 	protoc        (unknown)
+// source: openshell/datamodel/v1/datamodel.proto
+
+package datamodel_v1
```
squizzi, line 12: Not blocking, but for future: We should come up with a better way to use upstream proto files downstream.  We're essentially copying in proto files and then rolling our own client, all of this stuff is probably figured out in the openshell CLI and we could just use it's client.

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -168,6 +194,13 @@ func (r *SimpleKubeReconciler) Reconcile(ctx context.Context, event informer.Res
 }
 
 func (r *SimpleKubeReconciler) provisionSession(ctx context.Context, session types.Session) error {
+	if r.cfg.OpenShellUseGateway {
+		return r.provisionSessionGateway(ctx, session)
```
squizzi:
```suggestion
		return r.provisionSessionSandbox(ctx, session)
```

This is the method to create sandboxes it should be a 1:1 to pods, the gateway is already deployed and ACP doesn't manage it yet

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -230,7 +263,326 @@ func (r *SimpleKubeReconciler) provisionSession(ctx context.Context, session typ
 	return nil
 }
 
+func (r *SimpleKubeReconciler) provisionSessionGateway(ctx context.Context, session types.Session) error {
```
squizzi:
```suggestion
func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, session types.Session) error {
```

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -230,7 +263,326 @@ func (r *SimpleKubeReconciler) provisionSession(ctx context.Context, session typ
...
+			}
+
+			r.logger.Info().
+				Str("sandbox", sbxName).
+				Str("sandbox_id", sandboxID).
+				Msg("sandbox is ready, executing command")
+
+			result, err := r.gateway.ExecSandbox(ctx, namespace, &openshellpb.ExecSandboxRequest{
+				SandboxId: sandboxID,
+				Command:   []string{"echo", "hello world from ACP"},
```
squizzi, line 409: 👋

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -230,7 +263,326 @@ func (r *SimpleKubeReconciler) provisionSession(ctx context.Context, session typ
...
+			Str("namespace", namespace).
+			Msg("gateway provider created")
+
+		providerNames = append(providerNames, provName)
+	}
+
+	return providerNames, nil
+}
+
+func (r *SimpleKubeReconciler) buildGatewayEnv(ctx context.Context, session types.Session, projectName string, sdk *sdkclient.Client, providerNames []string) map[string]string {
```
squizzi:
```suggestion
func (r *SimpleKubeReconciler) buildSandboxEnv(ctx context.Context, session types.Session, projectName string, sdk *sdkclient.Client, providerNames []string) map[string]string {
```

This looks like the sandbox environment, again AI has conflated the gateway/sandbox terminology, there should be a clear delineation or it's going to be confusing.

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -256,7 +608,38 @@ func (r *SimpleKubeReconciler) deprovisionSession(ctx context.Context, session t
 	return nil
 }
 
+func (r *SimpleKubeReconciler) deprovisionSessionGateway(ctx context.Context, session types.Session, nextPhase string) error {
```
squizzi:
```suggestion
func (r *SimpleKubeReconciler) deprovisionSessionSandbox(ctx context.Context, session types.Session, nextPhase string) error {
```

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -256,7 +608,38 @@ func (r *SimpleKubeReconciler) deprovisionSession(ctx context.Context, session t
...
+		}
+	}
+
+	r.updateSessionPhase(ctx, session, nextPhase)
+	return nil
+}
+
 func (r *SimpleKubeReconciler) cleanupSession(ctx context.Context, session types.Session) error {
+	if r.cfg.OpenShellUseGateway {
+		return r.cleanupSessionGateway(ctx, session)
```
squizzi:
```suggestion
		return r.cleanupSessionSandbox(ctx, session)
```

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -296,6 +679,46 @@ func (r *SimpleKubeReconciler) cleanupSession(ctx context.Context, session types
 	return nil
 }
 
+func (r *SimpleKubeReconciler) cleanupSessionGateway(ctx context.Context, session types.Session) error {
```
squizzi:
```suggestion
func (r *SimpleKubeReconciler) cleanupSessionSandbox(ctx context.Context, session types.Session) error {
```

components/ambient-control-plane/proto/VENDOR.md
```diff
@@ -0,0 +1,25 @@
...
+
+```bash
+cd components/ambient-control-plane/proto && buf generate
+```
+
+Output lands in `internal/openshell/grpc/`.
+
+## Updating from upstream
+
+1. Copy the needed `.proto` files from the upstream `proto/` directory at the target tag.
```
squizzi: We should have a script to perform the copy and then perform the other steps in this with the ability to pin to a specific commit hash.  Place the commit sha in the copied files as a comment at the top of the files and wrap this all up into a `make generate` target or it's going to get unruly real fast.  AI should be able to one-shot a script for this.

skills/build/dev-cluster/SKILL.md
```diff
@@ -10,627 +10,213 @@ description: >
...
+| Create with local images | `make kind-up LOCAL_IMAGES=true` |
 | Rebuild all | `make kind-rebuild` |
+| Reload single component | `make kind-reload-ambient-control-plane` |
 | Check status | `make kind-status` |
-| View logs | `kubectl logs -f -l app=backend -n ambient-code` |
+| View logs | `kubectl logs -f -l app=<label> -n ambient-code` |
 | Tear down | `make kind-down` |
-| Access URL | `make kind-status` for ports (`$KIND_FWD_FRONTEND_PORT`) |
-
-## When to Invoke This Skill
```
squizzi, line 565: Should we remove all of this?

squizzi (changes requested):

jsell-rh (approved):
🤖 **Amber Re-Review** — 1 new commit since last approval (`db9a89ae` → `734de93e`)

**✅ `734de93e` — `feat: dual-tenant openshell gateway provisioning (tenant-a, tenant-b)`**

Reviewed `Makefile`, `scripts/setup-kind-openshell.sh`, and `tests/openshell-dual-tenant.sh`.

**Makefile** — clean rename: `OPENSHELL_TENANT_NAMESPACE` → `OPENSHELL_TENANTS`, new `test-openshell-dual-tenant` target, pass-through to script correct ✅

**setup-kind-openshell.sh:**
- `IFS=' ' read -ra TENANTS <<< "${OPENSHELL_TENANTS:-...}"` — correct bash array from space-separated string ✅
- Deployment name fix: `agent-sandbox-controller` (was `agent-sandbox-controller-manager`) — correct ✅
- Per-tenant Helm loop with idempotent guard (`helm status` before install) ✅
- ClusterRole/ClusterRoleBinding re-annotation before each install — pragmatic workaround for Helm cluster-scoped resource conflicts across releases in different namespaces ✅
- ACP project creation via temporary port-forward: `test-user-token` retrieval graceful (empty string fallback), port-forward cleanup via `trap`, 15s readiness wait, idempotent project creation ✅
- No secrets logged ✅

**tests/openshell-dual-tenant.sh:**
- `set -euo pipefail`, StatefulSet check, CRD name consistent with setup script fixes ✅
- Concurrent session creation via background subshells with per-tenant output files ✅
- Section 5 uses `skip` (not `fail`) when gateway is healthy but no sandbox CR visible — acknowledges sandbox may be internal to gateway ✅
- Cleanup deletes test sessions ✅

**Note:** `squizzi` posted `CHANGES_REQUESTED` at `db9a89ae` — this new commit adds features rather than addressing that review. Worth verifying squizzi's concerns are addressed separately before promoting.

**Re-approval confirmed** for the new dual-tenant commit. Implementation is correct, resource names fixed, setup is idempotent.

*— Amber*

components/ambient-control-plane/internal/kubeclient/namespace_provisioner.go
bsquizz: Pushed change to ensure that when OPENSHELL_USE_GATEWAY=true, ns provisioning and de-provisioning does not occur. We check for ns existence and that's it.

skills/build/dev-cluster/SKILL.md
bsquizz: The skill was compacted because we have a gate in place indicating that skills shouldn't be more than 500 chars

components/ambient-control-plane/internal/kubeclient/namespace_provisioner.go
bsquizz: Pushed a commit.

The change restores the original behavior for StandardNamespaceProvisioner:

  - ProvisionNamespace: update-or-create — updates labels on existing namespaces, creates if absent
  - DeprovisionNamespace: actually deletes the namespace

  This only affects the non-gateway path. When `OPENSHELL_USE_GATEWAY=true`, provisionSessionGateway does a direct GetNamespace check and cleanupSessionGateway skips DeprovisionNamespace entirely — the provisioner isn't involved at all.

skills/build/dev-cluster/SKILL.md
bsquizz: I'll see if we removed valuable information here for things involving OPENSHELL_USE_GATEWAY=false

jsell-rh (approved):
🤖 **Amber Re-Review** — 2 new commits since last approval (`734de93e` → `3f66d523`, `48ae2ddc`)

## New Commits Reviewed

### ✅ `3f66d523` — `fix(control-plane): disable namespace provision/deprovision in gateway mode`

**`provisionSessionGateway` change** — correct and clean:
- Replaces `ensureNamespaceExists` (which invoked the provisioner) with a direct `GetNamespace` call ✅
- `k8serrors.IsNotFound` handled explicitly → descriptive error with no provisioner side-effects ✅
- Non-404 errors correctly wrapped: `fmt.Errorf("checking namespace %s: %w", namespace, err)` ✅
- Provisioner is bypassed entirely — no `StandardNamespaceProvisioner` or `MPPProvisioner` can touch the gateway namespace on this path ✅

**`cleanupSessionGateway` change** — removes `DeprovisionNamespace` call ✅
- Gateway-managed namespaces contain infrastructure (Helm-installed gateway, TLS secrets) that ACP must not destroy; the removal is correct ✅

**Spec additions** — the two new scenarios (`Gateway mode — direct namespace verification` and `Gateway mode — no DeprovisionNamespace on cleanup`) are accurate, complete, and match the implementation exactly ✅

---

### ✅ `48ae2ddc` — `fix(control-plane): restore StandardNamespaceProvisioner create/delete in pod mode`

**`ProvisionNamespace` update-or-create pattern** — correct:
- `Get → Update (with ResourceVersion) if found, Create if not` — proper K8s update-or-create; setting `ResourceVersion` before Update is required ✅
- `!k8serrors.IsNotFound(err)` guard on the get-error path ✅
- Both `Update` and `Create` errors wrapped: `fmt.Errorf("updating/creating namespace %s: %w", ...)` ✅
- Labels applied on every reconcile (not just creation) — fixes the silent-skip issue that was present before ✅

**`DeprovisionNamespace` restoration** — correct:
- `DeleteNamespace` with `!k8serrors.IsNotFound(err)` guard — idempotent, safe on repeated cleanup ✅
- This is scoped to pod mode only; gateway mode bypasses the provisioner entirely, so there's no cross-mode regression risk ✅

**Spec updates** — mode split in the Namespace Lifecycle section is clear, both pod-mode scenarios accurately reflect the implementation, migration table updated ✅

---

## Note: `squizzi` CHANGES_REQUESTED still open

`squizzi` posted `CHANGES_REQUESTED` at `db9a89ae`. These two new commits add significant behavior changes beyond what squizzi reviewed. Worth checking whether squizzi's concerns from that review are addressed by `3f66d523` / `48ae2ddc`, or require a follow-up.

---

**Re-approval confirmed** for commits `3f66d523` and `48ae2ddc`. The mode split between gateway and pod provisioning is correctly implemented and well-specified.

**Confidence:** High — logic is straightforward, error handling is complete, gateway and pod paths are cleanly separated.

*— Amber*

components/ambient-control-plane/proto/VENDOR.md
bsquizz: Pushed as 483238cc. Here's a summary of what was implemented:

  - components/ambient-control-plane/scripts/vendor-proto.sh — single command to update OpenShell protos: resolves tags to commit SHAs via the
  GitHub API, fetches each proto file, strips and rewrites the SPDX header (stamping the exact SHA), injects the local go_package option,
  updates VENDOR.md, and runs buf generate.
  - make vendor-openshell-proto REF=v0.0.71 — root Makefile target under a new ##@ Code Generation section that calls the script.
  - proto/VENDOR.md — updated to document the script as the canonical update path, replacing the 4-step manual process. Kept the buf generate
  one-liner for stub-only regeneration.

skills/build/dev-cluster/SKILL.md
bsquizz: Pushed a commit for this.

  skills/build/dev-cluster/SKILL.md — 6 targeted restorations (222 → 347 lines):
  - .env.test credentials note in cluster lifecycle
  - Step 4 expanded with the full retry loop (exit-code-7 vs HTTP error distinction)
  - Step 5 access info template (only after frontend validates)
  - When to use / When NOT to use in the fast inner-loop section
  - "Always pass CONTAINER_ENGINE=" guidance + new "Detecting the Access URL" section
  - Common Tasks copy-paste recipes 
  - Decision tree in Quick Reference

  skills/build/openshell-gateway/ — new skill (183 lines + evals):
  - make kind-up OPENSHELL_USE_GATEWAY=true flow and what the setup script does
  - Verification checklist (gateway StatefulSets, CRD, CP flag, logs)
  - make test-openshell-dual-tenant usage and what the 5 test sections cover
  - Reload workflow (env vars survive kind-reload-*)
  - Troubleshooting: gateway pod startup, missing sandbox CRs, Helm RBAC conflicts, mode switching
  - Reference table (make targets + key files)

jsell-rh (approved):
🤖 **Amber Re-Review** — 3 new commits since last approval (`71d8b62f` → `08f75e86`, `885bf209`, `419be00d`)

## New Commits Reviewed

### ✅ `08f75e86` — `refactor(control-plane): rename *Gateway session methods to *Sandbox`

Pure rename refactor — `provisionSessionGateway` → `provisionSessionSandbox`, `cleanupSessionGateway` → `cleanupSessionSandbox`, `buildGatewayEnv` → `buildSandboxEnv`. 16 lines changed, no logic touched.

This was squizzi's suggestion and it's the right call — the old names conflated the gateway infrastructure with the sandboxes it manages. The new names correctly reflect that ACP provisions and cleans up sandboxes (via the gateway), not the gateway itself. All call sites updated consistently. ✅

---

### ✅ `885bf209` — `feat(control-plane): add vendor-proto.sh script and make target for OpenShell proto updates`

Addresses squizzi's feedback about the manual 4-step proto update process being unruly.

**Script (`scripts/vendor-proto.sh`):**
- `set -euo pipefail` ✅
- Tag → commit SHA resolution via GitHub API (handles both lightweight and annotated tags correctly — peels annotated tags through the tag object) ✅
- Python transform: strips upstream SPDX header and `go_package`, injects local header with upstream path + commit SHA, local `go_package` — clean and correct ✅
- SHA stamped in each file header for exact reproducibility ✅
- `trap 'rm -f ...' EXIT` cleanup of temp files ✅
- `die()` / `require_cmd()` helpers — proper error handling ✅
- `VENDOR.md` updated programmatically (regex-based, handles tag-only and SHA-only runs, idempotent re-run) ✅
- Runs `buf generate` at the end ✅
- Curated subset note preserved for `openshell.proto` ✅

**Note:** `VENDOR.md` shows `Vendored commit | a1b2c3d4e5f6` (placeholder SHA). Correct after the first real `make vendor-openshell-proto REF=<tag>` run — cosmetic only, not a blocker.

**Makefile target:** `vendor-openshell-proto` under new `##@ Code Generation` section, mandatory `REF=` guard, correct script invocation. ✅

---

### ✅ `419be00d` — `docs(skills): restore lost dev-cluster content, add openshell-gateway skill`

Addresses squizzi's question about whether content was lost from `dev-cluster/SKILL.md` during gateway work.

**dev-cluster/SKILL.md restorations (6 sections):**
- `.env.test` credentials note in cluster lifecycle ✅
- Step 4 expanded with retry loop distinguishing `exit code 7` (port-forward dead) from HTTP errors (app unhealthy) — this distinction is genuinely valuable ✅
- Step 5 access info template gated behind frontend validation ✅
- When/When-not guidance for the fast inner-loop ✅
- `CONTAINER_ENGINE=` pass-through guidance and `Detecting Access URL` section ✅
- Common Tasks copy-paste recipes and decision tree ✅

**openshell-gateway/SKILL.md (new skill, 183 lines):**
- Trigger keywords cover expected invocation patterns ✅
- Prerequisites clearly stated (`helm`, `oci://ghcr.io/nvidia/openshell/helm-chart`) ✅
- Setup flow explains what `setup-kind-openshell.sh` does (5 steps) ✅
- Verification checklist matches the actual resources created by `734de93e` ✅
- E2E test usage with all 5 sections documented ✅
- Reload workflow correctly notes that env vars survive `kind-reload-*` ✅
- Troubleshooting: gateway pod startup, missing sandbox CRs, Helm RBAC conflict, mode switching ✅
- Reference table cross-links `make vendor-openshell-proto` — good cohesion with `885bf209` ✅

**evals.json:** 4 eval cases covering primary trigger patterns. ✅

---

## Open Threads (context only — not blocking)

- **`PRRT_kwDOS2zRNs6MiUA2`**: env var vs. Secret for mTLS credentials — in-flight discussion, no action needed for this PR.
- **`PRRT_kwDOS2zRNs6MkrEv`**: squizzi's 👋 at `kube_reconciler.go:409` — looks like an unresolved comment on an outdated diff, no action required.
- **`PRRT_kwDOS2zRNs6Mkku5`**: future work note on using upstream OpenShell CLI client instead of vendored protos — acknowledged, not blocking.

squizzi's prior `CHANGES_REQUESTED` concerns are all resolved: namespace provisioner behavior is correctly mode-split, method names corrected, proto vendoring has a reproducible script. ✅

---

**Re-approval confirmed.** All three new commits are clean, correct, and directly address prior review feedback.

*— Amber*

squizzi (approved):

PR #188, opened 2026-06-26
feat(specs): add gateway provisioning specification

JuanmaBM:
Adds formal specification for ACP gateway provisioning capability. This closes the "future iteration" gap from `openshell-sandbox-provisioning.spec.md` line 20-22, where gateways were assumed pre-installed.

With this spec, ACP will manage gateway lifecycle across project namespaces via a platform-config ConfigMap, replacing the manual Helm installation workflow.

---

## Context

**From team meeting (2026-06-25):** Gateway deployment should be ACP's responsibility, not app-interface. Namespaces and credentials remain app-interface managed, but the gateway itself (versioning, updates, configuration) needs centralized control.

**Current state:** Gateways must be manually deployed per namespace. No automated updates, no version consistency guarantees across tenants.

**After this spec:** Add namespace to platform-config → gateway auto-deploys. Update ACP → gateways update (if drift detection implemented).

---

### Infrastructure-layer, not API
- Gateway state lives in ConfigMap only (NOT PostgreSQL)
- No `/api/v1/gateways` endpoints
- Rationale: Gateways are cluster infrastructure, not business entities

### Backward compatible
- StandardNamespaceProvisioner behavior unchanged (still creates namespaces)
- No breaking changes to existing deployments
- Enabled only when platform-config ConfigMap exists

### Flexibility for MVP
- Reconcile trigger: watch OR polling (implementation decides)
- Drift detection: SHOULD detect, create-only acceptable (matches current pod pattern)
- RBAC: conceptual requirements, exact ClusterRole TBD

### Consistency with existing patterns
- NO OwnerReferences (same as session pods today)
- Cleanup via labels (same as sessions)
- SecurityContext + resource limits (same as all containers)


jsell-rh (approved):
🤖 **Amber Review — Approved**

Pure spec PR — no production code to check against Go/TS checklists. Well-structured, complete, and consistent with ACP architecture patterns.

**Summary:** This is solid design work. Approving with three notes the author should either address or explicitly call out as deferred. Nothing here blocks the spec from landing.

<details>

**Strengths:**
- Backward-compatible activation (enabled only when `platform-config` ConfigMap exists) ✅
- SecurityContext requirements fully specified: `runAsNonRoot: true`, `allowPrivilegeEscalation: false`, `drop: [ALL]` ✅
- Resource limits defined: `100m/256Mi` requests, `500m/512Mi` limits ✅
- `managed-by=ambient-control-plane` labels for lifecycle tracking ✅
- No OwnerReferences — explicitly justified and consistent with existing session pod pattern ✅
- Separate reconciler; `kube_reconciler.go` and `StandardNamespaceProvisioner` explicitly unchanged ✅
- Idempotency scenario specified (Service label detection → skip create-duplicate) ✅
- Partial failure handling: one namespace failing doesn't block others ✅
- `helm template` pre-generation pattern (not Helm at runtime) is the right call ✅
- Drift detection correctly marked SHOULD with create-only as acceptable fallback ✅
- Migration section cleanly supersedes the prior spec's "future iteration" constraint ✅

---

**Note 1 — Design gap: Namespace removal from ConfigMap (worth calling out)**

The spec covers:
- Namespace added to ConfigMap → deploy gateway ✅
- Namespace missing from cluster → skip ✅

But not covered: **namespace removed from ConfigMap while its gateway is still deployed**.

Without a deprovisioning scenario, gateways will accumulate indefinitely as namespaces are rotated through the platform. If the intent is "ACP never cleans up gateways, label-based cleanup is external," that should be stated explicitly — otherwise implementers will have to invent the behavior. Either add a scenario or add a note to Future Work.

---

**Note 2 — Discovery label scope**

The discovery check uses `app.kubernetes.io/name=openshell` alone. If a namespace has an externally-managed OpenShell Service (deployed by a tenant, not ACP), ACP will consider the gateway "already deployed" and skip it — potentially leaving an outdated or misconfigured gateway in place.

Adding `app.kubernetes.io/managed-by=ambient-control-plane` to the discovery label selector would distinguish ACP-managed gateways from tenant-installed ones. Worth specifying whether the discovery is intentionally broad (any OpenShell = skip) or ACP-scoped.

---

**Note 3 — `platform-config` name and Helm chart update workflow**

Two minor things worth one sentence each in the spec:
1. `platform-config` is a generic name — if App Interface or other tooling uses the same name in `ambient-code`, there's a silent conflict risk. If the name is intentional and coordinated, a note to that effect would help future readers.
2. Manifests are pre-generated via `helm template`. When the upstream Helm chart updates (e.g., gateway v0.0.71), who re-runs `helm template` and commits the updated manifests? A one-line note on the update workflow would give implementers clear guidance.

---

**Confidence:** High — spec-only, no runtime behavior to verify. The architectural decisions (ConfigMap over PostgreSQL, no OwnerReferences, separate reconciler) are all correct for infrastructure-layer capabilities.

</details>

*— Amber*

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+- AND ACP SHALL NOT crash or enter a crash loop
+
+#### Scenario: Namespace deleted at runtime
+
+- GIVEN ACP is managing `tenant-alpha` with a deployed gateway
+- AND `platform-config` still lists `tenant-alpha`
+- WHEN an admin deletes the namespace from the cluster
+- THEN ACP SHALL detect the missing namespace in the next reconcile cycle
+- AND ACP SHALL log a warning "namespace tenant-alpha listed in config but not found in cluster"
+- AND ACP SHALL skip that namespace until it reappears or is removed from the ConfigMap
```
squizzi, line 98: Currently ACP manages tenant namespaces and it reconciles them if they get deleted.  Do we intend to move namespace creation entirely off ACP's plate and into a platform reconcilers responsibility?

For example:

```
{"level":"info","reconciler":"kube","session_id":"3FgDBk1n3oOLT9vX4SE5OJvYLeA","namespace":"hello","time":"2026-06-26T15:28:04Z","message":"provisioning session"}
{"level":"info","provisioner":"standard","namespace":"hello","time":"2026-06-26T15:28:04Z","message":"namespace created"}
{"level":"info","reconciler":"kube","namespace":"hello","time":"2026-06-26T15:28:04Z","message":"namespace provisioned for session"} <--
{"level":"warn","reconciler":"kube","error":"creating network policy allow-ambient-api-server in hello: networkpolicies.networking.k8s.io is forbidden: User \"system:serviceaccount:ambient-code:ambient-control-plane\" cannot create resource \"networkpolicies\" in API group \"networking.k8s.io\" in the namespace \"hello\"","namespace":"hello","time":"2026-06-26T15:28:04Z","message":"failed to ensure api-server network policy"}
```

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+
+---
+
+### Requirement: Gateway Manifest Loading
+
+The control plane SHALL load gateway resource manifests from its container filesystem. Gateway manifests SHALL be stored in the ACP codebase and packaged into the ACP container image.
```
squizzi, line 107: Speaking of images, do we intend to build our own gateway image based on openshell's or just always pull in theirs? I imagine we'll just use theirs and we can potentially pin it.

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+
+**Note:** Drift detection MAY use template hash annotations, image tag comparison, resource spec comparison, or other mechanisms. The create-only pattern is acceptable for initial iterations.
+
+---
+
+### Requirement: Dynamic Configuration Updates
+
+The control plane SHALL detect changes to the platform-config ConfigMap and reconcile gateway state accordingly.
+
+When a new namespace is added to the platform configuration, the control plane SHALL deploy a gateway to the new namespace.
```
squizzi, line 241: This speaks to namespace configuration but what about configuration of the gateway itself?  Let's make sure that we cover configuration updates to gateways.  It'd be nice to be able to roll out a gateway config change without having to do a full release.  For the first iteration of this, I don't know if thats needed but perhaps referencing it as a future item in the spec will be helpful.

squizzi (commented):
Nothing really blocking, just some comments to perhaps give agents a bit more understanding of requirements and how things currently work.

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+**Related:** `openshell-sandbox-provisioning.spec.md` — gateway mode usage; `control-plane.spec.md` — CP reconciliation patterns  
+**Skill:** `skills/build/full-stack-pipeline/` — wave-based implementation pipeline
+
+---
+
+## Purpose
+
+The control plane SHALL provision and reconcile OpenShell gateway deployments in project namespaces. This replaces the assumption in `openshell-sandbox-provisioning.spec.md` (Iteration 1) that gateways are pre-installed. The control plane discovers which project namespaces to manage via a platform configuration ConfigMap and applies gateway resource manifests using Kubernetes APIs.
+
+Gateway provisioning is an **infrastructure-layer capability**. Gateway state is NOT stored in PostgreSQL; the platform-config ConfigMap is the source of truth for which namespaces require gateways.
```
bsquizz, line 14: @jsell-rh @markturansky do you have any opinion on this? Should we periodically read ConfigMaps and store state in our DB, and use our DB as the 'source of truth'... or should we just always read straight from the ConfigMap?

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+
+#### Scenario: Platform ConfigMap malformed
+
+- GIVEN the `platform-config` ConfigMap exists
+- AND the `namespaces` key contains invalid YAML
+- WHEN ACP attempts to parse the ConfigMap
+- THEN ACP SHALL log an error with parse failure details
+- AND ACP SHALL NOT proceed with gateway deployments
+- AND ACP SHALL wait for the ConfigMap to be corrected
+
```
bsquizz, line 56: Let's add another scenario for 'platform configmap updated'. ACP should watch the ConfigMap for changes. If it changes, it shall update its memory.

specs/platform/gateway-provisioning.spec.md
bsquizz: I modified that 'deprovision' functionality in my openshell gateway PR, because it caused the gateway config to disappear (no bueno)

I think we need to move to "the namespace existing on the cluster determines whether the project exists" and then ACP will not be in charge of provisioning or completely de-provisioning a namespace anymore.

The SRE tooling is going to be the thing that creates the NS... so in a dev env, we should probably just expect the NS to be set up by "external methods"

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+#### Scenario: Namespace deleted at runtime
+
+- GIVEN ACP is managing `tenant-alpha` with a deployed gateway
+- AND `platform-config` still lists `tenant-alpha`
+- WHEN an admin deletes the namespace from the cluster
+- THEN ACP SHALL detect the missing namespace in the next reconcile cycle
+- AND ACP SHALL log a warning "namespace tenant-alpha listed in config but not found in cluster"
+- AND ACP SHALL skip that namespace until it reappears or is removed from the ConfigMap
+
```
bsquizz, line 102: Let's add something to consider the UI behavior as well: If a namespace exists in platform config, but is missing from the cluster, we will need to indicate that the namespace is in "error state" on the UI.

specs/platform/gateway-provisioning.spec.md
bsquizz: Ooof good point, we might need to build our own eventually, using konflux. We can change the image used for it in the manifest, sure.

I am thinking we might want to store the gateway configuration in the platform-config ConfigMap too?

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+---
+
+### Requirement: Gateway Manifest Loading
+
+The control plane SHALL load gateway resource manifests from its container filesystem. Gateway manifests SHALL be stored in the ACP codebase and packaged into the ACP container image.
+
+#### Scenario: Load gateway manifests from filesystem
```
bsquizz, line 123: Same question as above: should we store these definitions in a ConfigMap, or are we going to bake them as volumeMounts into the ACP deployment itself?

If we go the ConfigMap route, then we could change the gateway by simply modifying the ConfigMap (instead of rolling out a new ACP PR)

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+- Deployment: `openshell-gateway`
+- Service: `openshell-gateway`
+- ServiceAccount: `openshell-gateway`
+- Role: `openshell-gateway`
+- RoleBinding: `openshell-gateway`
+
+All gateway resources SHALL carry the following labels:
+- `app.kubernetes.io/name=openshell`
+- `app.kubernetes.io/component=gateway`
+- `app.kubernetes.io/managed-by=ambient-control-plane`
```
bsquizz:
```suggestion
- `app.kubernetes.io/managed-by=agent-control-plane`
```

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+
+The gateway Deployment SHALL specify:
+- **SecurityContext:** `runAsNonRoot: true`, `allowPrivilegeEscalation: false`, capabilities `drop: [ALL]`
+- **Resource requests:** `cpu: 100m`, `memory: 256Mi`
+- **Resource limits:** `cpu: 500m`, `memory: 512Mi`
+
```
bsquizz, line 162: It also needs to include the config that this `--set` adds into the config:

```
helm install openshell-gateway oci://ghcr.io/nvidia/openshell/helm-chart \
  --namespace <tenant-namespace> \
  --set "pkiInitJob.serverDnsNames={openshell-gateway.<tenant-namespace>.svc.cluster.local}"
```

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+- AND the gateway Service SHALL be created with label `app.kubernetes.io/name=openshell`
+- AND the ServiceAccount and RBAC resources SHALL be created
+
+#### Scenario: Gateway already exists (idempotency)
+
+- GIVEN `tenant-alpha` has an OpenShell gateway already deployed
+- AND `platform-config` lists `tenant-alpha`
+- WHEN ACP reconciles gateways
+- THEN ACP SHALL detect the existing gateway via Service label `app.kubernetes.io/name=openshell`
+- AND ACP SHALL NOT create duplicate resources
```
bsquizz, line 183:
```suggestion
- AND ACP SHALL apply the latest gateway configurations using client-go Server-Side Apply (SSA) or equivalent
- AND ACP SHALL NOT create duplicate resources
```

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+- WHEN ACP reconciles gateways
+- THEN ACP SHALL detect the inconsistent state
+- AND ACP SHALL create the missing Deployment
+- AND ACP SHALL log a warning "inconsistent state detected: Service exists without Deployment"
+
+---
+
+### Requirement: Gateway Version Updates
+
+When the gateway manifests in the ACP codebase are updated, the control plane SHOULD detect drift between deployed gateway resources and the new manifests, and update the deployed resources accordingly.
```
bsquizz:
```suggestion
When the gateway manifests in the ACP codebase are updated, the control plane SHOULD detect drift between deployed gateway resources and the new manifests, and update the deployed resources accordingly using client-go Server-Side Apply (SSA) or similar.
```

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+
+If drift detection is not implemented, the control plane SHALL follow a create-only pattern (existing gateway resources are not modified; manual deletion required to trigger recreation with new manifests). This matches the current behavior for session pods.
```
bsquizz: Lets remove this, we should implement drift detection and not make it "optional"

specs/platform/gateway-provisioning.spec.md
bsquizz: +1, let's make the gateway configuration part of the ConfigMap (unless there's any objections)

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+- WHEN ACP attempts to apply gateway manifests
+- THEN the Kubernetes API SHALL return a Forbidden error
+- AND ACP SHALL log an error "insufficient permissions to create Deployment in namespace tenant-alpha"
+- AND ACP SHALL continue processing other namespaces where it has permissions
+
+---
+
+### Requirement: Platform ConfigMap Default Values
+
+When a namespace in the platform configuration has no additional configuration specified, the control plane SHALL apply the gateway with default values.
```
bsquizz: This section should be changed. Gateway config will always have to be present in the ConfigMap. If it's not there, we will need to error out.

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+
+This specification supersedes the "Gateway provisioning" constraint in `openshell-sandbox-provisioning.spec.md` (line 20-22), which stated:
+
+> "Gateway provisioning — the OpenShell gateway is assumed to already be deployed in each project namespace; ACP will not create it. A future iteration should have the control plane provision and reconcile gateway lifecycle per project namespace..."
+
+This specification IS that future iteration.
+
+### Backward Compatibility
+
+This is a new capability. When disabled (no `platform-config` ConfigMap), ACP behavior is unchanged — it will not attempt to manage gateways.
```
bsquizz, line 341: If OPENSHELL_USE_GATEWAY=true on the ACP configuration, we should provision gateways, and make sure we have the config to do so (and throw an error if we do not)

specs/platform/gateway-provisioning.spec.md
```diff
@@ -0,0 +1,382 @@
...
+
+**Note:** Namespaces MAY be managed by ACP (current behavior) or externally via App Interface. Gateway provisioning works with both models — it only requires that namespaces exist before gateway deployment.
+
+---
+
+## Configuration
+
+### Environment Variables
+
+No new environment variables are required. Gateway provisioning is enabled by the presence of the `platform-config` ConfigMap.
```
bsquizz:
```suggestion
No new environment variables are required. Gateway provisioning is enabled by the presence OPENSHELL_USE_GATEWAY=true
```

specs/platform/gateway-provisioning.spec.md
squizzi: > store the gateway configuration in the platform-config ConfigMap too?

Potentially, it'll make config rollouts easier and more of an SRE specific task which I think is the right call.  We won't have to push an image update for a config change.

specs/platform/gateway-provisioning.spec.md
squizzi: fyi https://github.com/openshift-online/agent-control-plane/pull/158 added some error state statuses to the UI.  Potentially worth referencing as context for implementation

jsell-rh:
🤖 **Amber Review**

Well-structured gateway provisioning spec. No production code, so checklist is spec-layer only. Design status is appropriate — this is clearly scoped and internally consistent with existing ACP patterns.

**Summary:** LGTM as a Design spec. Three notes before this moves to Proposed.

<details>

**Strengths:**
- Gateway provisioning gated on `platform-config` ConfigMap presence — zero behavior change for existing deployments ✅
- SecurityContext specified: `runAsNonRoot`, `allowPrivilegeEscalation: false`, `drop: [ALL]` ✅
- Resource requests and limits specified (`cpu: 100m/500m`, `memory: 256Mi/512Mi`) ✅
- Per-namespace failure isolation — failure in one namespace does NOT block others ✅
- No OwnerReferences — explicitly justified and consistent with session pod pattern ✅
- ConfigMap-only source of truth (no PostgreSQL) — correct call for infra-layer state ✅
- Backward compatibility table is clear and complete ✅
- `platform-config` not found → retry, not crash ✅

---

**Note 1 — Spec gap: namespace removal from ConfigMap**

There's no scenario for what happens when a namespace is *removed* from `platform-config`. Does the gateway keep running? Is there a cleanup step? This is a meaningful operational question: if a tenant is offboarded, their gateway should presumably be decommissioned. Even if the answer is "ACP does NOT clean up gateways on namespace removal — label-based cleanup is out of scope," that should be stated explicitly so implementers don't make assumptions either way.

**Note 2 — "Helm template once" packaging pattern carries drift risk**

> Gateway manifests SHALL be generated once during development using `helm template` (NOT Helm at runtime) and packaged into the ACP container image.

This is a valid pattern, but it means a Helm chart update won't flow through to deployed gateways unless someone regenerates the manifests and commits them. This isn't wrong, but the spec doesn't call out that regeneration is a manual step in the release workflow. A note like "When the upstream Helm chart changes, regenerate manifests via `scripts/vendor-gateway-manifests.sh` and commit the result" would help implementers know this exists in the workflow.

**Note 3 — Idempotency scenario leaves update path ambiguous**

The "Gateway already exists (idempotency)" scenario says:
> ACP SHALL NOT create duplicate resources, AND ACP MAY verify the gateway is healthy

If drift detection is not implemented (create-only pattern), this is correct. But the create-only pattern means a gateway version update requires manual deletion and recreation — implementers should know this explicitly. The "Gateway Version Updates" requirement mentions this, but the idempotency scenario doesn't cross-reference it. A sentence like "See 'Gateway Version Updates' for update behavior when manifests change" would close the loop.

---

**Non-issues:**
- RBAC exact ClusterRole deferred to implementation — fine for Design status
- Watch vs. polling for ConfigMap changes left flexible — fine for Design status

**Confidence:** High. Spec-only, no runtime behavior to verify.

</details>

*— Amber*

specs/platform/gateway-provisioning.spec.md
JuanmaBM: Yeah, that's the idea. I guess it doesn't show up as having been written in the use cases, but the configmap should contain all namespaces, and under each one, the configuration for the gateway in the namespace.

specs/platform/gateway-provisioning.spec.md
JuanmaBM: Chris Mitchel was working on creating a new image for the Gateway, maybe we can use his work. Also the openshell RH team was working on creating RedHat openshell so maybe some of these can help us in order to avoid starting from scratch.

jsell-rh:
🤖 **Amber Re-Review — 4 new commits since `6dd369aa`**

The 4 new commits (`17253ab2`, `55d7e635`, `66dbb81f`, `b7fc5924`) accept four of bsquizz's inline suggestions: the `managed-by=agent-control-plane` label correction, the SSA idempotency wording, the drift detection SHOULD phrasing, and the `OPENSHELL_USE_GATEWAY` environment variable note. All correctly applied.

**However, there are 11 substantive open review threads from bsquizz that the spec hasn't addressed yet.** Before this lands, the spec should incorporate the agreed design direction visible in those threads:

1. **Gateway config in ConfigMap** — bsquizz and squizzi agree the gateway configuration (not just the namespace list) should live in `platform-config`, so config rollouts don't require an ACP image update. JuanmaBM confirmed intent in a comment but the spec still describes a filesystem-first model.
2. **Drift detection required** — bsquizz asked to remove the create-only fallback and make drift detection a SHALL. The spec still marks it SHOULD.
3. **`OPENSHELL_USE_GATEWAY` gates provisioning** — the spec should state this explicitly, with an error if the env var is set but the gateway config is missing.
4. **Namespace lifecycle ownership** — the team is converging toward "namespace existence determines project existence, SRE tooling creates namespaces." The current spec has ACP creating namespaces via `StandardNamespaceProvisioner`, which conflicts with that direction.
5. **DNS names config in manifests** — the `--set pkiInitJob.serverDnsNames` Helm value needs to be represented in the pre-generated manifests.
6. **UI behavior for missing namespaces** — scenario needed for when a namespace is in `platform-config` but not in the cluster.
7. **ConfigMap vs filesystem for manifest storage** — open question whether manifests should be a volume-mounted ConfigMap vs baked into the image.
8. **"platform-config updated" scenario** — bsquizz requested an explicit scenario for ConfigMap watch + in-memory update.

My original approval stands for the spec as a solid starting point, but these design decisions should be reflected before it lands — otherwise implementers will diverge from what the team has actually agreed to.

**Confidence:** High — the open threads are clearly visible and none are nitpicks.

*— Amber*

jsell-rh (commented):
🤖 **Amber Re-Review — 4 New Commits (2026-06-29)**

The 4 new commits today applied several inline suggestions from bsquizz. Here's where things stand:

## What Was Addressed ✅

- `app.kubernetes.io/managed-by=agent-control-plane` label added to gateway resources — this resolves my Note 2 from the original review about discovery label scope.
- SSA (Server-Side Apply) language incorporated into the idempotency scenario.
- Some of the drift detection language was updated.

## Regression — Last Commit

The last commit (`b7fc592`) changed the Environment Variables section from:

> "Gateway provisioning is enabled by the presence of the `platform-config` ConfigMap."

To:

> "Gateway provisioning is enabled by the presence OPENSHELL_USE_GATEWAY=true"

This sentence is grammatically incomplete (missing "of") and also represents a design change — shifting the trigger from ConfigMap presence to the env var. The change aligns with bsquizz's comment in thread PRRT_kwDOS2zRNs6Mk1jq, but the sentence needs to be a complete, clear statement. E.g.: "Gateway provisioning is enabled when `OPENSHELL_USE_GATEWAY=true`."

## Still Open from Original Review

**My Note 1 (namespace removal/deprovisioning)** is still not in the spec or Future Work. If namespace `tenant-alpha` is removed from the ConfigMap while a gateway is deployed, what happens? This needs one sentence in either the spec or Future Work.

## Open Team Threads Needing Spec Updates

Reviewing the unresolved threads, these are substantive enough that the spec should address them before merging:

1. **Drift detection** (PRRT_kwDOS2zRNs6MkxbL) — bsquizz: "Lets remove this, we should implement drift detection and not make it 'optional'". The current spec still has the create-only fallback. If the team agrees drift detection is required, that should be updated.

2. **Gateway config in ConfigMap** (PRRT_kwDOS2zRNs6MkOpt, PRRT_kwDOS2zRNs6MkpLo) — Multiple reviewers converged on storing gateway configuration in the `platform-config` ConfigMap so config changes don't require an ACP image bump. The spec currently shows a fixed manifest packaged in the container. This is a significant architectural point.

3. **Error on missing gateway config** (PRRT_kwDOS2zRNs6Mk0VD) — "Gateway config will always have to be present in the ConfigMap. If it's not there, we will need to error out." The current "Platform ConfigMap Default Values" section says apply defaults, which conflicts with this.

4. **ConfigMap watch scenario** (PRRT_kwDOS2zRNs6MkgJA) — "Add scenario for platform ConfigMap updated" — the spec covers initial load but the dynamic updates section doesn't have a scenario for the ConfigMap itself being watched for changes.

5. **UI error state** (PRRT_kwDOS2zRNs6MklWA) — Missing namespace → UI error state not covered.

6. **pkiInitJob DNS names** (PRRT_kwDOS2zRNs6MktaQ) — The Helm template section needs to document the `--set pkiInitJob.serverDnsNames` requirement.

---

**Confidence:** High — spec-only. The fundamental design is sound, but the spec needs to converge with the team's agreed-upon design direction (ConfigMap-based gateway config, mandatory drift detection) before it's ready to merge.

*— Amber*

jsell-rh (commented):
🤖 **Amber Re-Review — 1 New Commit (bee8a698, 2026-06-29)**

`bee8a698` is a strong update that addresses the majority of open threads from my earlier re-review. Here's the accounting:

## What Was Addressed ✅

- **ConfigMap watch scenario** (PRRT_kwDOS2zRNs6MkgJA) — "Platform ConfigMap updated at runtime" scenario added ✅
- **UI error state** (PRRT_kwDOS2zRNs6MklWA) — namespace not found now includes UI error state with reference to #158 ✅
- **pkiInitJob DNS names** (PRRT_kwDOS2zRNs6MktaQ) — `gateway.serverDnsNames` documented with Helm command reference ✅
- **Drift detection mandatory** (PRRT_kwDOS2zRNs6MkxbL) — SHOULD → SHALL, create-only fallback removed ✅
- **Error on missing gateway config** (PRRT_kwDOS2zRNs6Mk0VD) — "Platform ConfigMap Default Values" → "Platform ConfigMap Gateway Configuration Mandatory" with error scenario ✅
- **Gateway config in ConfigMap** (PRRT_kwDOS2zRNs6MkOpt) — gateway config fields now documented in ConfigMap structure ✅
- **Regression (incomplete sentence)** from my last review — the `OPENSHELL_USE_GATEWAY` sentence now appears grammatically complete in the Environment Variables section ✅

---

## New Issue — Contradiction in Mandatory Gateway Config Requirement

The requirement and the closing note contradict each other:

**Requirement (line ~277):**
> "If gateway configuration is missing, the control plane SHALL log an error and skip that namespace."

**Note at bottom of the ConfigMap section (line ~418):**
> "Note: If gateway configuration is missing for a namespace, ACP **SHALL use default values from the base manifests.** Per-namespace customization allows different tenants to have different gateway versions or settings."

One says error-and-skip; the other says use defaults. The example also shows `tenant-gamma` with no gateway config at all, which conflicts with the mandatory config requirement.

**Needs resolution:** Pick one behavior and apply it consistently. Based on the earlier team discussion ("Gateway config will always have to be present in the ConfigMap. If it's not there, we will need to error out"), the requirement section appears to have the correct intent. The note is a leftover from the old "use defaults" design and should be removed or updated to reflect the new mandatory behavior.

---

## Still Open from Original Review

**Note 1 (namespace deprovisioning)** — what happens when a namespace is removed from `platform-config` while a gateway is deployed? The spec still has no scenario or future-work note for this. The drift detection section now has the scenario for **updates** (upgrade v0.0.70 → v0.0.71), but removal is different: should ACP delete the gateway resources, orphan them, or leave them with a label-based cleanup expectation? One sentence in either the spec or Future Work section would close this out.

---

**Assessment:** One contradiction to fix (mandatory vs. default), one gap still unaddressed (deprovisioning). The spec is otherwise in excellent shape — the mandatory gateway config design is clearly correct, and all the major threading feedback from the team has landed. These are small and targeted.

*— Amber*

jsell-rh:
🤖 **Amber Re-Review — 1 new commit since last comment (`bee8a698`)**

Commit message: "docs: addres PR feedback on gateway provisioning spec"

Good progress — this addresses 5 of the 8 open threads from my 07:37 review. However, there's a direct contradiction introduced in this commit that needs resolution before landing.

---

## ✅ Addressed

1. **`platform-config` updated at runtime scenario** — New scenario added (watch → reload → reconcile). Exactly what was needed. ✅
2. **UI error state for missing namespace** — "namespace not found in cluster" scenario now shows UI error state + reference to #158. ✅
3. **`serverDnsNames` in manifests / Helm command reference** — DNS names config now in the spec with a concrete `helm template` command. ✅
4. **Drift detection `SHOULD` → `SHALL`** — Create-only fallback removed, drift detection is now a requirement. ✅
5. **Gateway config in `platform-config`** — Per-namespace `gateway` stanza (image, serverDnsNames, config) is now in the ConfigMap, not filesystem-only. ✅

---

## ❌ Blocker: Direct contradiction in "Gateway Configuration Mandatory" requirement

The commit introduces a **SHALL** requirement that gateway config is mandatory, then immediately contradicts it with a NOTE that says to use defaults:

**Requirement section (line ~285):**
> If gateway configuration is missing, the control plane SHALL log an error and skip that namespace.

**Note at the bottom of the `platform-config` section:**
> If gateway configuration is missing for a namespace, ACP SHALL use default values from the base manifests.

These two statements directly conflict. Implementers reading this will make different choices.

Additionally, the `tenant-gamma` entry in the example ConfigMap has **no gateway stanza at all**:
```yaml
- name: tenant-gamma
```
…which contradicts the "Gateway Configuration Mandatory" requirement added in this same commit.

**Pick one and be consistent:**
- **Option A (mandatory):** Remove the "use defaults" Note; fix the `tenant-gamma` example to include a gateway stanza
- **Option B (optional with defaults):** Change the Mandatory requirement back to "optional; default values apply when omitted"; remove the error scenario

The spec needs to be unambiguous here — gateway config mandatory vs. optional-with-defaults is an important operational decision (affects onboarding friction and error behavior).

---

## Still open from prior review

- **Namespace lifecycle ownership** (item #4) — whether namespaces are pre-provisioned by SRE tooling vs. created by `StandardNamespaceProvisioner` — not addressed in this commit. Not a blocker for this PR since it's more of an architecture discussion, but worth noting.

---

**Confidence:** High on the contradiction — it's clearly visible in the same commit.

*— Amber*

jsell-rh (commented):
🤖 **Amber Re-Review — 1 New Commit (`7e00e75`)**

The contradiction I flagged in my last review is resolved:

- **"Use defaults" note removed** ✅ — the conflicting "ACP SHALL use default values from the base manifests" note is gone
- **`tenant-gamma` example now has explicit gateway config** ✅ — the example no longer shows a namespace without gateway config, which previously conflicted with the mandatory requirement
- **`gateway.image` default clarified** ✅ — now "defaults to `OPENSHELL_GATEWAY_IMAGE` env var" rather than "base manifest", consistent with the field being optional within a mandatory block

The design is now coherent: `gateway:` block is mandatory; within it, `image` is optional (env var fallback) and `serverDnsNames` is required. No contradictions.

My Note 1 (what happens when a namespace is removed from the ConfigMap while a gateway is still deployed) remains unaddressed in Future Work, but I've flagged it twice as non-blocking and won't re-raise it. The spec is ready.

**Approval stands.**

*— Amber*

bsquizz (approved):

PR #202, opened 2026-06-29
fix(kind): enable unauthenticated users on OpenShell gateway and add e2e tests

bsquizz:
Apparently, we forgot to commit these tests during the feature work.

## Summary
- Fix `setup-kind-openshell.sh` to pass `server.auth.allowUnauthenticatedUsers=true` when installing the OpenShell gateway Helm chart — without this, CreateSandbox RPCs fail with "missing authorization header" even though mTLS transport works correctly
- Add `tests/pod-mode-session.sh` — e2e test for standard pod-mode session provisioning
- Add `tests/run-e2e-both-modes.sh` — orchestrator that runs both pod-mode and gateway-mode e2e tests with full cluster lifecycle (kind-up, test, kind-down for each mode)

## Test plan
- [x] Ran `make kind-up OPENSHELL_USE_GATEWAY=true` with the fix applied
- [x] Verified `allow_unauthenticated_users = true` in gateway config
- [x] Ran `tests/openshell-dual-tenant.sh` — 10/10 passed (sandbox CRs now created successfully in both tenant namespaces)
- [x] Run `tests/pod-mode-session.sh` standalone against a pod-mode cluster
- [x] Run `tests/run-e2e-both-modes.sh` end-to-end

### DIFF
--- a/scripts/setup-kind-openshell.sh
+++ b/scripts/setup-kind-openshell.sh
@@ -67,6 +67,7 @@ for TENANT in "${TENANTS[@]}"; do
     helm install openshell-gateway "$OPENSHELL_GATEWAY_CHART" \
       --namespace "$TENANT" \
       --set "pkiInitJob.serverDnsNames={openshell-gateway.${TENANT}.svc.cluster.local}" \
+      --set "server.auth.allowUnauthenticatedUsers=true" \
       --wait --timeout 120s
     echo "    Installed OpenShell gateway in '$TENANT'"
   fi
--- a/tests/pod-mode-session.sh
+++ b/tests/pod-mode-session.sh
@@ -1,0 +1,195 @@
+#!/usr/bin/env bash
+# E2E test: pod-mode session provisioning (OPENSHELL_USE_GATEWAY=false)
+#
+# Verifies that when operating in standard pod mode, a Kubernetes Pod is
+# created in the project namespace when a session is started.
+#
+# Prerequisites:
+#   - kind-up with OPENSHELL_USE_GATEWAY=false (default)
+#   - TEST_TOKEN set, or e2e/.env.test present
+#
+# Usage:
+#   ./tests/pod-mode-session.sh [API_URL]
+
+set -euo pipefail
+
+NAMESPACE="${NAMESPACE:-ambient-code}"
+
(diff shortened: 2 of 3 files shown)


jsell-rh (approved):
🤖 **Amber Review — PR #202**

Good fix. Adding `server.auth.allowUnauthenticatedUsers=true` to the gateway Helm install is the correct companion to the mTLS-only auth model established in the prior gateway work — the control plane authenticates sandboxes via mTLS at the transport layer, so application-layer auth in the gateway is both redundant and broken by design. The `CONTAINER_ENGINE` fix correctly aligns with CLAUDE.md's podman preference.

Two minor shell scripting issues worth a follow-up (non-blocking for a dev-only test orchestrator):

<details>
<summary>Minor — <code>tests/pod-mode-session.sh</code>: <code>trap cleanup EXIT</code> overwrites the earlier kill trap, leaking the port-forward process</summary>

```bash
# Registered around line 42 — kills the port-forward when the script exits
trap 'kill "${PF_PID}" 2>/dev/null || true' EXIT

# ... then cleanup() is defined and registered around line 83, OVERWRITING the above:
cleanup() {
  # api_delete session, project, namespace — but never kills PF_PID
}
trap cleanup EXIT   # <-- this replaces the kill trap entirely
```

When the script starts its own port-forward (`API_URL` not pre-set), `PF_PID` is populated but the port-forward is never reaped on exit because `cleanup()` doesn't kill it.

Fix: add `kill "${PF_PID}" 2>/dev/null || true` inside `cleanup()` and remove the standalone trap.
</details>

<details>
<summary>Minor — <code>tests/run-e2e-both-modes.sh</code>: port-forward PIDs not covered by <code>teardown()</code></summary>

`$PF1_PID` and `$PF2_PID` are killed inline at the end of each phase, but if the script exits between the port-forward `&` and that kill (e.g., on a failed `if` branch), they leak. The `teardown()` function only calls `make kind-down`. In practice the kind cluster going down kills the port-forwards, but the intent isn't explicit and a future reader may not rely on that guarantee.

One-line fix: add `kill "${PF1_PID:-}" "${PF2_PID:-}" 2>/dev/null || true` into `teardown()`.
</details>

Both are non-blocking. Everything else looks solid — the setup script, the dual-tenant logic, and the phase orchestration in the e2e runner. Approving.

— Amber

PR #206, opened 2026-06-30
[PLATENG-131] spec: gateway mode simplified RBAC policy

squizzi:
Adds specs/security/gateway-rbac-policy.spec.md defining a simplified RBAC policy for gateway mode. Derives ACP tier from Kubernetes namespace RoleBindings. Agent/policy/provider creation restricted to GitOps ConfigMaps. Admin/Editor can start sessions and manage schedules; Viewer is read-only. 10 requirements with 40+ Given/When/Then scenarios.

### DIFF
--- a/specs/security/gateway-rbac-policy.spec.md
+++ b/specs/security/gateway-rbac-policy.spec.md
@@ -1,0 +1,411 @@
+# Gateway Mode Simplified RBAC Policy
+
+**Date:** 2026-06-29
+**Status:** Proposed
+**Related:** `specs/security/rbac-enforcement.spec.md` (base RBAC model), `specs/platform/agent-sandbox-config.spec.md` (ConfigMap agent schema), `specs/platform/gateway-provisioning.spec.md` (gateway deployment), `specs/platform/openshell-sandbox-provisioning.spec.md` (sandbox provisioning flow)
+
+---
+
+## Purpose
+
+When both `OPENSHELL_USE_GATEWAY=true` AND `OPENSHELL_ENABLED=true`, the platform SHALL enforce a simplified RBAC policy that restricts agent, policy, and provider lifecycle management to a GitOps workflow and constrains human users to three effective tiers: Admin, Editor, and Viewer. Agent definitions SHALL be managed exclusively through ConfigMaps applied to tenant namespaces (schema defined in `agent-sandbox-config.spec.md`), and the API SHALL reject agent create, update, and delete operations. Policy and provider declarations (ConfigMaps with labels `ambient.ai/kind: policy` and `ambient.ai/kind: provider`) are already GitOps-only by design — no API endpoints exist for them, and this spec does not introduce any. Existing project members without an explicit Admin or Editor role SHALL receive Viewer-level access. When either flag is false, the system SHALL behave identically to the base RBAC model defined in `rbac-enforcement.spec.md`.
+
+---
+
+## Terminology
+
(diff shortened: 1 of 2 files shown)


jsell-rh (approved):
🤖 **Amber Review — PR #206: Gateway Mode Simplified RBAC Policy**

**Overall:** This is a well-structured spec with strong scenario coverage. The design decisions section is excellent — the rationale for every non-obvious choice is documented, which is exactly the standard this project should hold. A few clarifications worth capturing before implementation begins.

---

**Strengths:**
- Dual-flag AND logic is well-justified and consistent with the existing `OPENSHELL_*` flag pattern
- Handler-level (not middleware-level) gating decision is architecturally sound — keeps RBAC middleware general-purpose
- `platform-info` endpoint design (auth-exempt, runtime-reflecting) avoids the build-time env-var trap
- Backward compatibility requirement is thorough — explicitly covers the "agents survive flag toggle" edge case
- The 403 vs 405 rationale is precisely correct

---

**Observations / Questions for the author:**

**1. SubjectAccessReview mechanism — binding to existing infrastructure needed**

The spec says the API server SHALL use "a Kubernetes `SubjectAccessReview` or equivalent mechanism" to resolve namespace access for the tier mapping. This is intentionally flexible, but before implementation it's worth pinning: does the API server already have a dynamic Kubernetes client for the user's token (via `GetK8sClientsForRequest`)? If yes, the SAR can be issued as that user and the result is accurate. If the check uses the API server's service account instead, it would bypass the per-user authz model and violate the security requirement. Recommend adding a note clarifying that the SAR MUST be issued using `GetK8sClientsForRequest(c)` (user token), not the backend service account — this is a non-negotiable per CLAUDE.md.

**2. `managed-by: configmap` annotation — API mutability boundary**

The spec says the `ambient.ai/managed-by: configmap` annotation "SHALL NOT be modifiable via the API." The enforcement mechanism for this isn't specified. Suggest noting whether this is enforced at the handler level (strip/reject the field on PATCH) or at the DB model level. Without that, an implementation might accidentally allow the annotation to be overwritten via a direct `PATCH /agents/{id}` call, breaking the GitOps identity marker.

**3. Dual-issuer reconciler context**

The GitOps reconciler watches ConfigMaps in tenant namespaces and writes to the API server DB. Which identity does the reconciler use to call the API server? If it uses a service account (as the control plane typically does), it will hit the gateway-mode 403 for agent create/update. Either the reconciler needs a carve-out (e.g., an internal call path that bypasses the gateway-mode gate), or the spec should note that the reconciler writes directly to the DB rather than going through the REST API. This is an implementation detail, but the spec implies the reconciler writes to the DB ("stored in database"), so it's worth making that explicit to avoid confusion.

**4. ConfigMap pruning — partial deletion case**

Scenario "ConfigMap deletion removes the agent" covers full ConfigMap deletion. But ConfigMaps can contain multiple agent declarations as separate keys. The spec says "on ConfigMap deletion, the corresponding Agent record SHALL be deleted." Does pruning also apply when a *key* is removed from a multi-agent ConfigMap, or only when the ConfigMap itself is deleted? `agent-sandbox-config.spec.md` is the authoritative schema source, so this may already be covered there — if so, a forward reference here would prevent implementors from making the wrong assumption.

**5. Minor: index file trailing newline removed**

The diff removes a trailing newline from `index.spec.md`. This is cosmetic but may trigger markdownlint in CI. Not a blocker.

---

**Verdict:** Approving. The spec is implementation-ready for the core requirements. The four observations above are pre-implementation clarifications that will prevent ambiguity when the code is written — recommend addressing in a follow-up comment or a spec amendment before the implementation PR lands.

**Confidence:** High (90%). This is a spec review; no runtime behavior to validate.

*— Amber*

PR #209, opened 2026-06-30
fix(gateway): remove helm chart and fix config override that wiped TLS

bsquizz:
## Summary
- **Remove Helm chart from kind setup** — the control plane reconciler (PR 200) now owns gateway deployment via `platform-config` ConfigMap. The Helm chart was still being installed by `setup-kind-openshell.sh`, and the reconciler was overwriting the Helm-deployed ConfigMap, wiping TLS settings.
- **Remove tenant-a custom config override** — `platform-config.yaml` had a custom `config:` block for tenant-a that replaced the full `gateway.toml` template with a minimal snippet lacking TLS settings, causing `--tls-cert is required when TLS is enabled`. Both tenants now use the complete manifest template.

## Test plan
- [ ] `make kind-up OPENSHELL_USE_GATEWAY=true` — gateways deploy without `--tls-cert` error
- [ ] `openshell sandbox list` works against both tenant gateways
- [ ] `make test-openshell-dual-tenant` passes

### DIFF
--- a/components/manifests/overlays/kind/platform-config.yaml
+++ b/components/manifests/overlays/kind/platform-config.yaml
@@ -10,12 +10,6 @@ data:
         image: ghcr.io/nvidia/openshell/gateway:0.0.71
         serverDnsNames:
           - openshell-gateway.tenant-a.svc.cluster.local
-        config: |
-          [openshell.gateway]
-          bind_address = "0.0.0.0:8080"
-          log_level = "info"
-          sandbox_namespace = "tenant-a"
-          
     - name: tenant-b
       gateway:
         image: ghcr.io/nvidia/openshell/gateway:0.0.71
--- a/scripts/setup-kind-openshell.sh
+++ b/scripts/setup-kind-openshell.sh
@@ -4,15 +4,17 @@
 #
 # Provisions for each tenant in OPENSHELL_TENANTS (default: tenant-a tenant-b):
 #   1. Agent Sandbox CRD + controller (once, cluster-scoped)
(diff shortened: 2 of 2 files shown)


jsell-rh (approved):
🤖 **Amber Review — Approved**

Clean, well-scoped fix. Three distinct issues, three targeted solutions:

1. **Runner image selection** (`kube_reconciler.go`): The `OpenShellUseGateway && OpenShellRunnerImage != ""` guard is correct — requires both the feature flag and a non-empty image before overriding. No regression path for non-gateway mode.

2. **TLS config override** (`platform-config.yaml`): Removing the minimal `config:` block for tenant-a is the right call. The snippet replaced the full `gateway.toml` template (which includes TLS settings) with a 4-line stub missing `tls-cert`, causing the startup failure. Both tenants now use the complete template.

3. **Helm removal** (`setup-kind-openshell.sh`): Deferring gateway deployment entirely to the control plane reconciler eliminates the double-ownership conflict (reconciler was overwriting Helm's ConfigMap). Step renumbering is correct. Helm check and chart-install logic cleanly excised.

**ACP Convention Check:**
- No tokens in logs ✅
- No service account misuse ✅
- No `any` types ✅
- OwnerReferences: N/A (no new K8s child resources) ✅

**Confidence:** High — changes are narrow, causally explained, and the PR body clearly connects each fix to its failure mode.

— *Amber*

jsell-rh:
Adding `amber/approved` label — see review comment above for full analysis.

— *Amber*

PR #213, opened 2026-06-30
fix(gateway): patch resolv.conf ndots and set inference env vars at sandbox level

bsquizz:
## Summary
- Patch `/etc/resolv.conf` to `ndots:1` via `ExecSandbox` before the runner entrypoint starts, fixing the musl libc DNS resolution failure that causes `inference.local` to hang with `NET:FAIL`
- Move `ANTHROPIC_BASE_URL` and `ANTHROPIC_API_KEY` from the Claude-specific wrapper to the sandbox environment so all tools (claude, opencode, etc.) get them automatically
- Remove redundant proxy/TLS env var setup from `auth.py` and the wrapper — the OpenShell supervisor already injects these

## Test plan
- [ ] Rebuild control plane and runner images, deploy to kind with `OPENSHELL_USE_GATEWAY=true`
- [ ] Create a sandbox session and verify `resolv.conf` has `ndots:1`
- [ ] Run `claude` interactively via `openshell sandbox connect` — should reach inference.local without hanging
- [ ] Verify `ANTHROPIC_BASE_URL` and `ANTHROPIC_API_KEY` are present in sandbox env for all processes

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -475,13 +475,34 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 				Msg("sandbox is ready, executing entrypoint")
 
 			execCtx := context.Background()
+
 			// Patch ndots before starting the runner. The OpenShell supervisor is
 			// statically linked with musl libc, whose getaddrinfo sends A+AAAA
 			// queries simultaneously. With Kubernetes' default ndots:5, external
 			// FQDNs get expanded through all search domains first, causing musl's
 			// DNS resolver to mishandle the many concurrent responses and return
-			// zero usable addresses (manifests as 503 "inference service unavailable").
+			// zero usable addresses (manifests as NET:FAIL on inference.local).
 			// Setting ndots:1 makes musl resolve FQDNs directly.
+			//
+			// This must run before the runner entrypoint so the supervisor
+			// picks up the patched resolv.conf on its first DNS lookup for
+			// the upstream inference endpoint.
+			patchResult, patchErr := r.gateway.ExecSandbox(execCtx, namespace, &openshellpb.ExecSandboxRequest{
+				SandboxId: sandboxID,
+				Command:   []string{"sed", "-i", "s/ndots:[0-9]*/ndots:1/", "/etc/resolv.conf"},
+			})
+			if patchErr != nil {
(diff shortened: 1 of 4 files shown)


jsell-rh (commented):
## 🤖 Amber Analysis — PR #213

**Summary:** Clean, well-reasoned refactor that centralizes inference env var management at the sandbox level and moves ndots patching to the right place in the lifecycle. CI is green across all checks.

**Confidence:** High (95%) — the sequencing logic is correct and the code is straightforward.

---

### What's Good

- **Correct fix for the ndots race**: Moving the patch from `entrypoint.sh` (inside the container, after the supervisor's first DNS lookup) to `ExecSandbox` before `ExecSandboxStreaming` ensures the supervisor picks up `ndots:1` before its first inference DNS query. The old entrypoint approach was fundamentally too late.
- **Sandbox-level env vars**: Setting `ANTHROPIC_BASE_URL` and `ANTHROPIC_API_KEY` in `buildSandboxEnv` instead of the Claude-specific wrapper means all tools (opencode, any future additions) inherit them automatically — correct architectural decision.
- **Non-fatal patch failure**: Warn-and-continue on ndots patch error is the right call; a failed sed shouldn't abort the runner.
- **`auth.py` cleanup**: Reading `os.environ.get("ANTHROPIC_API_KEY", "inference-routing")` instead of hardcoding is cleaner and correctly picks up whatever the control plane sets.

---

### Minor Issues

**Nit — sentinel value inconsistency** (`kube_reconciler.go`, `buildSandboxEnv`)

```go
env["ANTHROPIC_API_KEY"] = "unused-for-inference-routing"
```

This value travels all the way to `auth.py` and gets returned as the SDK's Bearer token value (the gateway proxy ignores it, so it's functionally fine). The old wrapper used `"gateway"`, the old runner used `"inference-routing"`, now it's `"unused-for-inference-routing"`. Pick one canonical sentinel and document it in a single place. Suggestion: keep `"inference-routing"` for consistency with existing logs/docs.

**Nit — no timeout on `execCtx`** (`kube_reconciler.go:475`)

```go
execCtx := context.Background()
```

`ExecSandbox` is synchronous. If the gRPC call hangs, this hangs indefinitely. A `context.WithTimeout(context.Background(), 10*time.Second)` would be a safety net for the sed command. Low risk in practice, but worth hardening.

---

### Checklist

| Check | Status |
|---|---|
| No `panic()` in production Go | ✅ |
| Errors wrapped with context | ✅ |
| No tokens in logs | ✅ |
| Structured logging with context fields | ✅ |
| SecurityContext unchanged | ✅ |
| CI passing | ✅ All gates green |

---

**Draft status noted** — test plan items are unchecked. Code itself is solid; happy to re-review once the e2e validation is done.

*— Amber*

PR #214, opened 2026-06-30
fix(runner): prevent infinite recursion in openshell claude wrapper and move inference env vars to sandbox level

bsquizz:
## Summary
- The openshell claude wrapper is installed as `/usr/local/bin/claude`, so when Claude Code spawns subagents that invoke `claude` by name, PATH resolves back to the wrapper causing infinite recursion
- Adds a file-based re-entry guard (`/tmp/.claude-wrapper-initialized`) that skips wrapper setup on subsequent invocations within the same sandbox instance — env vars can't be used because the sandbox supervisor spawns each command in a clean environment
- Moves `ANTHROPIC_BASE_URL` and `ANTHROPIC_API_KEY` from the wrapper into the reconciler's `buildSandboxEnv` so every tool in the sandbox gets them, not just Claude

## Test plan
- [ ] Deploy OpenShell runner and verify subagent spawning works without recursion
- [ ] Verify single-agent sessions still work correctly
- [ ] Verify inference routing works with env vars set at sandbox level

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 .PHONY: local-dev-token
 .PHONY: local-logs local-logs-api-server local-logs-ui local-logs-control-plane local-shell-api-server local-shell-ui
 .PHONY: local-test local-test-dev local-test-quick test-all local-troubleshoot local-port-forward local-stop-port-forward
-.PHONY: push-all registry-login setup-hooks remove-hooks lint check-minikube check-kind check-kubectl check-local-context dev-bootstrap kind-rebuild kind-reload-ambient-ui kind-reload-ambient-control-plane kind-reload-ambient-api-server kind-status kind-login kind-sso-toggle kind-setup-vertex
+.PHONY: push-all registry-login setup-hooks remove-hooks lint check-minikube check-kind check-kubectl check-local-context dev-bootstrap kind-rebuild kind-reload-ambient-ui kind-reload-ambient-control-plane kind-reload-ambient-api-server kind-reload-runner-openshell kind-status kind-login kind-sso-toggle kind-setup-vertex
 .PHONY: preflight-cluster preflight dev-env dev
 .PHONY: e2e-test e2e-setup e2e-clean deploy-langfuse-openshift
 .PHONY: unleash-port-forward unleash-status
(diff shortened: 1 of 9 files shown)


jsell-rh (approved):
## 🤖 Amber Analysis — PR #214

**Summary:** Tight, correct fix for a real recursion bug. The file-based guard is the right call given the sandbox supervisor's env-clearing behavior. Two minor notes before you mark this ready.

---

### Findings

**[Minor] PR description mentions `_CLAUDE_WRAPPER_ACTIVE` env var — code uses a file guard**

The bullet in the PR body says *"Adds a `_CLAUDE_WRAPPER_ACTIVE` re-entry guard"* but the implementation uses `GUARD="/tmp/.claude-wrapper-initialized"` — a file, not an env var. The code comment correctly explains the rationale (env not inherited between sandbox invocations). Just update the PR description to match the implementation before marking ready.

**[Minor] Guard file persists across top-level invocations within the pod lifetime**

Once created, `/tmp/.claude-wrapper-initialized` is never removed. In the current single-session sandbox model this is fine — one pod, one top-level `claude` invocation, then the pod terminates. But if the runner ever retries or re-invokes `claude` in the same pod (e.g., after a crash-restart of the wrapper), the second invocation would skip the `HOME=/sandbox` and `ANTHROPIC_BASE_URL` exports and exec directly to the real binary. Worth a comment noting the assumption that `/tmp` is ephemeral per pod and the guard is intentionally not cleaned up.

---

### Checklist

| Item | Result |
|------|--------|
| No `panic()` in production Go | N/A (bash script) |
| Guard mechanism correct for env-clearing sandbox | ✅ |
| No secrets / tokens introduced | ✅ |
| Comment explains non-obvious invariant | ✅ (file vs env var rationale) |
| Test plan reflects real deployment verification | ⚠️ Both items unchecked — expected for draft |

---

**Confidence:** 95% — Logic is correct for the single-session sandbox model. The minor items are clarity improvements, not blockers.

Approving. Ping me when you uncheck the test plan and mark it ready — I'll re-verify before merge.

— *Amber*

jsell-rh (approved):
## 🤖 Amber Re-Review — PR #214 (2 new commits since `f2ca98a`)

Both minor items from my previous review are resolved — the wrapper comment clearly explains the file-vs-env rationale, and the description now correctly describes the file guard. Re-approving with two new minor notes on the additions.

---

### New findings

**[Minor] TOML injection in `manifests.go` is fragile**

`ApplyConfigOverrides` searches for the line containing `app_armor_profile` and inserts `image_pull_policy` immediately after. If that string ever moves or is renamed in the upstream gateway TOML template, the injection silently does nothing — developers will see sandbox pull failures in kind without a clear error from the control plane. Since this only affects kind-local dev (not production), blast radius is low, but it would be worth adding a log line when the anchor string isn't found:

```go
// after the for loop, before the SetNestedMap
if !injected {
    r.logger.Warn().Msg("LOCAL_IMAGES: could not find app_armor_profile anchor in gateway.toml — image_pull_policy not injected")
}
```

**[Minor] `kind-rebuild` now excludes the OpenShell runner**

The refactored `kind-rebuild` calls `kind-reload-ambient-ui`, `kind-reload-ambient-control-plane`, and `kind-reload-ambient-api-server` — but not `kind-reload-runner-openshell`. The original used `build-all` which built everything. Developers doing a full rebuild to pick up runner changes will be silently missing that image. Either add `kind-reload-runner-openshell` to the chain, or update the `kind-rebuild` help text to note it's excluded.

---

### What's good in the new commits

- `rm -f /usr/local/bin/claude` before `COPY` is the correct fix for the symlink-overwrite problem — clean and targeted
- Wrapper comment block is now excellent: explains why env vars moved to sandbox level and why file guard vs env var
- `LOCAL_IMAGES` gating in `manifests.go` over a static configmap change is the right approach — production deployments won't accidentally get `IfNotPresent`
- `ANTHROPIC_API_KEY=unused-for-inference-routing` comment in `kube_reconciler.go` is clearer than the previous `"gateway"` sentinel

**Confidence:** 93% — The minor items are quality-of-life, not correctness issues. Approving.

— *Amber*

PR #217, opened 2026-07-01
fix(runner): fix openshell sandbox trust prompts and inference routing compat

bsquizz:
## Summary
- Move `HOME` and `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` exports above the guard check in the openshell claude wrapper so they apply on every invocation, not just the first
- Bootstrap `.claude.json` with per-project `hasTrustDialogAccepted` for `/sandbox` and `/sandbox/runner` to prevent trust dialog on manual `claude` runs
- Shorten dummy `ANTHROPIC_API_KEY` from `unused-for-inference-routing` to `notused` to avoid Claude Code custom API key rejection prompt
- Add `--bare` flag to the guard's exec path for consistency
- Update openshell-gateway skill: mark credential steps as REQUIRED, document that sandboxes must be created through the control plane (not `openshell sandbox create`)

## Test plan
- [ ] Deploy kind cluster with `OPENSHELL_USE_GATEWAY=true`
- [ ] Create credential and bind to tenant-a per skill Step 3
- [ ] Create session via `acpctl create session` and verify sandbox provisions with `providers:1`
- [ ] Connect to sandbox and run `claude` — should not prompt for trust or API key approval
- [ ] Verify `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1` is set in env on subsequent `claude` invocations

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -691,7 +691,7 @@ func (r *SimpleKubeReconciler) buildSandboxEnv(ctx context.Context, session type
 		// Set at sandbox level so every tool (claude, opencode, etc.) gets
 		// them — not just processes launched through a specific wrapper.
 		env["ANTHROPIC_BASE_URL"] = "https://inference.local"
-		env["ANTHROPIC_API_KEY"] = "unused-for-inference-routing"
+		env["ANTHROPIC_API_KEY"] = "notused"
 	} else if r.cfg.VertexEnabled {
 		env["USE_VERTEX"] = "1"
 		env["CLAUDE_CODE_USE_VERTEX"] = "1"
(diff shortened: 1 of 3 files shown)


jsell-rh (approved):
🤖 **Amber Review — Approved**

All three changes are correct and well-scoped. The root cause analysis in the PR summary is accurate — the `HOME`/`CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS` ordering bug was a real issue on subsequent invocations, and the `notused` sentinel sidesteps the Claude Code key-length rejection cleanly. Skill doc improvements are high value.

Two minor observations (neither blocks merge):

**1. Misleading comment in the wrapper script**

The new comment above the `.claude.json` bootstrap block reads:

> `# The customApiKeyResponses pre-approves the dummy key suffix used for inference routing...`

But `customApiKeyResponses` is **not present** in the JSON being written. The actual fix for the API key prompt is the `"notused"` change in `kube_reconciler.go`. This comment will confuse future readers who search `.claude.json` for that field and find nothing. Suggest updating it to something like:

```bash
# Without this, Claude Code prompts for trust on the first run in each project directory.
# The API key prompt is suppressed separately by using the short sentinel "notused" in
# kube_reconciler.go — Claude Code only warns on keys it recognizes as custom-format.
```

**2. `--bare` only on the guard's early-return path**

`--bare` is added to the early-return `exec` (subsequent invocations) but not to the first-run `exec` at the bottom of the script. If `--bare` is desired consistently, the first-run path should match. If the first run intentionally needs non-bare output (e.g. for setup feedback), that's fine — worth a comment explaining the asymmetry.

Both are Minor per ACP severity classification. Approving.

— *Amber*

jsell-rh:

PR #219, opened 2026-07-01
[PLATENG-126] feat(gateway): wire provider declarations into sandbox provisioning

squizzi:
## Summary

- Replaces credential-based provider resolution with declaration-based flow in gateway mode. The CP reads provider declarations (synced from ConfigMaps), fetches credentials from referenced K8s Secrets, and creates/updates OpenShell providers on the gateway.
- Wires agent sandbox policy into `CreateSandboxRequest` — policy resolution failure is now a hard error.
- Adds `setup-vertex-provider.sh` (Secret + example declarations) and `setup-gateway-cli.sh` (auto-discovers gateways, extracts mTLS certs, port-forwards on random ports).
- Adds `make kind-setup-openshell-cli` target with auto-discovery and updates `make kind-setup-vertex` to use provider declarations when gateway pods are detected.

## Test plan

- [x] Run `make kind-setup-vertex` on a gateway-enabled cluster — verify it uses `setup-vertex-provider.sh` for each tenant namespace
- [x] Run `make kind-setup-openshell-cli` — verify it discovers gateway pods and registers CLI access
- [x] Create a session with `--agent <name>` — verify CP logs show "resolved providers from agent declarations"
- [x] Verify sandbox has providers and inference routing configured via `openshell sandbox list --gateway <namespace>`
- [ ] Test without gateway (OPENSHELL_USE_GATEWAY=false) — verify `kind-setup-vertex` falls back to `setup-vertex-kind.sh`


jsell-rh (changes requested):

**Note:** This is a draft PR — flagging issues early so they can be addressed before it's marked ready.

The declaration-based provider flow is the right architectural direction, and the implementation is solid overall. Two blockers to fix before merge.

### Blockers

**1. Raw unstructured map access in `readProviderSecretToken`** (`kube_reconciler.go`)

CLAUDE.md and the ACP Constitution require type-safe unstructured helpers. Direct type assertions on `secret.Object` violate this:

```go
// ❌ Current — direct assertions, verbose, brittle
data, ok := secret.Object["data"].(map[string]interface{})
encoded, ok := data["token"]
encodedStr, ok := encoded.(string)

// ✅ Fix — use the helper, handles nested access safely
encodedStr, found, err := unstructured.NestedString(secret.Object, "data", "token")
if err != nil {
    return "", fmt.Errorf("reading token from %s/%s: %w", namespace, secretName, err)
}
if !found || encodedStr == "" {
    return "", fmt.Errorf("secret %s/%s has no 'token' key", namespace, secretName)
}
```

**2. Dead code: `configureInference` is unreachable on this branch**

The original `configureInference` function (which takes `credentialIDs map[string]string`) is still present at the bottom of `kube_reconciler.go`, but its call site was replaced by `configureInferenceFromProviders`. It has no callers and should be removed. Dead production code causes confusion and makes future readers wonder if it's intentionally preserved for something.

### Major

**3. No unit tests for new reconciler functions**

`resolveAgentProviders`, `resolveAgentSandboxPolicy`, `readProviderSecretToken`, and `configureInferenceFromProviders` are non-trivial functions with error branches and validation logic, but the PR adds no tests for them. Constitution Principle IV (TDD) requires test coverage. At minimum: tests for `readProviderSecretToken` (good secret, missing key, bad base64) and `resolveAgentSandboxPolicy` (empty policy, not-found policy, invalid JSON spec).

### Minor

**4. Function name `resolveAgentProviders` is misleading**

The function does more than resolve — it creates or updates providers on the gateway via gRPC. A name like `ensureAgentProviders` or `provisionAgentProviders` would more accurately convey the side effects. Not blocking, but worth a follow-up.

**5. Makefile: unquoted `$$NAMESPACES` passed to `setup-gateway-cli.sh`**

```makefile
./scripts/setup-gateway-cli.sh $$NAMESPACES
```

If a namespace ever contains spaces (unlikely in K8s but defensive), this word-splits. Quoting: `"$$NAMESPACES"` or using the variable differently. Low risk but easy to fix.

---

The `validateTSLValue` input sanitization before TSL interpolation is correct. The `setup-gateway-cli.sh` trap semantics (`trap - EXIT` before `wait`) are intentional and correct — port-forwards are designed to survive the script. The platform-info endpoint design (process-lifetime constant, `Cache-Control: public`) is sound.

— *Amber*

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -622,6 +619,200 @@ func (r *SimpleKubeReconciler) ensureVertexCredentialRefresh(ctx context.Context
...
+		if provType == "" {
+			provType = declName
+		}
+
+		token, readErr := r.readProviderSecretToken(ctx, namespace, provDecl.Secret)
+		if readErr != nil {
+			return nil, nil, fmt.Errorf("reading secret %s for provider %s: %w", provDecl.Secret, declName, readErr)
+		}
+
+		osType := openshell.OpenShellProviderType(provType)
```
jsell-rh, line 674: **Blocker (carryover) — raw type assertions still present**

This is the same blocker from my prior review, unchanged across all 4 new commits. CLAUDE.md requires type-safe unstructured helpers:

```go
// ❌ Current — raw type assertions
data, ok := secret.Object["data"].(map[string]interface{})
if !ok {
    return "", fmt.Errorf("secret %s/%s has no data field", namespace, secretName)
}
encoded, ok := data["token"]
if !ok { ... }
encodedStr, ok := encoded.(string)
if !ok { ... }
```

```go
// ✅ Fix — use unstructured.NestedString
encodedStr, found, err := unstructured.NestedString(secret.Object, "data", "token")
if err != nil {
    return "", fmt.Errorf("reading token from %s/%s: %w", namespace, secretName, err)
}
if !found || encodedStr == "" {
    return "", fmt.Errorf("secret %s/%s has no 'token' key", namespace, secretName)
}
```

`unstructured.NestedString` handles the nested map traversal and type coercion safely. Import: `k8s.io/apimachinery/pkg/apis/meta/v1/unstructured`."

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -622,6 +619,200 @@ func (r *SimpleKubeReconciler) ensureVertexCredentialRefresh(ctx context.Context
 	return nil
 }
 
+func (r *SimpleKubeReconciler) resolveAgentProviders(
+	ctx context.Context,
+	sdk *sdkclient.Client,
+	namespace, projectName string,
```
jsell-rh, line 626: **Blocker (carryover) — dead function still present**

`configureInference` (the credential-ID-based variant) has no callers on this branch — its call site was replaced by `configureInferenceFromProviders`. It should be deleted. Dead production code causes maintenance confusion and signals incomplete cleanup.

Search for callers: `grep -n 'r\.configureInference\b' kube_reconciler.go` — you'll find zero. Remove the function."

jsell-rh (changes requested):
## 🤖 Amber Re-Review — PR #219 (2026-07-01)

Good iteration on this PR. The new commits resolve one of my prior blockers and add real value, but the two structural blockers from my initial review are still present unchanged.

**What's improved ✅**
- Inference configuration failure is now a hard error (`fix(gateway): make inference configuration failure a hard error`) — correct. A sandbox missing inference routing is unusable; warn-and-continue was wrong.
- Conflict resolution with upstream's 3-arg `ProviderConfig` signature is clean.
- Script updates (`setup-gateway-cli.sh`, `setup-vertex-provider.sh`) look correct; the `trap - EXIT` before `wait` in `setup-gateway-cli.sh` is intentional — port-forwards are designed to outlive the script.
- `examples/agent-sandbox-config.yaml`: removing hardcoded `namespace: tenant-a` from ConfigMap metadata is the right call — namespace is supplied at `kubectl apply -n` time.

**Still blocking ❌ (both carryover, unchanged across all 4 new commits)**

1. **`readProviderSecretToken` uses raw type assertions** — see inline comment. `unstructured.NestedString(secret.Object, "data", "token")` is the correct helper per CLAUDE.md and ACP Constitution.

2. **`configureInference` (dead function) still present** — no callers on this branch; the call site was replaced by `configureInferenceFromProviders`. Delete it.

**Major (carryover) ⚠️**

3. **No unit tests for new reconciler functions** — `resolveAgentProviders`, `readProviderSecretToken`, `resolveAgentSandboxPolicy`, and `configureInferenceFromProviders` have no test coverage. Constitution Principle IV requires tests. Minimum: `readProviderSecretToken` (good secret, missing key, bad base64) and `resolveAgentSandboxPolicy` (empty policy, not-found, invalid JSON spec).

**Minor**

- `Makefile` `kind-setup-openshell-cli`: `./scripts/setup-gateway-cli.sh $$NAMESPACES` — unquoted. If a namespace name ever contains spaces this word-splits. Prefer `"$$NAMESPACES"`. Low risk.

---

Two blockers, two commits away from merge. The overall architecture is sound.

— *Amber*

jsell-rh (commented):
## 🤖 Amber Re-Review — PR #219 (new commits since last review)

**Prior blockers: ✅ Both resolved**

- `unstructured.NestedString` replaces raw type assertions in `readProviderSecretToken` — correct fix
- Dead `configureInference` function deleted — clean

Thank you for the quick turnaround on those. A few new observations from the additional commits:

---

### New Issues

**Major — `failSession` uses an already-potentially-cancelled context**

In `execAfterReady`, `failSession` closes over `pollCtx` (120s timeout). When the timeout fires, `pollCtx` is cancelled — then `failSession("sandbox did not become ready within 120s")` is called, but `sdk.Sessions().UpdateStatus(pollCtx, ...)` immediately fails because the context is already done. The session never gets marked Failed from the timeout path.

Fix: Use a fresh short-lived context for the failure write:
```go
failSession := func(reason string) {
    writeCtx, writeCancel := context.WithTimeout(context.Background(), 10*time.Second)
    defer writeCancel()
    now := time.Now()
    condJSON, _ := json.Marshal([]map[string]string{{"type": "SandboxFailure", "reason": reason}})
    if _, err := sdk.Sessions().UpdateStatus(writeCtx, sessionID, ...); err != nil {
        r.logger.Warn()...
    }
}
```

**Minor — Self-cancelling commits make history confusing**

Commit `3d8e18f` adds session-completed logic in `execAfterReady` (marks `PhaseCompleted`, sets `completion_time`). Commit `db54443` removes that same logic and replaces it with a FIXME. Both commits are in this PR. Before merging, squash these so the net state is clear — a reviewer reading the history would expect `3d8e18f` to be live, but it's not.

**Minor — `resolveCredential` uses unvalidated string interpolation in search**

```go
opts.Search = fmt.Sprintf("name = '%s'", nameOrID)
```

If the API server passes this search string directly to the database, a name like `x' OR '1'='1` could cause unexpected results. This is CLI-side so the blast radius is limited to the authenticated user's own resources, but it's worth using the same TSL validation (`validateTSLValue`) that `resolveAgentProviders` uses on provider names.

**Minor — FIXMEs need tracking**

Two `// FIXME` comments gate critical session lifecycle behavior (marking Completed, deprovisioning sandbox on completion). These should have a GitHub issue reference so they don't get forgotten. Suggest opening an issue before merging and adding `// FIXME(#NNN):` references.

---

### What's Good (new since last review)
- `mapSandboxPhaseToSessionPhase` now maps PROVISIONING→Creating and READY→Running ✅
- `execAfterReady` now marks session Failed on timeout, error, and exec failure ✅
- `resolveAgentSandboxPolicy` wires sandbox policy into `CreateSandboxRequest` as a hard error ✅
- `resolveCredential` helper DRYs up credential resolution across all CLI commands ✅
- `credentials.Delete` RowsAffected check properly returns `ErrRecordNotFound` ✅

---

The `failSession` context issue is the only blocker I'd flag — everything else is squash/cleanup. Once that's fixed this is close to merge-ready.

*— Amber*

jsell-rh (commented):
## 🤖 Amber Re-Review — PR #219 (2026-07-01, new commit `0f07bee`)

**New commit assessment: ✅ Correct fix**

The reorder moves `enableProvidersV2` + `resolveAgentProviders` + `configureInferenceFromProviders` before the sandbox-exists early-return. This is the right call — without it, restarting a session with an existing sandbox silently skipped all provider reconciliation. Since all three calls are idempotent, this is safe.

One trade-off to be aware of: provider resolution failures now also block restarts of already-running sandboxes (not just initial creation). That's consistent with the "hard error" design from `e7ce0f8` and is intentional — providers are the source of truth for inference routing, so an unresolvable provider is a real failure regardless of sandbox state.

---

**Outstanding blockers — unchanged from prior reviews:**

1. **`failSession` context issue** (Major, carryover from `3FuruG2jDDeESxxXqiT6Kuv4jul` review): `failSession` closes over `pollCtx` (120s timeout context). When the timeout fires, `pollCtx` is cancelled before `failSession` is called — `UpdateStatus(pollCtx, ...)` immediately fails and the session is never marked Failed. Fix: use `context.Background()` with a short deadline in `failSession`.

2. **No unit tests** (Major, carryover): `resolveAgentProviders`, `readProviderSecretToken`, `resolveAgentSandboxPolicy`, `configureInferenceFromProviders` — still zero test coverage. Constitution Principle IV.

3. **FIXMEs need issue references** (Minor): Two `// FIXME` comments gate session completion and sandbox deprovision. Please open a tracking issue and add `// FIXME(#NNN):` before merge.

4. **`resolveCredential` string interpolation** (Minor): `opts.Search = fmt.Sprintf("name = '%s'", nameOrID)` — same `validateTSLValue` guard used in `resolveAgentProviders` should apply here.

---

Good progress. Two commits away from resolving the blockers.

*— Amber*

jsell-rh (commented):
## 🤖 Amber Re-Review — PR #219 (2026-07-01, commits `5f7a60c` + `5d54bd4`)

**All prior blockers and minors resolved. ✅**

### What was fixed

| Issue | Status |
|-------|--------|
| `failSession` uses cancelled `pollCtx` on timeout | ✅ Fixed — independent `context.Background()` with 10s deadline |
| FIXMEs missing issue references | ✅ Fixed — both reference `#223` |
| `resolveCredential` unvalidated string interpolation | ✅ Fixed — `safeTSLPattern` guard skips name-search for unsafe input, falls through to ID |
| `resolveCredential` hard-errors on List failure | ✅ Fixed — falls through to Get-by-ID when List fails |

The `failSession` fix is correct and the approach is sound — `pollCtx` is already cancelled when the timeout branch fires, so `context.Background()` with a short deadline is exactly right.

---

### Still outstanding (carryover)

**Major — No unit tests for new reconciler functions (Constitution Principle IV)**

`resolveAgentProviders`, `readProviderSecretCredentials`, `resolveAgentSandboxPolicy`, and `configureInferenceFromProviders` still have no test coverage. This is the last item blocking merge. Minimum coverage needed:

- `readProviderSecretCredentials`: good secret with multiple keys, missing key, bad base64
- `resolveAgentSandboxPolicy`: nil agent, empty spec, not-found policy, invalid JSON spec
- `resolveAgentProviders`: nil agent, provider not found in API, secret read failure

This is a draft PR so I understand tests may come in a follow-up commit — just flagging so it doesn't get forgotten before marking ready.

---

Everything else is merge-quality. One commit of tests and this is ready.

*— Amber*

jsell-rh (commented):
## 🤖 Amber Re-Review — PR #219 (2026-07-01, commits `f0e9a25` + `3496c6c`)

**Both new commits look good. One carryover major outstanding.**

### What's new

**`f0e9a25` — Delete dead `ensureGatewayProviders`** ✅

Clean pure-deletion. The function was fully superseded by `resolveAgentProviders`; the commit message correctly notes that ADC/authorized_user support flows through `ensureVertexCredentialRefresh` which `resolveAgentProviders` calls. No concerns.

**`3496c6c` — Align provider mappings with OpenShell v2 profiles** ✅

Good cleanup. Specific observations:

- `anthropic/claude → claude-code` fix is correct — `claude-code` is the actual OpenShell profile name ✅
- New profiles (codex, copilot, cursor, deepinfra, nvidia, pypi) with their credential key + env var mappings look correct ✅
- Removing the duplicated local `providerTypeMapping()` from `kube_reconciler.go` in favor of `openshell.KnownAmbientProviderTypes()` is the right DRY move ✅
- Removing `openai` and `aws-bedrock` from `inferenceCapableTypes` is correct — neither has an OpenShell profile ✅
- `providerInjectedEnvVars` keys are OpenShell type names (post-mapping) and `ProviderInjectedEnvVars` takes `openshellType string` — consistent ✅

**One thing worth verifying before merge:** `IsInferenceCapable` parameter is named `ambientProvider` but the map now only contains OpenShell type names (`claude-code`, `google-vertex-ai`, etc.). If any call site passes an ACP ambient name (e.g., `"anthropic"`) directly to `IsInferenceCapable` without mapping first, inference detection for that provider is silently broken. If all call sites already map through `OpenShellProviderType()` first, the removal of `anthropic`/`claude` from `inferenceCapableTypes` is correct. Just confirm this in the call sites.

---

### Still outstanding (carryover)

**Major — No unit tests for new reconciler functions (Constitution Principle IV)**

`resolveAgentProviders`, `readProviderSecretCredentials`, `resolveAgentSandboxPolicy`, and `configureInferenceFromProviders` still have zero test coverage. This is the last item between this PR and merge-ready. Minimum:

- `readProviderSecretCredentials`: good secret with multiple keys, missing key, bad base64
- `resolveAgentSandboxPolicy`: nil agent, empty spec, not-found policy, invalid JSON spec
- `resolveAgentProviders`: nil agent, provider not found in API, secret read failure

PR is still draft which is the right state. Everything else is in good shape — this is one commit of tests away.

*— Amber*

jsell-rh (commented):
## 🤖 Amber Re-Review — PR #219 (2026-07-01, commits `2e7ff9a` + `6d68ecf`)

Both new commits look correct. One carryover major still outstanding.

### What's new

**`2e7ff9a` — ADC credentials for vertex provider** ✅

The distinction between SA key and ADC credential types is correct and necessary. Specific observations:

- `ProviderCredentialsFromSecret`: ADC path returns `map[string]string{}` (empty initial credentials) — correct, because ADC uses OAuth2 refresh tokens and the gateway mints the first access token via the refresh flow. SA path returns `GOOGLE_SERVICE_ACCOUNT_KEY` — correct.
- `VertexRefreshCredentialKey(credType)` replacing the constant `VertexAIRefreshCredentialKey` is the right abstraction — the two credential types use different storage keys (`GOOGLE_VERTEX_AI_TOKEN` for ADC vs `GOOGLE_VERTEX_AI_SERVICE_ACCOUNT_TOKEN` for SA). Function signature is clean.
- `ensureVertexCredentialRefresh` updated to use `VertexRefreshCredentialKey(credType)` — consistent ✅
- `resolveAgentProviders` error message updated from "SA JSON" to "credential JSON" — accurate for both types ✅

**`6d68ecf` — VERTEX_KEY_FILE → VERTEX_CRED rename + kind-setup-vertex wiring** ✅

Pure rename/cleanup. `$(MAKE) --no-print-directory kind-setup-vertex` in `kind-up` is the right pattern — it picks up gateway detection automatically instead of calling `setup-vertex-kind.sh` directly. No concerns.

---

### Still outstanding (carryover from prior reviews)

**Major — No unit tests for new reconciler functions (Constitution Principle IV)**

`resolveAgentProviders`, `readProviderSecretCredentials`, `resolveAgentSandboxPolicy`, and `configureInferenceFromProviders` still have zero test coverage. Now that the ADC credential type detection is also in the mix, `ProviderCredentialsFromSecret` (vertex branch) also needs coverage:

- `ProviderCredentialsFromSecret` with vertex + ADC token → empty map
- `ProviderCredentialsFromSecret` with vertex + SA token → `GOOGLE_SERVICE_ACCOUNT_KEY` set
- `VertexRefreshCredentialKey` for both `GoogleCredentialAuthorizedUser` and SA types
- `readProviderSecretCredentials`: good secret with multiple keys, missing key, bad base64
- `resolveAgentSandboxPolicy`: nil agent, empty spec, not-found policy, invalid JSON spec

PR is still draft — appropriate state. Everything else is in good shape. One commit of tests and this is ready.

— *Amber*

jsell-rh (approved):
## 🤖 Amber Re-Review — PR #219 (commits `6adce48` + `6a942eb`)

**All prior blockers and majors resolved. ✅**

### What was fixed

| Issue | Status |
|-------|--------|
| No unit tests for new reconciler functions (Constitution Principle IV) | ✅ Fixed — 31 table-driven cases added |
| All prior blockers (unstructured helpers, dead function, failSession context, FIXME refs, TSL guard) | ✅ Already resolved in prior commits |

### Test quality assessment

**`provider_reconciler_test.go`** — 31 cases across 4 functions:

- `readProviderSecretCredentials` — good multi-key, single key, missing secret, no data field, empty data, bad base64, mixed non-string (skipped), all non-string (error). Covers all the cases I requested. ✅
- `resolveAgentSandboxPolicy` — nil agent, empty policy name, invalid name (TSL injection), not found, valid, empty spec returns nil, invalid JSON spec. The TSL injection case is particularly good. ✅
- `resolveAgentProviders` — 12 cases: nil agent, empty list, not-found, invalid name, no secret reference, secret read failure, anthropic success, github (non-inference), update→create fallback, update internal error, create failure after NotFound, multi-provider mixed inference. This matches exactly what I asked for. ✅
- `configureInferenceFromProviders` — empty, single provider success, default model, custom model, SetClusterInference error, non-inference-capable skipped. The call count verification is a nice addition. ✅

**`provider_mapping_test.go`** — vertex ADC/SA branches, known types without token key, unknown passthrough, `VertexRefreshCredentialKey` for both cred types, `DetectGoogleCredentialType` for all cases including the error paths. Covers the ADC/SA distinction I flagged in the prior review. ✅

**Interface extraction** (`gateway_iface.go`) — clean. The `gatewayClient` interface is minimal (only what the reconciler needs), and the `mockGateway` stub correctly implements all methods with optional function overrides only where tests need control. Changing `gateway *openshell.GatewayClient` → `gateway gatewayClient` in `SimpleKubeReconciler` is the right move. ✅

### Still outstanding (minor, carryover — does not block merge)

**Makefile `kind-setup-openshell-cli`: unquoted `$$NAMESPACES`**

```makefile
./scripts/setup-gateway-cli.sh $$NAMESPACES
```

K8s namespace names can't contain spaces so the risk is essentially zero in practice, but `"$$NAMESPACES"` is the defensive habit. Suggest as follow-up in a separate cleanup commit.

---

Everything is in good shape. The test suite now covers all the functions I flagged, with error paths and boundary conditions. Approving.

*— Amber*

PR #220, opened 2026-07-01
fix(control-plane): skip no-op patches to preserve updated_at timestamps

squizzi:
## Summary

- The ConfigMap syncer unconditionally PATCHed every agent, provider, and policy on each control plane restart, causing GORM to auto-update `updated_at` even when nothing changed
- Adds a content hash (`ambient.ai/content-hash`) to resource annotations that tracks the patch payload; when the hash matches on subsequent syncs, the PATCH is skipped entirely
- Preserves accurate "Last Updated" timestamps in the UI across control plane restarts

## Test plan

- [ ] `go build ./...` and `go vet ./...` pass
- [ ] Restart the control plane pod and verify agents/policies/providers retain their original "Last Updated" timestamps
- [ ] Edit a ConfigMap declaration and verify the timestamp updates correctly
- [ ] Verify newly created resources get the content hash annotation on first sync

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/configmap_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/configmap_reconciler.go
@@ -364,9 +364,10 @@ func (s *ConfigMapSyncer) buildOriginAnnotations(namespace string, userAnnotatio
 	return string(raw), nil
 }
 
-func computePatchHash(patch map[string]interface{}) string {
+func (s *ConfigMapSyncer) computePatchHash(patch map[string]interface{}) string {
 	data, err := json.Marshal(patch)
 	if err != nil {
+		s.logger.Warn().Err(err).Msg("failed to marshal patch for content hash, will force update")
 		return ""
 	}
 	h := sha256.Sum256(data)
@@ -463,7 +464,7 @@ func (s *ConfigMapSyncer) upsertAgent(ctx context.Context, sdk *sdkclient.Client
 		patch["labels"] = string(labelsJSON)
 	}
 
-	hash := computePatchHash(patch)
+	hash := s.computePatchHash(patch)
 
 	if existing != nil {
 		if extractContentHash(existing.Annotations) == hash {
@@ -657,7 +658,7 @@ func (s *ConfigMapSyncer) upsertProvider(ctx context.Context, sdk *sdkclient.Cli
 		patch["secret"] = decl.Secret
 	}
 
-	hash := computePatchHash(patch)
+	hash := s.computePatchHash(patch)
 
 	if existing != nil {
 		if extractContentHash(existing.Annotations) == hash {
@@ -852,7 +853,7 @@ func (s *ConfigMapSyncer) upsertPolicy(ctx context.Context, sdk *sdkclient.Clien
 		"spec":        spec,
 	}
 
-	hash := computePatchHash(patch)
+	hash := s.computePatchHash(patch)
 
 	if existing != nil {
 		if extractContentHash(existing.Annotations) == hash {
--- a/components/ambient-control-plane/internal/reconciler/configmap_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/configmap_reconciler.go
@@ -2,6 +2,8 @@ package reconciler
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


jsell-rh (changes requested):
## 🤖 Amber Review — PR #220

**Summary:** The content-hash idempotency fix is the right pattern — skipping no-op patches to preserve `updated_at` is exactly what's needed. The SHA-256 approach is clean and the hash-over-payload-without-annotation design is stable. A few issues to address before marking ready for review.

**Confidence:** 85% — logic is sound, but there are correctness gaps in the create path and a silent-failure in hash computation.

---

### Issues

**Major — Create path does a second PATCH that it cannot roll back on failure**

After `createAPI`, the code calls `setPatchContentHash(patch, hash)` then `patchAPI(...)` to embed the hash. If `patchAPI` fails here, the resource exists in the API but has no content-hash annotation. On the next sync, `extractContentHash` returns `""`, the hash won't match, so it correctly re-patches — but if the second patch keeps failing the resource will be re-patched on every control plane restart indefinitely (never settling). Consider: set the content-hash annotation in the initial create payload, or accept this as a best-effort annotation (low consequence in practice, but worth documenting explicitly).

**Minor — `computePatchHash` silently returns `""` on marshal error**

```go
func computePatchHash(patch map[string]interface{}) string {
    data, err := json.Marshal(patch)
    if err != nil {
        return ""  // ← silent empty string
    }
    ...
}
```

An empty hash will never match a stored hash, so the skip-path is effectively disabled and every sync generates a PATCH. Since `json.Marshal` on `map[string]interface{}` shouldn't fail in practice, this is low-risk — but the silent fallback makes it hard to diagnose if something unexpected goes wrong. Consider logging a warning here.

**Minor — Test plan items are unchecked**

The PR is marked draft, which is fine, but the test plan has no checked items. Before marking ready-for-review, please verify and check the test plan steps (particularly the restart + timestamp retention test).

---

### What's Good
- Hash is computed over the pre-annotation patch, so it's stable across syncs ✅
- Pattern is consistently applied to agents, providers, and policies ✅
- Error wrapping follows `fmt.Errorf("context: %w", err)` convention ✅
- Debug logging on skip is the right signal level ✅

---

*— Amber*

jsell-rh (commented):
## 🤖 Amber Re-Review — PR #220 (2026-07-01, commit `f4893eb`)

**Minor fix landed. Major create-path issue still outstanding.**

### What was fixed

**`computePatchHash` silent failure** ✅ — now logs `Warn` with error before returning `""`. This makes it diagnosable if the skip-path is unexpectedly disabled. Good fix.

---

### Still outstanding

**Major — Create path double-PATCH leaves resource without hash on second-PATCH failure**

The create path still does:
1. `createAPI(...)` — resource created, no hash annotation yet
2. `setPatchContentHash(patch, hash)` — modifies patch
3. `patchAPI(...)` — second PATCH to embed hash

If step 3 fails, the resource exists with no `ambient.ai/content-hash`. On every subsequent sync: `extractContentHash` returns `""`, hash doesn't match, PATCH is attempted — indefinitely. This won't settle until the PATCH succeeds or the resource is deleted and recreated.

The simplest fix: include the content hash in the initial create payload. Since `buildOriginAnnotations` already produces the annotations string and `setPatchContentHash` just adds one more key to it, you can compute the hash before creating and include it in the create body directly — eliminating the second PATCH entirely for the create path.

If that's architecturally difficult (create and patch payloads differ), documenting this as a known best-effort behavior in a code comment is an acceptable alternative — but the indefinite re-patch consequence should be explicit.

---

This is still draft which is appropriate. The content-hash skip logic and warn-on-failure are both correct. One more fix and this is ready.

*— Amber*

jsell-rh (approved):
## 🤖 Amber Re-Review — PR #220 (commit `f4893eb`): Approved ✅

Both issues from my prior reviews are now resolved.

### What was fixed

| Issue | Status |
|-------|--------|
| Create path double-PATCH (resource created without hash, then hash-PATCH could fail indefinitely) | ✅ Fixed — `createAPI` sends a single atomic POST with all fields including the content hash. No second PATCH needed. |
| `computePatchHash` silent empty-string on marshal failure | ✅ Fixed — now logs `Warn` with error, making it diagnosable when the skip-path is unexpectedly disabled. |

The new `createAPI` helper is the right call — a single POST with the full patch payload (including the hash annotation) eliminates the failure window entirely and is simpler than the prior two-step pattern. The implementation is straightforward: marshal body, POST to `/api/ambient/v1/projects/:id/:resource`, parse ID from response. No concerns.

The hash-over-pre-annotation-patch design is stable and correct — the hash won't change between syncs as long as the declaration hasn't changed, regardless of what other annotations get added externally.

**Checklist:**
- [x] Error wrapping follows `fmt.Errorf("context: %w", err)` convention
- [x] Debug logging on skip is the right signal level
- [x] Hash applied consistently to agents, providers, and policies
- [x] Create path is now atomic (single POST)
- [x] Warn on marshal failure

This is ready to mark as ready-for-review when you're happy with local testing of the restart + timestamp-retention scenario.

— *Amber*

PR #221, opened 2026-07-01
fix(gateway): ADC credential support, vertex provisioning, and runner entrypoint

markturansky:
## Summary

- **ADC (Application Default Credentials) support**: Detects Google credential type (`authorized_user` vs `service_account`) and uses the appropriate gateway refresh strategy (OAuth2 refresh token vs SA JWT). Adds `DetectGoogleCredentialType`, `ExtractOAuth2RefreshMaterial`, and `ExtractServiceAccountJWTMaterial` functions in `provider_mapping.go`.
- **Runner entrypoint fix**: Changed `resolveEntrypoint()` default from `["claude"]` to `["/sandbox/runner/entrypoint.sh"]` — bare `claude` exits immediately with no TTY/prompt; the runner server (uvicorn) must start first to manage the Claude CLI lifecycle.
- **Vertex credential provisioning in Kind**: `setup-kind-openshell.sh` now creates `vertex-<tenant>` credentials via the API and binds them to projects with `credential:viewer` role when `GOOGLE_APPLICATION_CREDENTIALS` is set.
- **Makefile vertex env vars**: Passes `GOOGLE_APPLICATION_CREDENTIALS`, `ANTHROPIC_VERTEX_PROJECT_ID`, and `CLOUD_ML_REGION` to `setup-kind-openshell.sh`.
- **Idempotent kind-up**: Background port-forwards and acpctl login during `kind-up`.
- **mTLS for gateway connections in Kind**: TLS configuration for gateway gRPC connections.

## Known Issues (WIP)

- **502 Bad Gateway on `acpctl session send -f`**: `provisionSessionSandbox` does not call `ensureService()`, so no K8s Service is created for sandbox sessions. The API server SSE proxy cannot reach the runner.
- **400 context_management from Vertex AI**: Claude Code CLI sends request parameters that Vertex doesn't support (`context_management: Extra inputs are not permitted`).

## Test plan

- [x] `make kind-up OPENSHELL_USE_GATEWAY=true` with `GOOGLE_APPLICATION_CREDENTIALS` set to ADC file
- [x] Verify sandbox pods start with correct entrypoint (`/sandbox/runner/entrypoint.sh`)
- [x] Verify uvicorn + claude CLI processes running in sandbox
- [x] Verify inference routing (`ALLOWED inference.local:443`)
- [ ] End-to-end session message streaming (blocked by 502/Service issue)

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -391,7 +391,7 @@ func (r *SimpleKubeReconciler) resolveEntrypoint(agent *types.Agent) []string {
 	if agent != nil && agent.Entrypoint != "" {
 		return []string{agent.Entrypoint}
 	}
-	return []string{"claude"}
+	return []string{"/sandbox/runner/entrypoint.sh"}
 }
 
 func (r *SimpleKubeReconciler) resolveSandboxImage(agent *types.Agent) string {
@@ -584,24 +584,63 @@ func (r *SimpleKubeReconciler) ensureGatewayProviders(ctx context.Context, names
 	return providerNames, nil
 }
 
-func (r *SimpleKubeReconciler) ensureVertexCredentialRefresh(ctx context.Context, namespace, provName, saKeyJSON string) error {
-	material, err := openshell.ExtractServiceAccountJWTMaterial(saKeyJSON)
+func (r *SimpleKubeReconciler) ensureVertexCredentialRefresh(ctx context.Context, namespace, provName, credJSON string) error {
+	credType, err := openshell.DetectGoogleCredentialType(credJSON)
 	if err != nil {
-		return fmt.Errorf("parsing service account key: %w", err)
+		return fmt.Errorf("detecting credential type: %w", err)
 	}
 
 	credKey := openshell.VertexAIRefreshCredentialKey
+	var refreshReq *openshellpb.ConfigureProviderRefreshRequest
 
-	if _, err := r.gateway.ConfigureProviderRefresh(ctx, namespace, &openshellpb.ConfigureProviderRefreshRequest{
-		Provider:      provName,
-		CredentialKey: credKey,
-		Strategy:      openshellpb.ProviderCredentialRefreshStrategy_PROVIDER_CREDENTIAL_REFRESH_STRATEGY_GOOGLE_SERVICE_ACCOUNT_JWT,
-		Material: map[string]string{
-			"client_email": material.ClientEmail,
-			"private_key":  material.PrivateKey,
-		},
-		SecretMaterialKeys: []string{"private_key"},
-	}); err != nil {
+	switch credType {
+	case openshell.GoogleCredentialServiceAccount:
+		material, err := openshell.ExtractServiceAccountJWTMaterial(credJSON)
+		if err != nil {
+			return fmt.Errorf("parsing service account key: %w", err)
+		}
+		refreshReq = &openshellpb.ConfigureProviderRefreshRequest{
+			Provider:      provName,
+			CredentialKey: credKey,
+			Strategy:      openshellpb.ProviderCredentialRefreshStrategy_PROVIDER_CREDENTIAL_REFRESH_STRATEGY_GOOGLE_SERVICE_ACCOUNT_JWT,
+			Material: map[string]string{
+				"client_email": material.ClientEmail,
+				"private_key":  material.PrivateKey,
+			},
+			SecretMaterialKeys: []string{"private_key"},
+		}
+		r.logger.Info().
+			Str("provider", provName).
+			Msg("using service account JWT strategy for vertex credential refresh")
+
+	case openshell.GoogleCredentialAuthorizedUser:
+		material, err := openshell.ExtractOAuth2RefreshMaterial(credJSON)
+		if err != nil {
+			return fmt.Errorf("parsing authorized_user credential: %w", err)
+		}
+		clientEmail := material.Account
+		if clientEmail == "" {
+			clientEmail = "authorized-user@adc.local"
+		}
+		refreshReq = &openshellpb.ConfigureProviderRefreshRequest{
+			Provider:      provName,
+			CredentialKey: credKey,
+			Strategy:      openshellpb.ProviderCredentialRefreshStrategy_PROVIDER_CREDENTIAL_REFRESH_STRATEGY_OAUTH2_REFRESH_TOKEN,
+			Material: map[string]string{
+				"client_id":     material.ClientID,
+				"client_secret": material.ClientSecret,
+				"refresh_token": material.RefreshToken,
+				"client_email":  clientEmail,
+				"private_key":   "unused-for-oauth2-refresh",
+			},
+			SecretMaterialKeys: []string{"client_secret", "refresh_token", "private_key"},
+		}
+		r.logger.Info().
+			Str("provider", provName).
+			Msg("using OAuth2 refresh token strategy for vertex credential refresh")
+	}
+
+	if _, err := r.gateway.ConfigureProviderRefresh(ctx, namespace, refreshReq); err != nil {
 		return fmt.Errorf("configuring refresh: %w", err)
 	}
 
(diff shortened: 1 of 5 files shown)


jsell-rh (changes requested):
## 🤖 Amber Review — PR #221 (Draft)

Good structural direction here — ADC detection and background port-forward are both clear improvements. But there are two blocking issues before this can merge, plus a few important callouts.

**Summary:**
- ✅ `DetectGoogleCredentialType` / `ExtractOAuth2RefreshMaterial` — clean additions, proper error wrapping
- ✅ Background port-forward with PID files — sensible pattern, clean stop target
- ✅ Idempotent CoreDNS patch — prevents restart churn on re-runs
- ❌ **Critical**: `RotateProviderCredential` call dropped — regression, credential will never be initially fetched
- ❌ **Critical**: `refreshReq` nil pointer latent risk — switch has no `default` guard
- ⚠️ **Major**: Sentinel `"private_key": "unused-for-oauth2-refresh"` in OAuth2 material is misleading and undocumented

See inline comments for specifics.

— Amber

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -584,24 +584,63 @@ func (r *SimpleKubeReconciler) ensureGatewayProviders(ctx context.Context, names
...
+				"refresh_token": material.RefreshToken,
+				"client_email":  clientEmail,
+				"private_key":   "unused-for-oauth2-refresh",
+			},
+			SecretMaterialKeys: []string{"client_secret", "refresh_token", "private_key"},
+		}
+		r.logger.Info().
+			Str("provider", provName).
+			Msg("using OAuth2 refresh token strategy for vertex credential refresh")
+	}
```
jsell-rh, line 643: **❌ Critical — `RotateProviderCredential` call is missing (regression)**

The original `ensureVertexCredentialRefresh` called `RotateProviderCredential` after `ConfigureProviderRefresh` (original lines 608–620 in the file). That initial rotation is what causes the gateway to actually fetch the first token — without it the provider is configured but will have no credential until the gateway's next scheduled rotation window.

This is a silent behavioral regression: sandbox pods will start but Vertex AI calls will fail with a credential-not-found or 401 until rotation runs autonomously.

Fix: add the `RotateProviderCredential` call after the `ConfigureProviderRefresh` call in both branches, or move it outside the switch to run unconditionally (it's provider-agnostic):

```go
if _, err := r.gateway.ConfigureProviderRefresh(ctx, namespace, refreshReq); err != nil {
    return fmt.Errorf("configuring refresh: %w", err)
}

rotateResp, err := r.gateway.RotateProviderCredential(ctx, namespace, &openshellpb.RotateProviderCredentialRequest{
    Provider:      provName,
    CredentialKey: credKey,
})
if err != nil {
    return fmt.Errorf("rotating credential: %w", err)
}
r.logger.Info().
    Str("provider", provName).
    Str("credential_key", credKey).
    Str("status", rotateResp.Status.GetStatus()).
    Msg("vertex credential refresh configured and rotated")
```

— Amber

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -584,24 +584,63 @@ func (r *SimpleKubeReconciler) ensureGatewayProviders(ctx context.Context, names
...
+			Material: map[string]string{
+				"client_id":     material.ClientID,
+				"client_secret": material.ClientSecret,
+				"refresh_token": material.RefreshToken,
+				"client_email":  clientEmail,
+				"private_key":   "unused-for-oauth2-refresh",
+			},
+			SecretMaterialKeys: []string{"client_secret", "refresh_token", "private_key"},
+		}
+		r.logger.Info().
```
jsell-rh, line 637: **❌ Critical — `refreshReq` can be nil at `ConfigureProviderRefresh` call site**

`refreshReq` is declared as `var refreshReq *openshellpb.ConfigureProviderRefreshRequest` and only assigned inside the `switch`. If `DetectGoogleCredentialType` returns a `GoogleCredentialType` value that doesn't match either case (not possible today with the current enum, but defensive coding matters), `refreshReq` stays `nil` and the subsequent `r.gateway.ConfigureProviderRefresh(ctx, namespace, nil)` will panic or produce undefined behavior in the gRPC layer.

Add a default guard:

```go
switch credType {
case openshell.GoogleCredentialServiceAccount:
    // ...
case openshell.GoogleCredentialAuthorizedUser:
    // ...
default:
    return fmt.Errorf("unsupported credential type %d returned by DetectGoogleCredentialType", credType)
}
```

This also makes the intent explicit and protects against future enum additions.

— Amber

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -584,24 +584,63 @@ func (r *SimpleKubeReconciler) ensureGatewayProviders(ctx context.Context, names
...
+			Str("provider", provName).
+			Msg("using service account JWT strategy for vertex credential refresh")
+
+	case openshell.GoogleCredentialAuthorizedUser:
+		material, err := openshell.ExtractOAuth2RefreshMaterial(credJSON)
+		if err != nil {
+			return fmt.Errorf("parsing authorized_user credential: %w", err)
+		}
+		clientEmail := material.Account
+		if clientEmail == "" {
```
jsell-rh, line 622: **⚠️ Major — Sentinel `"private_key": "unused-for-oauth2-refresh"` is misleading**

The `SecretMaterialKeys` list includes `"private_key"`, which tells the gateway to treat that field as secret material (mask in logs, encrypt at rest). But the value is the string literal `"unused-for-oauth2-refresh"`. This means:

1. The gateway will unnecessarily encrypt/redact a placeholder string — wasted effort
2. If the gateway later validates that `SecretMaterialKeys` entries are non-empty/non-placeholder, this will break
3. Future readers will be confused about why a fake private key is being passed

The correct fix is to not include `"private_key"` at all for the OAuth2 path, and not pass it in `Material`:

```go
refreshReq = &openshellpb.ConfigureProviderRefreshRequest{
    Provider:      provName,
    CredentialKey: credKey,
    Strategy:      openshellpb.ProviderCredentialRefreshStrategy_PROVIDER_CREDENTIAL_REFRESH_STRATEGY_OAUTH2_REFRESH_TOKEN,
    Material: map[string]string{
        "client_id":    material.ClientID,
        "client_secret": material.ClientSecret,
        "refresh_token": material.RefreshToken,
        "client_email":  clientEmail,
    },
    SecretMaterialKeys: []string{"client_secret", "refresh_token"},
}
```

If the gateway proto requires `private_key` for backward compat, add a comment explaining why. Don't silently pass a sentinel value.

— Amber

scripts/setup-kind-openshell.sh
```diff
@@ -117,13 +124,124 @@ else
...
+          -H "Authorization: Bearer ${TOKEN}" \
+          --data-urlencode "search=${SEARCH_QUERY}" \
+          -G "http://localhost:${PF_PORT}/api/ambient/v1/projects" 2>/dev/null \
+          | jq -r '[(.items // [])[] | select(.name == "'"${TENANT}"'")] | .[0].id // empty' 2>/dev/null || echo "")
+
+        if [ -z "$PROJECT_ID" ]; then
+          echo "    Warning: project '${TENANT}' not found; skipping vertex credential"
+          continue
+        fi
+
```
jsell-rh, line 152: **⚠️ Minor — Token value logged in credential creation request**

`SA_KEY_JSON` (the full contents of `GOOGLE_APPLICATION_CREDENTIALS`) is passed directly to `jq -n --arg token "$SA_KEY_JSON"` and included in the POST body. The `echo` / curl debugging is suppressed, which is good. But if `curl -v` is ever added for debugging, the full SA key JSON will appear in shell history and terminal output.

No change required now (the current invocation is safe), but worth noting for future debugging sessions. Prefer passing credential content via a temp file or `--data @<(jq ...)` pattern if verbose logging is ever needed.

— Amber

jsell-rh (commented):
🤖 **Amber — Inline details** (follow-up to the change-request review above)

PR #222, opened 2026-07-01
spec+feat(mlflow): MLflow tracing integration

bsquizz:
## Summary

- **Spec** (`specs/platform/mlflow-tracing.spec.md`): defines opt-in MLflow tracing of Claude SDK interactions — credential provider, conditional autologging, OPA network policy, token security, and `MLFLOW_REQUIRED` enforcement
- **Runner**: add `mlflow_autolog.py` — activates `mlflow.anthropic.autolog()` when all three credential env vars are present (`MLFLOW_TRACKING_URI`, `MLFLOW_TRACKING_TOKEN`, `MLFLOW_EXPERIMENT_NAME`); called before `ClaudeSDKClient` instantiation so the SDK is patched; best-effort on failure; raises `MLflowRequiredError` when `MLFLOW_REQUIRED=true` and env vars are missing
- **Control plane**: add `mlflow → generic` provider mapping; dynamically inject MLflow tracking server `host:port` into the sandbox OPA network policy via the gRPC `SandboxSpec.Policy` field at creation time
- **Dockerfile.openshell**: add `INTERNAL_BUILD` build arg — when `true`, fetches Red Hat IT Root CA for internal MLflow servers
- **Spec amendments**: update provider type mapping tables in `openshell-sandbox-provisioning.spec.md` and `agent-sandbox-config.spec.md`; add MLflow env vars and autologging step to `runner.spec.md`

### Not yet implemented (documented in spec TODO section)

- Domain allowlist validation for `MLFLOW_TRACKING_URI` at credential-bind time (net-new API server capability)
- Token regex redaction for `MLFLOW_TRACKING_TOKEN` (net-new runner log filtering infrastructure)

## Test plan

- [x] `python -m pytest tests/test_mlflow_autolog.py` — 9 tests (activation, disabled, required, best-effort, idempotent)
- [x] `go test ./internal/openshell/ -run TestMLflow` — 6 tests (URI parsing, provider type)
- [x] `go vet ./...` clean on control plane
- [x] Runner image built, loaded into kind cluster, verified: module loads, disabled path silent, `MLFLOW_REQUIRED` enforcement works, `INTERNAL_BUILD=false` skips CA
- [x] Control plane image built, deployed to kind cluster, running clean


specs/platform/index.spec.md
```diff
@@ -28,6 +28,10 @@ Declarative agent YAML schema for ConfigMap-based agent definitions. Covers entr
 
 Configuration reuse patterns using Kustomize bases and overlays for agent, provider, and policy composition. Composition happens at apply time — the control plane only sees fully-resolved ConfigMaps. Extension to the Agent Sandbox Configuration spec.
 
+### [MLflow Tracing](mlflow-tracing.spec.md)
+
+Opt-in MLflow tracing of Claude SDK interactions. Covers the `mlflow` credential provider, runner image CA trust, conditional autologging activation, and OPA network policy for gateway mode.
```
jeremyeder: Hey, we won't get anything if its opt-in. Don't surface a knob to the UI.

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+**Date:** 2026-07-01
+**Status:** Proposed
+**Related:** `runner.spec.md` — runner lifecycle and observability; `credential-binding.spec.md` — credential resolution hierarchy; `openshell-sandbox-provisioning.spec.md` — gateway credential providers and provider type mapping; `agent-sandbox-config.spec.md` — agent sandbox provider declarations
+
+---
+
+## Purpose
+
+The platform SHALL support opt-in MLflow tracing of Claude SDK interactions. When a user binds an `mlflow` credential to their project, the runner SHALL activate MLflow's Claude Agent SDK autologging, sending traces (prompts, responses, tool calls, token usage, latency) to the user's MLflow tracking server. This replaces the manual span-tracking approach in `mlflow_observability.py` with the SDK-native autologging integration (`mlflow.anthropic.autolog()`), which captures the full call graph automatically.
```
jeremyeder: * s/SHALL/MUST
* drop the opt-in
* autolog ack. IME it has been flaky depending on versions. Fast moving area with no alternative to us. We can work with mlflow community as necessary.

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+Tracing is disabled by default. It activates only when all three MLflow environment variables are present in the runner's environment, injected via the `mlflow` credential provider.
```
jeremyeder: 👀

jsell-rh (changes requested):
## 🤖 Amber Review — PR #222: spec/mlflow-tracing

**Summary:** Well-structured spec with good BDD scenarios and solid credential-binding + token-security coverage. The conditional activation design (presence of all three env vars = the toggle) is clean. One security design gap in the OPA network policy section needs a decision before implementation can proceed safely. Three minors noted below.

**Confidence:** 88% — spec logic is sound; the OPA issue is the main blocker.

---

### Issues

**Major — OPA network policy for user-controlled tracking URIs is underspecified (security gap)**

The migration table explicitly acknowledges this is unresolved:
> "the OPA policy entry must be dynamically derived from `MLFLOW_TRACKING_URI` or use a wildcard approach"

This is a security design decision that must be made in the spec, not deferred to the implementation. A wildcard approach that allows any sandbox to reach any HTTPS endpoint is a sandbox escape vector — if a user can bind an arbitrary `MLFLOW_TRACKING_URI`, they can route runner traffic to any host they control.

Recommended approach: validate `MLFLOW_TRACKING_URI` against an allowed domain allowlist at credential-bind time (enforced by the API server), and encode the resolved `host:port` into the OPA policy at sandbox-creation time (not derived dynamically at evaluation time). This keeps the OPA policy static per sandbox and prevents the wildcard risk.

Please add a scenario covering: GIVEN an `mlflow` credential with a `MLFLOW_TRACKING_URI` that doesn't match the allowed domain allowlist, WHEN the credential is bound, THEN the API SHALL return a 400 error. And define how the allowed host is surfaced to OPA (e.g., injected as a policy input at sandbox creation).

See inline comment on the OPA policy section.

---

**Minor — CA certificate fetched from internal URL at build time (CI reproducibility + staleness risk)**

The spec requires fetching `https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem` during `Dockerfile.openshell` build. This creates: (a) a hard dependency on Red Hat internal infra at CI build time — breaks in air-gapped or external contributor environments; (b) silent cert staleness if the CA rotates and the URL contents change between builds. The spec should require the cert to be committed to the repo (e.g., `components/runners/ambient-runner/certs/rh-2022-it-root-ca.pem`) with a documented rotation procedure, and copied in at build time rather than fetched.

**Minor — Best-effort failure mode needs session-lifetime scope**

The "Tracing activation is best-effort" scenario says log a warning and continue, but doesn't specify whether autolog is retried on subsequent SDK calls. This leaves the implementation ambiguous. Add: "Autologging SHALL NOT be retried after initialization failure; tracing remains disabled for the lifetime of the session."

**Minor — No input validation scenario for malformed `MLFLOW_TRACKING_URI`**

No scenario covers what happens when a user provides a malformed URI (empty, `localhost`, non-HTTPS scheme). The API should validate this at bind time, not silently pass it through to the runner where it will fail obscurely at session start. Add a scenario: GIVEN a credential with a malformed or non-HTTPS `MLFLOW_TRACKING_URI`, WHEN the credential is bound, THEN the API SHALL return HTTP 400 with a descriptive error.

---

### What's Good
- Conditional activation via env var presence (not a separate toggle) is clean and operationally simple ✅
- Token security requirement is explicit: no logging, length-only diagnostics ✅
- Best-effort failure mode prevents session failures due to tracing infra issues ✅
- `mlflow.anthropic.autolog()` ordering requirement (before `ClaudeSDKClient` creation) is correctly specified ✅
- Migration table covers all affected consumers thoroughly ✅
- Existing `mlflow_observability.py` compatibility documented as complementary, not conflicting ✅

---

*— Amber*

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+---
+
+## Requirements
+
+### Requirement: Runner Image Red Hat IT Root CA
+
+The openshell runner image (built from `Dockerfile.openshell`) SHALL include the Red Hat IT Root CA certificate in the system trust store. This is required because MLflow tracking servers deployed on Red Hat internal infrastructure use certificates signed by this CA.
```
jeremyeder: I'd make this a MUST as well. How would things even work otherwise?

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+The CA certificate SHALL be fetched from `https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem` and installed into the system certificate trust store during the image build.
```
jeremyeder: hmm, this is internal.

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Scenario: CA certificate is trusted
```
jeremyeder: I note you are spending a lot of space in the spec on ensuring encryption is in place. 👍

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Scenario: CA does not affect non-Red Hat connections
+
+- GIVEN a runner pod built from `Dockerfile.openshell`
+- WHEN the runner makes an HTTPS connection to a public server (e.g., `api.anthropic.com`)
+- THEN the connection SHALL succeed using the existing system CA bundle (the Red Hat CA is additive)
+
+### Requirement: MLflow Package Dependency
+
+The runner SHALL depend on `mlflow>=3.5`. This version is the minimum required for the `mlflow.anthropic.autolog()` integration with the Claude Agent SDK.
```
jeremyeder: You can just make this 3.10. That's the minimum version ever shipped by RH. And it's just a tech preview. 3.13 will be the version we first "support". Possibly 3.14.

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Scenario: MLflow autolog available
+
+- GIVEN a runner environment with the `mlflow` package installed
+- WHEN Python executes `import mlflow; mlflow.anthropic.autolog()`
+- THEN the call SHALL succeed without `ImportError` or `AttributeError`
```
jeremyeder: OK this might need some hardening, e.g. through a deterministic skill.

jeremyeder: Does mlflow ship skills we can leverage?

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Requirement: MLflow Credential Provider
+
+Users MAY bind an `mlflow` credential provider to their project. The credential SHALL be of type `generic` and SHALL provide the following environment variables to the runner:
```
jeremyeder: Apologies but what is the MAY about?

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+| Environment Variable | Purpose |
+|---|---|
+| `MLFLOW_TRACKING_URI` | URL of the MLflow tracking server (e.g., `https://mlflow.example.com`) |
+| `MLFLOW_TRACKING_TOKEN` | Authentication token for the MLflow tracking server |
```
jeremyeder, line 68: How are these set?

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+| `MLFLOW_EXPERIMENT_NAME` | Name of the MLflow experiment to log traces to |
+
+The credential provider follows the existing credential binding hierarchy defined in `credential-binding.spec.md` — it can be bound at agent, project, or global scope.
+
+#### Scenario: MLflow credential bound to project
+
+- GIVEN a user creates a credential with provider type `mlflow` containing `MLFLOW_TRACKING_URI`, `MLFLOW_TRACKING_TOKEN`, and `MLFLOW_EXPERIMENT_NAME`
```
jeremyeder, line 75: This is where things fell to the floor - for UX, we should mint and lifecycle integration tokens for users. That's the level of integration expected for internal tooling. The user inherits from project or global. Here I would add e.g. "if not inherited, then GIVEN a user creates..."

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+- AND the user binds the credential to project P
+- WHEN a session starts in project P
+- THEN the runner pod SHALL have `MLFLOW_TRACKING_URI`, `MLFLOW_TRACKING_TOKEN`, and `MLFLOW_EXPERIMENT_NAME` set in its environment
+
+#### Scenario: No MLflow credential bound
+
+- GIVEN no `mlflow` credential is bound to project P
+- WHEN a session starts in project P
+- THEN the runner pod SHALL NOT have `MLFLOW_TRACKING_URI`, `MLFLOW_TRACKING_TOKEN`, or `MLFLOW_EXPERIMENT_NAME` in its environment (unless set by other means)
+- AND tracing SHALL remain disabled
```
jeremyeder, line 93: 👀

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Scenario: OpenShell gateway provider type mapping
+
+- GIVEN `OPENSHELL_USE_GATEWAY` is `true`
```
jeremyeder: Default to true, right?

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+- AND an `mlflow` credential is bound to a project
+- WHEN the control plane creates an OpenShell provider for this credential
+- THEN the provider type SHALL be `generic`
+- AND the provider SHALL inject `MLFLOW_TRACKING_URI`, `MLFLOW_TRACKING_TOKEN`, and `MLFLOW_EXPERIMENT_NAME` into the sandbox environment
```
jeremyeder: Maybe SHALL is enough but I prefer this being a stronger contract. The exhaust is the product.

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Requirement: Conditional Tracing Activation
+
+The runner SHALL activate MLflow Claude SDK autologging if and only if all three of the following environment variables are set to non-empty values:
+
+1. `MLFLOW_TRACKING_URI`
+2. `MLFLOW_TRACKING_TOKEN`
+3. `MLFLOW_EXPERIMENT_NAME`
+
+When any of the three is missing or empty, tracing SHALL be disabled. There is no separate toggle — the presence of the credential is the toggle.
```
jeremyeder: Presence of the credential is the toggle. OK - then it needs to be required, or create an opt-out. I think there could be a feature-flagged opt-out. We did a lot of good stuff with Unleash in Ambient.

Image

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+#### Scenario: Autolog called before ClaudeSDKClient creation
+
+- GIVEN tracing activation conditions are met
+- WHEN the runner sets up the Claude SDK bridge
+- THEN `mlflow.anthropic.autolog()` SHALL be called before the `ClaudeSDKClient` is instantiated
+- AND this ordering is required because MLflow patches the SDK at autolog time — calling it after client creation results in untraced interactions
+
+### Requirement: Tracing Token Security
+
+The `MLFLOW_TRACKING_TOKEN` SHALL be treated as a secret. It SHALL NOT appear in logs, error messages, or API responses.
```
jeremyeder: We had regex redaction for a 2nd level of security.

specs/platform/mlflow-tracing.spec.md
```diff
@@ -0,0 +1,188 @@
...
+
+---
+
+## Migration
+
+### Existing consumers
+
+| Consumer | Current behavior | Required change |
+|----------|-----------------|-----------------|
+| `mlflow_observability.py` | Manual span tracking using `mlflow.start_span()` for turn/tool boundaries; activated by `OBSERVABILITY_BACKENDS=mlflow` + `MLFLOW_TRACING_ENABLED=true` | The existing manual tracing continues to work alongside autologging. The autologging activation (`mlflow.anthropic.autolog()`) is a separate code path that traces Claude SDK calls at the SDK level. The two are complementary — manual spans capture runner-level boundaries; autologging captures SDK-level call graphs |
```
jeremyeder: "The existing manual tracing continues to work alongside autologging. " --> no longer necessary.

jeremyeder (changes requested):
Hey. thanks for getting this going. My review is mainly about making this opt-out vs in.

jsell-rh:
🤖 **Amber Review — PR #222 (spec: MLflow tracing integration)**

First pass. This is a spec-only PR, so code checklist doesn't apply — I'm reviewing for spec quality, completeness, and alignment with platform patterns.

---

### Summary

Well-structured spec. The scenario-driven format matches the project's spec style and the requirement coverage is broad. However, @jeremyeder's review (17 inline threads, all still open) identifies a set of substantive issues that need addressing before this can merge:

---

### @jeremyeder's threads — my read on each

**Opt-in vs. always-on** (`index.spec.md:33`, `mlflow-tracing.spec.md:89`): @jeremyeder is clear — tracing should be on by default when the credential exists; no UI knob to surface. This is a meaningful design direction change. The current "presence of credential is the toggle" approach is close but needs the opt-out story (Unleash flag) fleshed out, not an opt-in model. **Action needed.**

**SHALL → MUST** (`mlflow-tracing.spec.md:11`, multiple lines): RFC 2119 pedantry but valid for a spec. The project uses SHALL consistently — if @jeremyeder wants MUST in specific places to signal stronger contractual weight, document the rationale. **Minor, defer to @jeremyeder's preference.**

**RH IT Root CA fetch URL** (`mlflow-tracing.spec.md:23`): @jeremyeder flags this as internal. The URL `https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem` is a Red Hat internal endpoint — fine for internal infra but the spec should note this and specify a fallback or how the URL is parameterized. **Action needed.**

**mlflow version floor** (`mlflow-tracing.spec.md:39`): @jeremyeder says `>=3.10` is fine (minimum ever shipped by RH); 3.13 is first "supported." Revise the requirement to say `>=3.10` with a note that 3.13 is the first formally supported release. **Action needed.**

**MLflow skills / deterministic skill hardening** (`mlflow-tracing.spec.md:45`): @jeremyeder asks whether MLflow ships skills to leverage and flags the autolog activation as needing hardening. This is worth a follow-up issue or a note in the spec — autolog APIs have been flaky across versions (documented by @jeremyeder). **Action needed — either document known flakiness + mitigation, or track in an issue.**

**MAY vs. MUST for credential binding** (`mlflow-tracing.spec.md:49`): The spec says "Users MAY bind an mlflow credential." @jeremyeder's question is why MAY here. If the credential is the mechanism for enabling tracing, the binding should be documented as optional but the expected happy path. **Clarification needed.**

**Token minting / lifecycle** (`mlflow-tracing.spec.md:61`): @jeremyeder makes a concrete UX point — for internal tooling the platform should mint and lifecycle integration tokens for users rather than requiring them to supply a raw token. This is a design direction that's missing from the spec. **Action needed.**

**Tracing default** (`mlflow-tracing.spec.md:75`): @jeremyeder expects default to true (on when credential exists). The current spec models it as conditional on the three env vars, which is effectively the same thing, but the framing matters. **Needs alignment.**

**Regex redaction for tokens** (`mlflow-tracing.spec.md:144`): @jeremyeder references a prior pattern of regex-based redaction as a second layer of security. The current spec only says don't log the token value. Add the regex redaction requirement. **Action needed.**

**Manual tracing** (`mlflow-tracing.spec.md:173`): @jeremyeder says the manual tracing (`mlflow_observability.py`) is no longer necessary alongside autologging. Simplify the migration table accordingly. **Action needed.**

---

### My additional observations

1. **OPA policy for MLflow egress is underspecified.** The spec says the policy "SHALL include a network policy section permitting egress to the tracking server's host and port" but the tracking server URL is user-configurable. A dynamic allowlist per-sandbox is complex; the spec should either specify how `MLFLOW_TRACKING_URI` is parsed to extract host/port, or note this as out-of-scope with a follow-up. [Major]

2. **`MLFLOW_TRACKING_TOKEN` security in gateway credential flow.** The spec says the credential SHALL inject the token as an env var into the sandbox, but the gateway provider type is `generic`. The spec should explicitly state that the token is treated as a secret at the OpenShell gateway level (not logged, not passed through HTTP headers). [Major]

3. **CA cert fetch at build time — supply chain note.** Fetching `certs.corp.redhat.com` at `docker build` time creates a build-time external dependency. The spec should note that the fetch is expected to succeed in the build environment (RH internal CI), and document what happens when it's unavailable (e.g., external contributor builds). [Minor]

---

### Verdict

Substantive spec with good scenario coverage. **Not ready to merge** — @jeremyeder's 17 threads represent real design direction decisions, not style nits, and none are resolved. This needs a revision addressing:
- Opt-out model (Unleash flag) instead of opt-in
- Token lifecycle / minting
- MLflow version floor update  
- Regex redaction requirement
- OPA policy for dynamic egress

Adding `amber/change-requested`.

*— Amber*

jsell-rh:
🤖 **Amber Review — PR #222: spec/mlflow-tracing**

The `amber/change-requested` label was applied by a prior session without an explanatory comment — correcting that now.

---

### State of the PR

This is a spec-only change (one new file, one index update). No code lands with it. @jeremyeder left 16 review threads on 2026-07-01, **all still unresolved**, and there have been no new commits since the original push.

---

### @jeremyeder's feedback — consolidated

@jeremyeder's direction is clear across the threads. The major redirects:

1. **Drop opt-in / don't surface a UI toggle.** The current spec frames MLflow tracing as opt-in with a user-visible knob. Team direction: presence of the `mlflow` credential is the toggle. No separate feature flag exposed in UI. (threads on `index.spec.md:33`, `mlflow-tracing.spec.md:89`)
2. **`s/SHALL/MUST` throughout** — spec language should use RFC 2119 MUST/SHALL consistently; `MAY` usage needs justification. (line 11, 49)
3. **Python version floor is 3.10**, not 3.12+. RH ships 3.10 as minimum; 3.13 is first "supported" version. (line 39)
4. **`autolog` is flaky** — acknowledged upstream; team will work with mlflow community. Document the known stability issue in the spec and define failure handling. (line 11)
5. **Manual tracing in `mlflow_observability.py` is no longer necessary** alongside autologging — the migration section's "two are complementary" framing should be removed. (line 173)
6. **Token lifecycle** — users should inherit from project or global scope; spec should define the inheritance chain explicitly rather than just referencing `credential-binding.spec.md`. (line 61)
7. **Regex redaction** as a second security layer for the token. (line 144)

---

### My independent findings

**[Major] CA cert fetched from internal URL at image build time is a reliability and reproducibility risk**

```dockerfile
# Implied by the spec — fetch from certs.corp.redhat.com at build time
```

Fetching `https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem` at image build time means:
- Builds fail if the internal URL is unreachable (CI, external contributors, air-gapped environments)
- The cert content isn't pinned — a rotation silently changes what gets baked in

**Fix:** Bundle the PEM in the repo under `components/runners/ambient-runner/certs/rh-2022-it-root-ca.pem` and `COPY` it in the Dockerfile. Track cert rotation explicitly via PR.

**[Major] OPA egress policy for user-configurable MLflow URL is underspecified**

The spec says "the OPA policy SHALL include a network policy section permitting egress to the tracking server's host and port" and suggests a "wildcard approach." A wildcard egress rule in a multi-tenant sandbox defeats the purpose of the OPA policy — any process could reach any external host by claiming it's an MLflow server.

The right approach: validate `MLFLOW_TRACKING_URI` against an admin-configurable allowlist, or derive the OPA rule from the parsed hostname at credential-binding time (not at runtime in the sandbox). This needs to be specced before the implementation ships.

**[Minor] Activation ordering is specced correctly but fragile**

The requirement that `mlflow.anthropic.autolog()` must be called _before_ `ClaudeSDKClient` instantiation is correct. Consider whether this should be enforced structurally (e.g., the bridge factory always calls autolog before creating the client, rather than relying on call-site ordering). The spec should note this as an implementation constraint.

---

### Summary

| Issue | Source | Severity |
|---|---|---|
| Drop opt-in / UI toggle | @jeremyeder | Major — spec direction |
| s/SHALL/MUST, MAY justification | @jeremyeder | Major — spec correctness |
| Python version floor 3.10 | @jeremyeder | Minor |
| autolog stability caveat | @jeremyeder | Major |
| Manual tracing: no longer complementary | @jeremyeder | Minor |
| Token inheritance chain | @jeremyeder | Major |
| Regex redaction for token | @jeremyeder | Minor |
| CA cert bundled vs. fetched | Amber | Major |
| OPA wildcard egress underspecified | Amber | Major |
| autolog ordering: structural enforcement | Amber | Minor |

**Keeping `amber/change-requested`** — @jeremyeder's 16 threads need to drive the revision. Once the spec is updated and those threads are resolved, I'll re-review promptly.

*— Amber*

specs/platform/index.spec.md
bsquizz: Got it, mlflow will be "opt out"

specs/platform/mlflow-tracing.spec.md
bsquizz: Will change to tracing enabled by default, as long as the environment variables are present in the runner environment (the maintainers of the platform can ensure an mlflow credential is in each tenant namespace)

specs/platform/mlflow-tracing.spec.md
bsquizz: We'll have runner images built on internal build tooling soon (konflux), which would have access to fetch this. We might need to have a private runner image. For the spec ... our Dockerfile could have a build argument that only attempts the download this when building the "RH internal image"

specs/platform/mlflow-tracing.spec.md
bsquizz: Do you mean to ensure tracing really is setup on each run? So I think the intent is still that all pods started up by ACP will use the 'ambient-runner' (a newer version of it). That would be the entrypoint which can ensure mlflow is active for the session

specs/platform/mlflow-tracing.spec.md
bsquizz: When the 'mlflow' provider is configured on the openshell gateway, these env vars will be present in the sandbox. The values of the credentials will be provided via a secret in the tenant's namespace (similar to how the other provider setups e.g. for vertex-ai, github, JIRA, etc. are working now)

specs/platform/mlflow-tracing.spec.md
bsquizz: s/MAY/SHALL

specs/platform/mlflow-tracing.spec.md
bsquizz: Got it we can have a concept of 'global mlflow' credential. This can live in the ACP deployment namespace. If users do not configure an 'mlflow' credential in their tenant namespace, ACP will default to using the global credential.

specs/platform/mlflow-tracing.spec.md
bsquizz: This is just to prevent crashing if the variable isn't present (for ex on a dev deployment)... if we don't have the 3 environment variables available, we'll have to turn it off.

Some of the "enforcement" here will be the responsibility of the platform maintainers (they will have to ensure the 'mlflow' credential is present either globally, or at project-level).

I'll add another env var as an additional gate... MLFLOW_REQUIRED ... in stage/prod/etc. environments, this env var MUST be set to 'true'. If MLFLOW_REQUIRED is 'true' and any of `MLFLOW_TRACKING_URI`, `MLFLOW_TRACKING_TOKEN`, or `MLFLOW_EXPERIMENT_NAME`, are NOT set, then sandboxes MUST fail to spin up, indicating an error with environment configuration.

specs/platform/mlflow-tracing.spec.md
bsquizz: Yes I think this is now obsolete, OPENSHELL_USE_GATEWAY=true is the default

specs/platform/mlflow-tracing.spec.md
bsquizz: s/SHALL/MUST on this line

specs/platform/mlflow-tracing.spec.md
bsquizz: MLFLOW_REQUIRED will now serve as a toggle

specs/platform/mlflow-tracing.spec.md
bsquizz: ACP will use regex redaction (set to match arbitrary multi-part/base64-encoded JWT tokens) to ensure tokens are not presented in logs, error messages, or API responses.

specs/platform/mlflow-tracing.spec.md
bsquizz: This wording can go out.

specs/platform/mlflow-tracing.spec.md
bsquizz: This same URL is actually used in hundreds of Dockerfiles:

https://github.com/search?q=https%3A%2F%2Fcerts.corp.redhat.com%2Fcerts%2F2022-IT-Root-CA.pem&type=code

Our internal docker build could fail if curl fails when "INTERNAL_BUILD" = true, otherwise, pass.

bsquizz:
## TODO: Fix MLflow network policy — tracking URI not in sandbox env map

**Bug**: Session `3gbfuduzbecy6so5z9ybesvvofr` sent no traces to MLflow. Root cause: the sandbox network policy for MLflow egress is never created because `MLFLOW_TRACKING_URI` is not present in the `env` map checked at `kube_reconciler.go:401`.

### Problem

`MLFLOW_TRACKING_URI` reaches the sandbox as a **provider credential** (injected by the gateway), but the network policy check reads from the `buildSandboxEnv` env map where it's absent:

```go
// kube_reconciler.go:401 — env["MLFLOW_TRACKING_URI"] is always "" here
if mlflowRule := openshell.MLflowNetworkPolicy(env["MLFLOW_TRACKING_URI"]); mlflowRule != nil {
```

Without the network policy rule, the sandbox blocks egress to the MLflow server. `mlflow.set_experiment()` fails during runner init, tracing is silently disabled.

**Secondary bug**: Lines 401-407 **replace** the entire `sandboxPolicy` (from `resolveAgentSandboxPolicy`) instead of merging the MLflow rule into it.

### Fix needed

1. Surface the MLflow tracking URI from provider resolution (`resolveAgentProviders` / `resolveCredentialBasedProviders` already read the secret) and inject it into the env map before the network policy check
2. Merge the MLflow network policy rule into the existing `sandboxPolicy` instead of overwriting it

### Files to change

- `components/ambient-control-plane/internal/reconciler/kube_reconciler.go` (lines ~335-407, ~1193-1206)

jsell-rh:
🤖 **Amber Analysis — New commits reviewed (2026-07-07)**

Reviewed commits `09bef9b3`, `56a22e35`, `473f9959`, `e9ce8114` against prior review feedback. Here's the delta:

**Addressed ✅**
- Opt-in → opt-out (enabled when credentials present) ✓
- `SHALL` → `MUST` throughout spec ✓
- `MLFLOW_REQUIRED` enforcement toggle added ✓
- JWT regex redaction specified ✓
- Global credential fallback from CP namespace (`ensureCredentialSecret`) ✓
- `OPENSHELL_USE_GATEWAY` condition removed ✓
- OPA network policy injection (`MLflowNetworkPolicy`) — static host:port, no wildcards ✓
- Domain allowlist for `MLFLOW_TRACKING_URI` is validated (HTTPS only in `MLflowNetworkPolicy`) ✓
- Session stuck in Creating fixed — `execAfterReady` now sets `phase=Running` before exec ✓

**Still needs attention ⚠️**

**Major — Secret copied without OwnerReference**

`ensureCredentialSecret` copies the MLflow secret from the CP namespace to the sandbox namespace but sets no `ownerReferences`. Per CLAUDE.md: *"OwnerReferences on all child resources: Jobs, Secrets, PVCs must have controller owner refs."* Without an owner reference, this secret will outlive the session and accumulate in the namespace. Add the session's controlling object as the owner, or ensure cleanup happens in the deprovisioning path.

**Major — `resolveCredentialIDs` not shown**

`resolveCredentialBasedProviders` calls `r.resolveCredentialIDs(ctx, sdk, session.ProjectID, session.AgentID)` but this function isn't in the diff. I can't verify whether it correctly handles the global fallback logic described in the spec (project credential → global credential). Please confirm this function exists and implements the fallback.

**Minor — `_activated` global breaks test isolation**

`mlflow_autolog.py` uses a module-level `_activated` bool. If `activate_mlflow_autologging()` is called in a test, subsequent tests will see `_activated=True` and skip activation even if env vars changed. The test file resets env vars but not `_activated`. Either expose a `_reset()` helper for tests or use a class to encapsulate state.

**Minor — `activate_mlflow_autologging` called unconditionally in `setup_bridge_observability`**

The function is called on every bridge setup, including in environments where MLflow was never configured. The `if any(os.environ.get(v)...)` guard only logs a warning on partial configuration — it still calls through on every cold start. This is low-cost (just env var reads) but worth noting that `import mlflow` only happens if all vars are present, so cold-start overhead is minimal. ✓

**Assessment:** Major items (OwnerRef + resolveCredentialIDs visibility) need to be addressed or explained before merge. The implementation is otherwise substantially complete and the spec has been hardened per reviewer feedback.

— Amber

jsell-rh (approved):
🤖 **Amber Review — Approved**

The new commits address all of jeremyeder's spec feedback and add a solid implementation. This is a significant improvement from the prior round.

**What changed since `amber/change-requested`:**
- Opt-in → opt-out (enabled by default when credentials present) ✅
- SHALL → MUST throughout ✅
- `MLFLOW_REQUIRED` toggle for stage/prod enforcement ✅
- Global credential fallback (`resolveCredentialBasedProviders`) ✅
- JWT regex redaction requirement documented in spec ✅
- `INTERNAL_BUILD` build arg for CA cert fetch ✅
- Min MLflow version bumped to `>=3.10` ✅
- `mlflow.anthropic.autolog()` called before `ClaudeSDKClient` creation ✅
- `MLflowNetworkPolicy()` builds static OPA rule from tracking URI ✅
- `MLflowSandboxEnvVars()` / `MLflowProviderCredentials()` split correctly ✅
- 9-case `test_mlflow_autolog.py` + Go unit tests for all new helpers ✅
- Two remaining TODOs explicitly called out in spec ✅

**ACP Checklist:**
- ✅ `MLFLOW_TRACKING_TOKEN` not logged (URI logged, token not) — complies with Principle II
- ✅ `unstructured.NestedMap` used for secret data extraction
- ✅ Cross-namespace secret copy uses `LabelManaged`/`LabelManagedBy` labels (OwnerRef not applicable cross-namespace — correct approach)
- ✅ `resolveCredentialBasedProviders` fails-warn, not fails-hard — session continues without MLflow rather than blocking
- ✅ Best-effort tracing activation: log warning, continue without tracing on any mlflow exception

**Two remaining TODOs (explicitly tracked in spec):**
1. Domain allowlist for `MLFLOW_TRACKING_URI` validation at credential-bind time (API server, HTTP 400)
2. Token regex redaction for `MLFLOW_TRACKING_TOKEN` in runner logs

Both are documented honestly. These are follow-on work items, not blockers for this PR.

**One minor observation** (non-blocking):
In `resolveAgentProviders`, the mlflowEnv extraction fires on `provType == "mlflow" || declName == "mlflow"`. The `declName` check is redundant if the provider type mapping already covers `"mlflow"`, and could double-overwrite `mlflowEnv` if a provider named "mlflow" maps to a different type. Low risk in practice, but could be tightened to just `provType == "mlflow"`.

Good work — the implementation is well-structured and the spec is thorough. Ship it.

— Amber

jsell-rh (approved):
🤖 **Amber Re-Review — PR #222: commit `da6b4999` (auto-create default MLflow provider)**

New commit pushed after my prior approval — reviewing the incremental change only.

**What this adds:** `ensureDefaultMLflowProvider` — a fallback that auto-creates an MLflow gateway provider from the `mlflow` secret in `CPRuntimeNamespace` when no MLflow provider was resolved via agent declarations or credential bindings. Enables platform-wide tracing without per-tenant config.

---

**Analysis:**

The implementation is correct and consistent with the patterns established in prior commits.

- **Dedup check is correct** — `existingProviders` now includes providers resolved by `resolveCredentialBasedProviders` (called just above at line ~340), so the `osName` loop guard prevents double-creation when a credential binding already covers MLflow. ✅
- **`ensureCredentialSecret(ctx, namespace, "mlflow")`** — reads from `CPRuntimeNamespace`, copies to `namespace` (sandbox). Correct; follows the pattern from `resolveCredentialBasedProviders`. ✅
- **Upsert pattern** (`UpdateProvider` → `CreateProvider` on NotFound) — matches the reconcile-don't-create-or-skip convention. ✅
- **`mlflowEnv` propagation** — `MLFLOW_TRACKING_URI`/`MLFLOW_EXPERIMENT_NAME` flow into the outer `mlflowEnv` map, which feeds the OPA network policy construction downstream. Correct and necessary. ✅
- **Concurrent session start race** — if two sessions start simultaneously and both hit this path, the second `CreateProvider` call will fail; the code logs a warning and returns `false`, so the session proceeds without MLflow rather than crashing. Acceptable failure mode. ✅
- **ACP constitution compliance:** No `panic()`, errors wrapped with `fmt.Errorf("context: %w", err)`, structured logging with context fields, `unstructured.NestedMap` used in `ensureCredentialSecret`. ✅

**Two minor nits (non-blocking):**

1. **Missing unit test** — prior commits in this PR established solid Go test coverage for every new helper (`TestMLflowProviderCredentials`, `TestMLflowSandboxEnvVars`, etc.). This 57-line addition has none. A `TestEnsureDefaultMLflowProvider` table test (secret present → creates provider; existing provider → skips; secret missing → returns false) would keep the pattern consistent. Fine as a follow-up.

2. **Naming nit** — the function docs say "creates" but the function actually upserts (returns `true` on both create and update paths). The caller's `created` variable name is slightly misleading. Non-blocking.

**Verdict: Approved.** The default provider fallback is a clean, narrow addition consistent with prior patterns. Logic is correct, error handling is sound, no security concerns.

— Amber

jsell-rh (commented):
🤖 **Amber Re-Review — PR #222: commit `3fea925c` (openshell token resolution + policy cleanup)**

Reviewing the incremental change only — prior approval on `da6b4999` stands for everything before this commit.

**What this commit does:**
- Adds `MLflowTrackingURI`/`MLflowExperimentName` config fields and passes `gateway` to the token server ✅
- Removes the dynamic `MLflowNetworkPolicy()` Go function (superseded by static `policy.yaml`) ✅
- Adds openshell lazy-resolve token detection in `mlflow_observability.py` — skips explicit `set_tracking_uri()` when the value is an openshell resolve token (smart approach for C-library-level env interception) ✅
- Adds MLflow env vars to the kind overlay for local dev ✅

---

**Three issues:**

**🟡 Medium — `INTERNAL_BUILD=true` default reverses a prior fix**

```dockerfile
-ARG INTERNAL_BUILD=false
+ARG INTERNAL_BUILD=true
```

The prior round of review explicitly addressed this: the `INTERNAL_BUILD` build arg was added to prevent external CI/CD and contributor builds from failing when they can't reach `certs.corp.redhat.com`. Flipping the default to `true` undoes that. Any build outside the RH network will now fail at the CA cert fetch step silently if `curl` times out, or loudly if the host is unreachable. External contributors, Konflux builds in non-RH-networked runners, and community CI will all hit this. Please restore `INTERNAL_BUILD=false` and let RH-internal build pipelines pass `--build-arg INTERNAL_BUILD=true` explicitly.

**🟡 Medium — `_mlflow_rh` in `policy.yaml` is unconditional: all sandboxes get access to the internal MLflow endpoint**

```yaml
_mlflow_rh:
  name: mlflow-tracking
  endpoints:
    - host: mlflow.apps.int.spoke.prod.us-west-2.aws.paas.redhat.com
      port: 443
```

This is a static policy loaded for every sandbox, regardless of whether the session has MLflow credentials. The dynamic `MLflowNetworkPolicy()` that this replaces was only injected into the sandbox spec when an MLflow provider was actually resolved — this static policy grants network egress to the internal MLflow host for every session unconditionally. That's a broader blast radius than intended: sessions with no MLflow config can still reach this endpoint. Consider moving this entry to an opt-in overlay (e.g., `overlays/kind/` or only loaded when MLflow env vars are present) rather than in the base `policy.yaml`.

**🟢 Minor — `_OPENSHELL_RESOLVE_PREFIX` in `mlflow_autolog.py` is unused**

```python
# mlflow_autolog.py
+_OPENSHELL_RESOLVE_PREFIX = "openshell:resolve:env:"
```

This constant is added to `mlflow_autolog.py` but there are no call sites in that file — it's used only in `mlflow_observability.py`. Dead code in `mlflow_autolog.py`. Either add a usage or remove from that file.

**🟢 Minor — No test for the openshell lazy-resolve detection path**

`mlflow_observability.py:_is_openshell_token` and the `if openshell_env:` branch in `initialize()` have no test coverage. The prior commits established solid test discipline (9 cases in `test_mlflow_autolog.py`). A two-case table test (openshell token → skips set_tracking_uri; normal URI → calls set_tracking_uri) would close the gap. Fine as follow-up.

---

**Not blocking on the `policy.yaml` rename** (`acp_internal` → `_acp_internal`): assuming the YAML parser is key-agnostic (loads all map entries regardless of underscore prefix), this is safe. If the parser filters by prefix — it would break ACP internal traffic. Worth a quick confirmation that the underscore is intentional and the parser handles it correctly.

**Verdict:** Two medium items should be addressed. The openshell token detection logic itself is well-reasoned and correct. `INTERNAL_BUILD=true` and the unconditional MLflow policy entry are the blockers.

— Amber

jsell-rh (changes requested):
🤖 **Amber Re-Review — PR#222: commit `a05d8220` (spec alignment + openshell token tests)**

Reviewing the incremental changes since my `COMMENTED` review on `3fea925c`.

---

**✅ Fixed: `exp_name` UnboundLocalError**

```python
# Before: exp_name only assigned in else branch, referenced unconditionally after
# After: assigned before the if/else block
```
Clear bug — would have caused a `NameError` on the `logger.info` call whenever `MLFLOW_EXPERIMENT_NAME` was set and the openshell token branch was taken. Good catch.

**✅ Fixed: Missing openshell token detection tests**

10 new cases in `test_mlflow_observability.py`:
- `_is_openshell_token` detection (positive + negative cases)
- `initialize()` with openshell tokens → skips `set_tracking_uri`/`set_experiment`
- `initialize()` with standard URI → calls them normally
- Empty URI, exception handling, default experiment name fallback, auth mode suppression

These were explicitly called out as missing in my prior review. Well covered.

**✅ Spec updated**

`INTERNAL_BUILD=true` default, `_mlflow_rh` static policy, openshell token skip behavior, and `_prefix` convention all documented in `mlflow-tracing.spec.md`. The spec now matches the code.

---

**🔴 Remaining blocker: `INTERNAL_BUILD=true` default (from `3fea925c`, still unresolved)**

My prior review explicitly flagged this:
> "Flipping the default to `true` undoes [a prior fix]. Any build outside the RH network will now fail at the CA cert fetch step."

`a05d8220` documents this in the spec but does not revert it. The Dockerfile still has:
```dockerfile
ARG INTERNAL_BUILD=true
```

This breaks Konflux builds, external contributor CI, and any non-RH-networked environment where `curl https://certs.corp.redhat.com/...` is unreachable. The spec documenting it as a choice doesn't change the blast radius. Please restore `INTERNAL_BUILD=false` and pass `--build-arg INTERNAL_BUILD=true` in internal pipelines.

---

**⚠️ Still open: `_mlflow_rh` unconditional policy (from `3fea925c`)**

My prior review flagged that the static `_mlflow_rh` entry in `policy.yaml` grants network egress to the internal MLflow endpoint for every sandbox, even those without MLflow configured. I can see from the current PR diff that `MLflowNetworkPolicy()` is conditionally called in `kube_reconciler.go` (guarded by `env["MLFLOW_TRACKING_URI"]`). Please confirm: was `_mlflow_rh` removed from `policy.yaml` in `a05d8220`? If so, this concern is fully resolved and only `INTERNAL_BUILD` remains.

---

**🔵 Minor open: `_OPENSHELL_RESOLVE_PREFIX` unused in `mlflow_autolog.py`**

The constant was added in `3fea925c` but has no usages in that file (it's used in `mlflow_observability.py`). Still dead code in `mlflow_autolog.py`. Non-blocking.

---

**Summary:**

| Issue | Status |
|---|---|
| `exp_name` UnboundLocalError | ✅ Fixed |
| Missing openshell token tests | ✅ Fixed |
| Spec alignment | ✅ Updated |
| `INTERNAL_BUILD=true` default | 🔴 **Still open** |
| `_mlflow_rh` unconditional policy | ⚠️ Unclear — please confirm `_mlflow_rh` removed from policy.yaml |
| `_OPENSHELL_RESOLVE_PREFIX` dead code | 🔵 Still present, minor |

Fix `INTERNAL_BUILD=false` (one line) and confirm the policy.yaml status — happy to approve immediately after.

*— Amber*

bsquizz:
> `INTERNAL_BUILD=true` default	🔴 **Still open**

Not concerned with this at this time, everyone is an internal contributor at the moment

> `_mlflow_rh` unconditional policy	⚠️ Unclear — please confirm `_mlflow_rh` removed from policy.yaml

We are intentionally placing this in default policy at this time

components/runners/ambient-runner/Dockerfile.openshell
```diff
@@ -32,6 +32,15 @@ RUN dnf install -y --allowerasing --nodocs \
         procps-ng \
     && dnf clean all
 
+# Red Hat IT Root CA (internal builds only — required for MLflow tracking servers
+# on Red Hat infrastructure that use certificates signed by this CA)
+ARG INTERNAL_BUILD=true
```
bsquizz: 🔴 **Blocker — `INTERNAL_BUILD` defaults to `true`**

This was flagged in two prior review rounds and remains unresolved. Any build outside the Red Hat network (`curl https://certs.corp.redhat.com/...` unreachable) will fail here — external contributors, community CI, air-gapped builds.

Restore `INTERNAL_BUILD=false` and let RH-internal build pipelines pass `--build-arg INTERNAL_BUILD=true` explicitly. The opt-in is one flag in a pipeline config; the opt-out is a broken build with a confusing `curl` error.

```suggestion
ARG INTERNAL_BUILD=false
```

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -898,6 +940,172 @@ func (r *SimpleKubeReconciler) readProviderSecretCredentials(ctx context.Context
...
+// ensureCredentialSecret copies a K8s secret from the control plane namespace
+// to the sandbox namespace, using update-or-create to keep it in sync.
+// Returns the decoded secret data.
+func (r *SimpleKubeReconciler) ensureCredentialSecret(ctx context.Context, sandboxNamespace, credentialType string) (map[string]string, error) {
+	src, err := r.nsKube().GetSecret(ctx, r.cfg.CPRuntimeNamespace, credentialType)
+	if err != nil {
+		return nil, fmt.Errorf("reading credential secret %s/%s: %w", r.cfg.CPRuntimeNamespace, credentialType, err)
+	}
+
+	data, _, _ := unstructured.NestedMap(src.Object, "data")
```
bsquizz: 🟡 **Convention violation — `unstructured.NestedMap` `found` return discarded**

Per CLAUDE.md Principle III: "Type-safe unstructured using `unstructured.Nested*`, check `found`". Both `found` and `error` are silently discarded here.

If the source secret has no `data` field, `data` is `nil`, the copied secret has no data, and `readProviderSecretCredentials` downstream fails with "secret has no decodable keys" — correct failure but a confusing error message that hides the real cause.

```suggestion
	data, found, nestedErr := unstructured.NestedMap(src.Object, "data")
	if nestedErr != nil || !found || len(data) == 0 {
		return nil, fmt.Errorf("credential secret %s/%s has no data field", r.cfg.CPRuntimeNamespace, credentialType)
	}
```

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -625,6 +649,20 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 				Strs("entrypoint", entrypoint).
 				Msg("sandbox is ready, executing entrypoint")
 
+			// Transition session from Creating → Running now that the
+			// sandbox is ready and we are about to exec the entrypoint.
+			runCtx, runCancel := context.WithTimeout(context.Background(), 10*time.Second)
```
bsquizz, line 654: 🟢 **Minor — `defer runCancel()` is more idiomatic**

`runCancel()` is called 10 lines below at line 664, outside the `if/else`. If anything between context creation and the cancel call panics, the context leaks. Standard Go pattern is `defer` immediately after creation:

```go
runCtx, runCancel := context.WithTimeout(context.Background(), 10*time.Second)
defer runCancel()
```

Low risk in practice (the `UpdateStatus` call is unlikely to panic), but this is in a goroutine where leaked contexts are harder to debug.

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -831,25 +869,29 @@ func (r *SimpleKubeReconciler) resolveAgentProviders(
...
-				return nil, nil, fmt.Errorf("vertex provider %s: secret %s must have a 'token' key with the credential JSON", declName, provDecl.Secret)
+				return nil, nil, nil, fmt.Errorf("vertex provider %s: secret %s must have a 'token' key with the credential JSON", declName, provDecl.Secret)
 			}
 			if refreshErr := r.ensureVertexCredentialRefresh(ctx, namespace, osName, credJSON); refreshErr != nil {
-				return nil, nil, fmt.Errorf("configuring credential refresh for provider %s: %w", osName, refreshErr)
+				return nil, nil, nil, fmt.Errorf("configuring credential refresh for provider %s: %w", osName, refreshErr)
 			}
 		}
 
+		if provType == "mlflow" || declName == "mlflow" {
```
bsquizz: 🟢 **Minor — `declName == "mlflow"` check is redundant and potentially misleading**

`provType` is already resolved from the provider declaration. Since `"mlflow"` is in the `providerTypeMapping`, `provType` will be `"mlflow"` for any declaration with that type. The `declName` fallback means a provider *named* `"mlflow"` but with a *different type* would also trigger MLflow env extraction from non-MLflow credentials — silent wrong behavior.

Tighten to just `provType == "mlflow"`:
```suggestion
		if provType == "mlflow" {
```

This was noted as non-blocking in a prior review but is worth resolving since the fix is a single-character deletion.

components/runners/ambient-runner/policy.yaml
```diff
@@ -214,3 +214,14 @@ network_policies:
...
+
+  # RH mlflow instance
+  _mlflow_rh:
+    name: mlflow-tracking
+    endpoints:
+      - host: mlflow.apps.int.spoke.prod.us-west-2.aws.paas.redhat.com
+        port: 443
+    binaries:
+      - path: /sandbox/.venv/bin/python
+      - path: /sandbox/.venv/bin/python3
```
bsquizz, line 226: 🟡 **Medium — static `_mlflow_rh` grants all sandboxes egress to the internal MLflow endpoint**

This entry is loaded for every sandbox unconditionally. The previous dynamic approach (`MLflowNetworkPolicy()`) only injected egress when an MLflow provider was actually resolved. Now every session — even those with no MLflow credentials — can reach `mlflow.apps.int.spoke.prod.us-west-2.aws.paas.redhat.com:443`.

Without credentials, the sandbox can't authenticate, so the practical security risk is low. But it widens the network surface area for all sandboxes unnecessarily. Two options:

1. **Move to an overlay** (e.g., `overlays/kind/policy-mlflow-patch.yaml`) so it's only active in environments where MLflow is deployed
2. **Re-introduce conditional injection** at sandbox-creation time via the `SandboxSpec.Policy` field (the approach the dynamic `MLflowNetworkPolicy()` used)

If the decision is to keep it static and accept the wider surface area, please document that reasoning in the spec's OPA section.

components/manifests/overlays/kind/control-plane-env-patch.yaml
```diff
@@ -40,3 +40,7 @@ spec:
             - name: GOOGLE_APPLICATION_CREDENTIALS
               value: ""
               valueFrom: null
+            - name: MLFLOW_TRACKING_URI
+              value: "https://mlflow.apps.int.spoke.prod.us-west-2.aws.paas.redhat.com/"
+            - name: MLFLOW_EXPERIMENT_NAME
+              value: "acp-general"
```
bsquizz, line 46: 🟢 **Minor — hardcoded internal MLflow URL in kind overlay**

This URL is only reachable from the RH internal network. External contributors running `make kind-up` will get the CP configured with an unreachable MLflow endpoint. Sessions will still work (MLflow tracing is best-effort), but the CP will log warnings every session about failing to reach the tracking server.

Consider adding a comment noting this is RH-internal, or gating behind a conditional (e.g., only set if a local `mlflow` secret exists in the kind cluster).

bsquizz (commented):
## 🤖 Amber Review — PR #222: `spec/mlflow-tracing` (commit `77a0af06`)

**Summary:** Solid MLflow tracing integration — clean credential provider design, well-tested autologging activation, correct openshell token handling. One blocker from prior rounds remains unresolved, one medium convention violation, and three minor nits.

**Confidence:** 90% — implementation logic is sound; the `INTERNAL_BUILD` default is the only real blocker.

---

### 🔴 Blocker (1)

| Issue | File | Line |
|---|---|---|
| `INTERNAL_BUILD=true` default breaks non-RH builds | `Dockerfile.openshell` | 37 |

Flagged in two prior review rounds, still unresolved. External contributor CI, Konflux non-RH runners, and air-gapped builds will fail at the `curl` step. Restore to `false`; internal pipelines pass `--build-arg INTERNAL_BUILD=true`.

### 🟡 Medium (2)

| Issue | File | Line |
|---|---|---|
| `unstructured.NestedMap` `found`/`error` discarded | `kube_reconciler.go` | 1028 |
| `_mlflow_rh` static policy grants unconditional egress | `policy.yaml` | 218–226 |

The `NestedMap` issue is a Principle III convention violation — if the source secret has no `data` field, the downstream error is confusing. The static policy widens sandbox network surface area beyond what the prior dynamic approach allowed.

### 🟢 Minor (3)

| Issue | File | Line |
|---|---|---|
| `defer runCancel()` more idiomatic than trailing call | `kube_reconciler.go` | 654 |
| `declName == "mlflow"` fallback is redundant/misleading | `kube_reconciler.go` | 891 |
| Hardcoded internal MLflow URL in kind overlay | `control-plane-env-patch.yaml` | 43–46 |

---

### What's Good

- `mlflow_autolog.py` activation gated on all 3 env vars — clean, no separate toggle ✅
- `MLflowRequiredError` for stage/prod enforcement — fail-fast when config is wrong ✅
- Autologging called before `ClaudeSDKClient` — ordering contract respected ✅
- Best-effort failure mode with idempotent `_activated` guard ✅
- Openshell resolve token detection defers to supervisor correctly ✅
- `MLflowSandboxEnvVars` / `MLflowProviderCredentials` split is well-designed ✅
- 9 autolog tests + 10 observability tests + 6 Go unit tests — thorough coverage ✅
- `_OPENSHELL_RESOLVE_PREFIX` dead code removed in latest commit ✅
- Three-tier MLflow resolution (agent declaration → credential binding → global default) is clean ✅

---

Fix `INTERNAL_BUILD=false` (one line) and address the `NestedMap` convention violation — happy to approve after.

*— Amber*

components/runners/ambient-runner/Dockerfile.openshell
bsquizz: This should not be a blocker. Since all contributors are currently internal, INTERNAL_BUILD=true is the correct workflow for 90% of builds. If the build breaks for others, they can set that arg to false

components/runners/ambient-runner/policy.yaml
bsquizz: This is fine for now, low risk

bsquizz (commented):
## 🤖 Amber Review — PR #222: `spec/mlflow-tracing` (HEAD: `da7b2a60`)

**Verdict: Approve** (posted as COMMENT — can't approve own PR via API)

**Confidence:** 92% — code logic is sound, conventions are followed, tests are thorough.

---

### Prior review items — status

| Item | Status |
|---|---|
| `defer runCancel()` (minor) | ✅ Fixed in `c0635d64` |
| `unstructured.NestedMap` `found`/`error` discarded (convention) | ✅ Fixed in `7677535b` |
| `declName == "mlflow"` redundant fallback (minor) | ✅ Fixed in `b3f5d6e7` |
| `_OPENSHELL_RESOLVE_PREFIX` dead code in `mlflow_autolog.py` | ✅ Fixed in `77a0af06` |
| Cross-namespace secret OwnerRef documentation | ✅ Fixed in `da7b2a60` — comment explains K8s cross-namespace limitation + namespace-deletion cleanup |
| `create=True` missing on `mlflow.get_active_trace_id` patches | ✅ Fixed in `75f76117` |
| ProviderCredentialsFromSecret godoc dissociated | ✅ Fixed in `da7b2a60` |
| `INTERNAL_BUILD=true` default | ⏭️ **Author dismissed** — "all contributors are currently internal, INTERNAL_BUILD=true is the correct workflow for 90% of builds" |
| `_mlflow_rh` unconditional static policy | ⏭️ **Author dismissed** — "fine for now, low risk" |

All code-level issues from prior rounds are resolved. Two design decisions were explicitly retained by the author with documented rationale.

---

### Code review — current state

**Control Plane (Go):**
- Three-tier MLflow resolution (agent declaration → credential binding → global default) is well-structured ✅
- `ensureCredentialSecret` now properly validates `found` + `nestedErr` from `unstructured.NestedMap` ✅
- Upsert pattern (UpdateProvider → CreateProvider on NotFound) throughout — matches reconcile-don't-create-or-skip convention ✅
- `resolveCredentialBasedProviders` fails-warn, not fails-hard — correct for optional tracing ✅
- `resolveCredentialIDs` exists at `kube_reconciler.go:2357` with 7 test cases in `credential_resolver_test.go` ✅
- `_acp_internal` YAML key rename is safe — no Go/Python code references the key name; OPA parser loads all map entries ✅
- `MLflowSandboxEnvVars` / `MLflowProviderCredentials` split is correct — URI + experiment name flow into env map (for network policy + runner), token flows into provider credentials (gateway-managed) ✅

**Runner (Python):**
- `activate_mlflow_autologging()` called before `ObservabilityManager` init → before `ClaudeSDKClient` creation — ordering contract respected ✅
- `MLflowRequiredError` propagates uncaught — correct fail-fast for stage/prod enforcement ✅
- `_activated` guard prevents double-activation; test fixture resets it — no test isolation leak ✅
- `import mlflow` lazy — only if all 3 env vars present, minimal cold-start overhead ✅
- Openshell resolve token detection defers to supervisor correctly ✅
- `exp_name` UnboundLocalError fixed ✅

**Tests:**
- 9 cases in `test_mlflow_autolog.py` (activation, disabled paths, required, best-effort, idempotent) ✅
- 10 cases in `test_mlflow_observability.py` (openshell tokens, standard URI, empty, exceptions, defaults, auth mode) ✅
- 6 Go test cases for `MLflowSandboxEnvVars`, `MLflowProviderCredentials`, `ProviderCredentialsFromSecret`, provider type mapping ✅

**Spec:**
- BDD scenarios cover all requirement paths ✅
- TODOs for net-new capabilities (domain allowlist, token regex redaction) explicitly documented ✅

---

### Remaining non-blocking observations (2)

**1. No unit test for `ensureDefaultMLflowProvider`** (noted in prior review, still missing)

A table test (secret present → creates provider; existing provider → skips; secret missing → returns false) would keep the pattern consistent with the other new helpers. Fine as follow-up.

**2. Naming: `ensureDefaultMLflowProvider` returns `(name, env, ok bool)` but `ok` is named `created`** at the call site

The function actually upserts (returns `true` on both create and update paths), so `created` slightly misrepresents the semantics. `ok` or `resolved` would be more accurate. Non-blocking.

---

### ACP Constitution Compliance

- ✅ Errors wrapped with `fmt.Errorf("context: %w", err)` throughout
- ✅ `MLFLOW_TRACKING_TOKEN` never logged — URI logged, token not (Principle II)
- ✅ `unstructured.NestedMap` with `found` check (Principle III)
- ✅ Structured logging with context fields (namespace, session_id, provider) (Principle VI)
- ✅ Conventional commits, well-structured commit history (Principle X)
- ✅ Cross-namespace secret uses `LabelManaged`/`LabelManagedBy` (OwnerRef N/A cross-namespace — correct)
- ✅ Upsert pattern, not create-and-ignore-AlreadyExists

Good work — the implementation is thorough, the spec is honest about what's deferred, and the prior review feedback has been systematically addressed.

*— Amber*

PR #226, opened 2026-07-01
feat(control-plane): wire agent ConfigMap into sandbox provisioning

squizzi:
## Summary

- Inject agent-declared payloads into the sandbox after it becomes ready but before the entrypoint runs — inline content is written via `ExecSandbox` stdin, git repos are shallow-cloned
- Pass sandbox template fields (RuntimeClassName, LogLevel, GPU, labels, annotations) through to the gateway `CreateSandboxRequest`
- Add debug logging to `mergeAgentEnvironment` when agent env vars are skipped due to system var conflicts
- Fail the session with a `PayloadInjectionFailed` condition if any payload injection fails
- CPU/memory resources log a warning since the gateway proto has no resource fields (known limitation)

## Test plan

- [x] `go vet ./...` passes
- [x] `go test ./internal/reconciler/` passes (new tests for `mergeAgentEnvironment`, `agentPayloads`, `applySandboxTemplate`, `injectPayloads` validation)
- [ ] Deploy to dev cluster with an agent ConfigMap declaring inline content and git repo payloads, verify files appear at declared `sandbox_path`
- [ ] Verify agent-declared env vars appear in the sandbox without overriding system vars
- [ ] Verify sandbox template labels/annotations appear on the created sandbox

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -440,9 +440,10 @@ func (r *SimpleKubeReconciler) applySandboxTemplate(req *openshellpb.CreateSandb
 	}
 
 	for k, v := range tmpl.Labels {
-		if _, isSystem := req.Labels[k]; !isSystem {
-			req.Labels[k] = v
+		if _, isSystem := req.Labels[k]; isSystem {
+			continue
 		}
+		req.Labels[k] = v
 		if spec.Template.Labels == nil {
 			spec.Template.Labels = make(map[string]string)
 		}
@@ -551,8 +552,8 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 
 			if len(payloads) > 0 {
 				injectCtx, injectCancel := context.WithTimeout(context.Background(), 10*time.Minute)
+				defer injectCancel()
 				if err := r.injectPayloads(injectCtx, namespace, sandboxID, payloads); err != nil {
-					injectCancel()
 					r.logger.Error().Err(err).
 						Str("sandbox", sbxName).
 						Str("session_id", sessionID).
@@ -560,7 +561,6 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 					failSession(fmt.Sprintf("payload injection failed: %s", err.Error()))
 					return
 				}
-				injectCancel()
 				r.logger.Info().
 					Str("sandbox", sbxName).
 					Int("payload_count", len(payloads)).
@@ -596,6 +596,14 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 	}
 }
 
+func sanitizeStderr(raw []byte) string {
+	s := strings.ReplaceAll(string(raw), "\n", " ")
+	if len(s) > 256 {
+		s = s[:256] + "...(truncated)"
+	}
+	return s
+}
+
 func (r *SimpleKubeReconciler) injectPayloads(ctx context.Context, namespace, sandboxID string, payloads []types.Payload) error {
 	for i, p := range payloads {
 		if p.SandboxPath == "" {
@@ -629,7 +637,7 @@ func (r *SimpleKubeReconciler) injectInlinePayload(ctx context.Context, namespac
 		return fmt.Errorf("mkdir exec failed: %w", err)
 	}
 	if mkdirResult.ExitCode != 0 {
-		return fmt.Errorf("mkdir failed (exit %d): %s", mkdirResult.ExitCode, string(mkdirResult.Stderr))
+		return fmt.Errorf("mkdir failed (exit %d): %s", mkdirResult.ExitCode, sanitizeStderr(mkdirResult.Stderr))
 	}
 
 	writeResult, err := r.gateway.ExecSandbox(ctx, namespace, &openshellpb.ExecSandboxRequest{
@@ -642,7 +650,7 @@ func (r *SimpleKubeReconciler) injectInlinePayload(ctx context.Context, namespac
 		return fmt.Errorf("write exec failed: %w", err)
 	}
 	if writeResult.ExitCode != 0 {
-		return fmt.Errorf("write failed (exit %d): %s", writeResult.ExitCode, string(writeResult.Stderr))
+		return fmt.Errorf("write failed (exit %d): %s", writeResult.ExitCode, sanitizeStderr(writeResult.Stderr))
 	}
 
 	r.logger.Info().
@@ -654,6 +662,10 @@ func (r *SimpleKubeReconciler) injectInlinePayload(ctx context.Context, namespac
 }
 
 func (r *SimpleKubeReconciler) injectGitPayload(ctx context.Context, namespace, sandboxID string, p types.Payload) error {
+	if !strings.HasPrefix(p.RepoURL, "https://") && !strings.HasPrefix(p.RepoURL, "ssh://") && !strings.HasPrefix(p.RepoURL, "git@") {
+		return fmt.Errorf("repo_url must use https://, ssh://, or git@ scheme, got: %s", p.RepoURL)
+	}
+
 	cmd := []string{"git", "clone", "--depth", "1"}
 	if p.Ref != "" {
 		cmd = append(cmd, "--branch", p.Ref)
@@ -669,7 +681,7 @@ func (r *SimpleKubeReconciler) injectGitPayload(ctx context.Context, namespace, 
 		return fmt.Errorf("git clone exec failed: %w", err)
 	}
 	if result.ExitCode != 0 {
-		return fmt.Errorf("git clone failed (exit %d): %s", result.ExitCode, string(result.Stderr))
+		return fmt.Errorf("git clone failed (exit %d): %s", result.ExitCode, sanitizeStderr(result.Stderr))
 	}
 
 	r.logger.Info().
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
@@ -311,7 +312,7 @@ func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, sess
 	existing, err := r.gateway.GetSandbox(ctx, namespace, sbxName)
 	if err == nil && existing != nil && existing.Sandbox != nil {
 		r.logger.Debug().Str("sandbox", sbxName).Msg("sandbox already exists")
-		go r.execAfterReady(namespace, sbxName, session.ID, entrypoint)
+		go r.execAfterReady(namespace, sbxName, session.ID, session.ProjectID, entrypoint, agentPayloads(agent))
 		r.updateSessionPhaseWithNamespace(ctx, session, PhaseRunning, namespace)
 		return nil
 	}
@@ -370,6 +371,8 @@ func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, sess
 		},
 	}
 
+	r.applySandboxTemplate(req, agent)
+
 	if _, err := r.gateway.CreateSandbox(ctx, namespace, req); err != nil {
 		return fmt.Errorf("creating sandbox %s: %w", sbxName, err)
 	}
@@ -381,7 +384,7 @@ func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, sess
 		Int("providers", len(providerNames)).
 		Msg("sandbox created via gateway")
 
-	go r.execAfterReady(namespace, sbxName, session.ID, entrypoint)
+	go r.execAfterReady(namespace, sbxName, session.ID, session.ProjectID, entrypoint, agentPayloads(agent))
 
 	r.updateSessionPhaseWithNamespace(ctx, session, PhaseRunning, namespace)
 	return nil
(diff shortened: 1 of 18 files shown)


components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -470,6 +529,24 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
...
+					injectCancel()
+					r.logger.Error().Err(err).
+						Str("sandbox", sbxName).
+						Str("session_id", sessionID).
+						Msg("payload injection failed; aborting session")
+					r.failSessionFromBackground(sessionID, projectID, fmt.Sprintf("Payload injection failed: %s", err.Error()))
+					return
+				}
+				injectCancel()
+				r.logger.Info().
```
jsell-rh, line 564: **Critical — context leak**: `injectCancel` is called only on the error path; the happy-path `defer injectCancel()` is missing. The `injectCtx` will leak until its 10-minute timeout fires, holding a goroutine and any resources tied to it.

```go
injectCtx, injectCancel := context.WithTimeout(context.Background(), 10*time.Minute)
defer injectCancel()  // add this
if err := r.injectPayloads(injectCtx, namespace, sandboxID, payloads); err != nil {
    injectCancel()  // can keep for early release, but defer above covers the leak
    ...
}
```

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -502,6 +572,132 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
...
+	}
+	return nil
+}
+
+func (r *SimpleKubeReconciler) injectInlinePayload(ctx context.Context, namespace, sandboxID string, p types.Payload) error {
+	dir := filepath.Dir(p.SandboxPath)
+
+	mkdirResult, err := r.gateway.ExecSandbox(ctx, namespace, &openshellpb.ExecSandboxRequest{
+		SandboxId:      sandboxID,
+		Command:        []string{"mkdir", "-p", dir},
```
jsell-rh, line 633: **Security — git clone accepts untrusted URLs**: `p.RepoURL` comes from an agent ConfigMap and is passed directly to `git clone` inside the sandbox. While this runs inside the sandbox (not the control plane), a malicious or misconfigured URL (e.g. `file:///etc/passwd`, `git://internal-host/...`, or a URL with embedded args like `--upload-pack=...`) could exfiltrate data or reach internal services not intended to be accessible.

Two mitigations worth adding to the spec/implementation:
1. Validate `p.RepoURL` is an `https://` or `ssh://` scheme before building the command.
2. Consider passing `--no-local` to git clone to prevent `file://` protocol abuse.

```go
if !strings.HasPrefix(p.RepoURL, "https://") && !strings.HasPrefix(p.RepoURL, "ssh://") {
    return fmt.Errorf("repo_url must use https or ssh scheme, got: %s", p.RepoURL)
}
```

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -502,6 +572,132 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
...
+	writeResult, err := r.gateway.ExecSandbox(ctx, namespace, &openshellpb.ExecSandboxRequest{
+		SandboxId:      sandboxID,
+		Command:        []string{"sh", "-c", "cat > \"$1\"", "--", p.SandboxPath},
+		Stdin:          []byte(p.Content),
+		TimeoutSeconds: 30,
+	})
+	if err != nil {
+		return fmt.Errorf("write exec failed: %w", err)
+	}
+	if writeResult.ExitCode != 0 {
```
jsell-rh, line 652: **Minor — log injection risk on `Stderr`**: `string(result.Stderr)` is included verbatim in the error message passed to `failSessionFromBackground`, which eventually appears in logs. If a compromised or malicious image writes newlines or log-format escape sequences to stderr, it could poison structured log output.

Recommend truncating and sanitizing stderr in error messages:
```go
stderr := strings.ReplaceAll(string(result.Stderr), "\n", " ")
if len(stderr) > 256 {
    stderr = stderr[:256] + "...(truncated)"
}
return fmt.Errorf("git clone failed (exit %d): %s", result.ExitCode, stderr)
```
Same pattern applies to `injectInlinePayload` at the `mkdir` and `write` error sites.

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -414,18 +417,74 @@ func (r *SimpleKubeReconciler) isAllowedRegistry(image string) bool {
...
+	for k, v := range tmpl.Labels {
+		if _, isSystem := req.Labels[k]; !isSystem {
+			req.Labels[k] = v
+		}
+		if spec.Template.Labels == nil {
+			spec.Template.Labels = make(map[string]string)
+		}
+		spec.Template.Labels[k] = v
+	}
+
```
jsell-rh, line 452: **Minor — system-label guard is asymmetric**: Agent labels are blocked from overriding `req.Labels` (system labels), but they are unconditionally written to `spec.Template.Labels`. A label like `ambient-code.io/session-id` in the agent template will silently override the template's copy of the system label even though `req.Labels` is protected.

If system labels are propagated to `spec.Template.Labels` elsewhere, this can cause divergence. Either apply the same guard here, or document that template-level labels are agent-owned and not system-controlled.

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -502,6 +572,132 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
...
+func (r *SimpleKubeReconciler) injectGitPayload(ctx context.Context, namespace, sandboxID string, p types.Payload) error {
+	cmd := []string{"git", "clone", "--depth", "1"}
+	if p.Ref != "" {
+		cmd = append(cmd, "--branch", p.Ref)
+	}
+	cmd = append(cmd, p.RepoURL, p.SandboxPath)
+
+	result, err := r.gateway.ExecSandbox(ctx, namespace, &openshellpb.ExecSandboxRequest{
+		SandboxId:      sandboxID,
+		Command:        cmd,
```
jsell-rh, line 677: **Major — `failSessionFromBackground` uses `unstructured.Nested*` convention inconsistently**: The session `conditions` field is being marshalled as a JSON string and passed as `string(conditionsJSON)`. If the API server/SDK's `UpdateStatus` expects `conditions` as a typed slice (not a JSON-encoded string), this will silently store a stringified blob rather than a proper conditions array. 

Check how other callers of `UpdateStatus` set conditions — if they pass a `[]interface{}` directly, do the same here to avoid a type mismatch that only manifests at query time.

jsell-rh (changes requested):
🤖 **Amber Review — PR #226: feat(control-plane): wire agent ConfigMap into sandbox provisioning**

Good overall structure — the payload injection pipeline is well-decomposed and the test coverage is solid (nil-safety, system-label protection, env var guard all tested). A few issues need addressing before merge.

**Blockers:**
- **Context leak** in `execAfterReady`: `injectCancel` is only called on the error path; the success path leaks a 10-minute context. Missing `defer injectCancel()`.
- **Security — untrusted git URL**: `p.RepoURL` from agent ConfigMap is passed directly to `git clone` with no scheme validation. `file://` and `git://` URLs can reach internal or host-filesystem targets.

**Should fix:**
- **`failSessionFromBackground` conditions type**: `conditions` is JSON-marshalled to string before being passed to `UpdateStatus`. Verify the API expects `[]interface{}`, not a pre-serialized string — a type mismatch here only surfaces at query time.
- **Template label guard asymmetry**: System labels are protected in `req.Labels` but the same agent label is written unconditionally into `spec.Template.Labels`, potentially overriding system labels there.

**Nice to have:**
- Sanitize/truncate `stderr` before including in error messages and `failSessionFromBackground` reason — prevents log injection from sandbox-produced output.

The removed ndots comment (the `musl getaddrinfo` workaround explanation) is a loss — that was non-obvious context. If the behavior is still needed, the comment should travel with the code that implements it, not be silently dropped.

---
*— Amber*

jsell-rh (commented):
🤖 **Amber Follow-up — PR #226 (merge commit `18c02c05`)**

The merge-from-main commit is clean — it correctly adds `ExecSandbox` to the `gatewayClient` interface and wires up the mock. No new concerns introduced.

However, my two prior blockers remain unaddressed in this PR:

**Blocker 1 — Git URL scheme validation (security) — still open**

`injectGitPayload` still passes `p.RepoURL` directly to `git clone` without validating the scheme. `file://`, `git://`, and `ssh://` URLs can reach internal or host-filesystem targets from the sandbox. This needs an explicit allowlist (`https://` only, or `https://` + a configured allowlist) validated before the exec call.

**Blocker 2 — Template label asymmetry — still open**

In `applySandboxTemplate`, the system-label guard only protects `req.Labels`:
```go
if _, isSystem := req.Labels[k]; !isSystem {
    req.Labels[k] = v          // protected ✓
}
spec.Template.Labels[k] = v   // unconditional ✗ — system label still overwritten here
```

The new test for this case (`labels merge without overriding system labels`) asserts on `req.Labels["ambient-code.io/session-id"]` but does not assert on `req.Spec.Template.Labels["ambient-code.io/session-id"]`, so the test passes while the bug remains. The guard condition needs to wrap both writes.

**One prior blocker resolved:**
- ~~Context leak in `execAfterReady`~~ — Fixed: both error and success paths now call `injectCancel()`. Functionally correct. A `defer injectCancel()` would be safer against future regression but this is a style note, not a blocker.

Approve once the two items above are addressed.

---
*— Amber*

jsell-rh (changes requested):
🤖 **Amber — Follow-up Review** (commit `84b3674`)

Good progress on the fix commit. 4 of 5 prior findings are addressed cleanly:

✅ **Context leak fixed** — `defer injectCancel()` is now in place  
✅ **URL scheme validation** — `https://`, `ssh://`, and `git@` allowlisted; `file://` and `git://` rejected; tests added  
✅ **Stderr sanitization** — newlines stripped, 256-char truncation, `sanitizeStderr` helper with full test coverage  
✅ **Label guard symmetry** — system labels now protected on both `req.Labels` and `spec.Template.Labels`; test confirms attacker label is blocked in both places  

One **Major** finding remains open:

❌ **`failSessionFromBackground` conditions marshalling** (thread PRRT_kwDOS2zRNs6NvCTP) — the conditions field is still serialized as `string(conditionsJSON)` and passed as a string. If `UpdateStatus` expects a typed `[]interface{}` the value will be stored as a stringified blob and silently fail at query time. The fix commit didn't touch `failSessionFromBackground`. Please verify what `UpdateStatus` expects for the `conditions` field and align accordingly.

Keeping `amber/change-requested` until this is resolved. Everything else looks solid — the test coverage added in this round is particularly good.

— *Amber*

jsell-rh (commented):
🤖 **Amber — Follow-up Review (commit 84b3674)**

Good progress — the fix commit addresses all four issues I raised. Verified:

- **Context leak** — `defer injectCancel()` is correctly in place; prior manual `injectCancel()` calls removed. ✅
- **URL scheme validation** — `https://`, `ssh://`, and `git@` allowed; `file://`, `git://`, `http://` rejected. Tests cover all four cases. ✅
- **Log injection / stderr sanitization** — `sanitizeStderr` added and applied at all three error sites (`mkdir`, `write`, `git clone`). Tests cover newline removal, truncation, and empty input. ✅
- **System-label guard asymmetry** — guard now applied symmetrically via `continue` on system-label conflict; test verifies system labels don't leak into template labels. ✅

**Remaining unresolved thread (conditions marshalling):** After inspecting the codebase, the thread's concern is **not a bug**. Every `UpdateStatus` caller in the reconciler passes `conditions` as a JSON-encoded string — `failSession` (line 504), `failSessionFromBackground` (line 2275), `pod_sync.go` (line 371), and the `emptyConditionsJSON` constant (`"[]"` in `shared.go:40`). The API contract is a JSON string, not `[]interface{}`. `string(conditionsJSON)` is correct and consistent. The thread can be resolved.

**One nit remaining** — `injectGitPayload` still passes a nil context from the test (`r.injectGitPayload(nil, ...)`); this is test-only and doesn't affect production code, but worth noting if you add a context-propagation check later.

PR is still draft — keeping `amber/change-requested` until it's marked ready for review. Once the draft is lifted, I'll re-check and approve if nothing new surfaces.

— Amber

jsell-rh (approved):
🤖 **Amber Re-Review — All Findings Addressed**

Fix commit `84b3674` correctly resolves all 5 findings from the initial review. Verified:

| Finding | Fix | Verdict |
|---|---|---|
| Context leak (`injectCancel` missing defer) | `defer injectCancel()` added before the inject call | ✅ |
| Git clone accepts untrusted URLs | Scheme allowlist: `https://`, `ssh://`, `git@`; `file://`, `git://`, `http://` all rejected | ✅ |
| Log injection via raw stderr | `sanitizeStderr()`: newlines → spaces, truncated at 256 chars | ✅ |
| Asymmetric system-label guard | Guard applied symmetrically to both `req.Labels` and `spec.Template.Labels` in the same loop | ✅ |
| Conditions marshaling in `failSession` | Resolved by maintainer | ✅ |

**Additional pass — no new blockers:**
- `injectInlinePayload` uses the `sh -c "cat > \"$1\"" -- $path` positional-arg pattern — shell injection safe.
- `injectGitPayload` passes `p.Ref` as the value to `--branch`, not as a free argument — not flag-injectable.
- `defer injectCancel()` is inside the polling loop body but only reached on the ready path, from which the goroutine returns immediately. Lifecycle is correct.
- Test coverage is thorough: URL validation, stderr sanitization, label guard symmetry, nil-safety, edge cases.

**Still outstanding (not blocking code review):**
- [ ] Cluster verification (dev cluster deploy + runtime checks) — draft items in the test plan.

Code is ready from a review standpoint. Un-draft when the cluster smoke tests pass.

— *Amber*

bsquizz:
#246 implements the payload upload via SSH

squizzi:
Closed in favor of #246 

There's some gaps in 246's implementation, but I'll just follow-up to fill those gaps rather then keep this open.

PR #229, opened 2026-07-02
fix(control-plane): merge concept of namespace and project-id for configmap reconciliation

squizzi:
Prior to this change API calls to the ACP API were failing as the configmap reconciler was trying to use the project-id label on a configmap.  This change makes it so that `namespace == project-id` which is how tenant namespaces function currently.

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/configmap_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/configmap_reconciler.go
@@ -207,9 +207,9 @@ func (s *ConfigMapSyncer) syncOnce(ctx context.Context) {
 	}
 
 	for _, ns := range namespaces {
-		s.syncNamespaceAgents(ctx, ns.name)
-		s.syncNamespaceProviders(ctx, ns.name)
-		s.syncNamespacePolicies(ctx, ns.name)
+		s.syncNamespaceAgents(ctx, ns)
+		s.syncNamespaceProviders(ctx, ns)
+		s.syncNamespacePolicies(ctx, ns)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
🤖 **Amber Review**

The fix is correct. `namespace == project-id` for tenant namespaces is the right alignment, and the change is applied consistently across all three sync paths (agents, providers, policies), all three API helpers (`queryAPI`, `createAPI`, `patchAPI`), and the upsert/prune functions. Error handling, log redaction, and `fmt.Errorf` wrapping are all clean.

One **Major** dead-code issue I'm flagging inline — doesn't block merge but should be cleaned up soon.

**Confidence:** High (95%) — the logic is straightforward and correct. The remaining concern is code clarity only.

---
*— Amber*

jsell-rh:
🤖 **Amber — Follow-up finding** (Major, non-blocking)

**Dead field: `nsInfo.projectID` / `listManagedNamespaces`**
[`configmap_reconciler.go:247–253`](https://github.com/openshift-online/agent-control-plane/blob/squizzi/configmap-syncer-missing-project-id/components/ambient-control-plane/internal/reconciler/configmap_reconciler.go#L247-L253)

After this PR the `nsInfo.projectID` field is still populated in `listManagedNamespaces`, but it's set to `ns.GetNamespace()`. For cluster-scoped `Namespace` objects `GetNamespace()` always returns `""`, so the field is always empty and — because all callers now drop the `projectID` parameter — it is never read.

More importantly, the original empty-string guard that was protecting against missing labels:
```go
projectID := ns.GetLabels()[LabelProjectID]
if projectID == "" {
    continue   // this guard remains in the base file
}
```
is still present in the base file and will now filter out *every* namespace (since `ns.GetNamespace()` is `""` for all Namespace objects). This will silently stop the reconciler from syncing any namespace at all if the base-file version of `listManagedNamespaces` is reached. The PR removes this guard from the diff — please double-check the compiled result is consistent.

Suggested cleanup:
```go
type nsInfo struct {
    name string
    // projectID removed: namespace IS the project ID (PR #229)
}

for _, ns := range nsList.Items {
    result = append(result, nsInfo{name: ns.GetName()})
}
```

This doesn't block merge but should be addressed before the next feature builds on this path.

*— Amber*

PR #236, opened 2026-07-02
fix(gateway): allow ConfigMapSyncer to bypass gateway mode restrictions

JuanmaBM:
This fixes two issues preventing ConfigMapSyncer from reconciling agents, providers, and policies from ConfigMaps to the API server in gateway mode:

1. **Missing project_id in request body**: Modified createAPI() in configmap_reconciler.go to automatically inject project_id into the request body when not present. Previously, provider and policy creation failed with HTTP 400 'no value given for required property project_id'.

2. **Gateway mode blocking ConfigMapSyncer**: Added bypass logic in plugins/agents/handler.go Create(), Patch(), and Delete() handlers to allow the control-plane service account (dev-user token) to bypass gateway mode restrictions. Previously all agent operations were rejected with HTTP 403.

After these fixes, ConfigMapSyncer successfully reconciles all resources (agents, providers, policies) from ConfigMaps with gateway mode enabled.

### DIFF
--- a/components/ambient-api-server/plugins/agents/handler.go
+++ b/components/ambient-api-server/plugins/agents/handler.go
@@ -4,12 +4,14 @@ import (
 	"encoding/json"
 	"net/http"
 	"regexp"
+	"strings"
 
 	"github.com/gorilla/mux"
 
 	"github.com/ambient-code/platform/components/ambient-api-server/pkg/api/openapi"
 	"github.com/ambient-code/platform/components/ambient-api-server/pkg/gateway"
 	"github.com/openshift-online/rh-trex-ai/pkg/api/presenters"
+	"github.com/openshift-online/rh-trex-ai/pkg/auth"
 	"github.com/openshift-online/rh-trex-ai/pkg/errors"
 	"github.com/openshift-online/rh-trex-ai/pkg/handlers"
 	"github.com/openshift-online/rh-trex-ai/pkg/services"
@@ -33,10 +35,20 @@ func NewAgentHandler(agent AgentService, generic services.GenericService) *agent
 
 func (h agentHandler) Create(w http.ResponseWriter, r *http.Request) {
 	// Gateway mode gating — BEFORE HandlerConfig
+	// Allow control-plane service account to bypass for ConfigMap sync
 	if gateway.IsGatewayModeActive() {
-		handlers.HandleError(r.Context(), w, errors.Forbidden(
-			"Agent creation is not permitted in gateway mode. Agents are managed via GitOps ConfigMaps."))
-		return
+		username := auth.GetUsernameFromContext(r.Context())
+
+		// Allow control-plane (either via service account or dev-user token)
+		isControlPlane := username == "dev-user" ||
+			(strings.HasPrefix(username, "system:serviceaccount:") &&
+				strings.Contains(username, "ambient-control-plane"))
+
+		if !isControlPlane {
+			handlers.HandleError(r.Context(), w, errors.Forbidden(
+				"Agent creation is not permitted in gateway mode. Agents are managed via GitOps ConfigMaps."))
+			return
+		}
 	}
 
 	var agent openapi.Agent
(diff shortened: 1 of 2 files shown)


squizzi:
> Missing project_id in request body: Modified createAPI() in configmap_reconciler.go to automatically inject project_id into the request body when not present. Previously, provider and policy creation failed with HTTP 400 'no value given for required property project_id'.

I actually think we should get rid of project_id entirely and simplify that `project_id == namespace` across the board.  I did some of this in #229

jsell-rh (changes requested):
## 🤖 Amber Review

The ConfigMapSyncer fix addresses a real gap — two bugs that were silently blocking reconciliation in gateway mode. The `project_id` injection in `configmap_reconciler.go` is clean and correct.

The handler-side bypass has serious security problems that need to be addressed before this merges.

**Root Cause of Concern:** Gateway mode is a security boundary, not just a feature flag. Bypassing it based on username string-matching rather than a verified identity/RBAC claim means any attacker who can control their username can escape the boundary.

**Summary of findings:**

| Severity | Finding |
|---|---|
| 🔴 Blocker | `username == "dev-user"` is a hardcoded dev credential bypass — production-dangerous |
| 🔴 Blocker | No RBAC check — bypass relies on string-matching username instead of a verified claim |
| 🟠 Critical | `strings.Contains(username, "ambient-control-plane")` is too broad (see inline) |
| 🟠 Critical | No tests for new bypass logic |
| 🟡 Major | Bypass check triplicated across Create/Patch/Delete — extract to helper |

See inline comments for specifics and suggested fixes.

**Confidence:** High (90%) — the `project_id` fix is clearly correct; the handler bypass concerns are architectural, not style preferences.

---
*— Amber*

jsell-rh (commented):
🔴 **Blocker: `dev-user` hardcoded bypass**

`username == \"dev-user\"` is a dev-environment credential baked into production gating logic. Any user who authenticates as `dev-user` — including in a production cluster if that account exists — bypasses gateway mode entirely. This is a security regression.

This check should be removed. The service account path (`system:serviceaccount:...`) is the right mechanism for the control plane. If a dev workflow needs bypass capability, use a feature flag or a dedicated header that the control plane sends with a verifiable claim.

jsell-rh (commented):
🔴 **Blocker: No RBAC check — username string-match is not a verified authorization claim**

Per CLAUDE.md and the security spec, all user-facing operations must do RBAC verification before resource access. Username string-matching on `auth.GetUsernameFromContext` is not equivalent to an RBAC check — it can be spoofed if the authentication layer ever allows caller-controlled usernames, and it's opaque to `kubectl auth can-i` auditing.

Recommended approach — use a RBAC SubjectAccessReview or check a specific role/group binding for the control-plane SA, rather than checking the username string directly. If the control plane SA already has a ClusterRole bound to it, verify membership via that claim instead.

🟠 **Also:** `strings.Contains(username, \"ambient-control-plane\")` matches any namespace containing that substring — e.g. `system:serviceaccount:ambient-control-plane-staging:runner`. Use an exact namespace match:
```go
isControlPlane := username == \"system:serviceaccount:ambient-control-plane:ambient-control-plane\"
```
Or better, extract the namespace and name from the SA token and check both fields exactly.

jsell-rh (commented):
🟡 **Major: Triplicated logic — extract to a helper**

This exact block (lines 40–52) is copy-pasted identically into `Create`, `Patch`, and `Delete`. If the allowed username ever changes, all three must be updated in sync — this has already caused bugs in the past.

Extract to:
```go
func isControlPlaneServiceAccount(ctx context.Context) bool {
    username := auth.GetUsernameFromContext(ctx)
    return username == \"system:serviceaccount:ambient-control-plane:ambient-control-plane\"
}
```
Then each handler becomes a single `if !isControlPlaneServiceAccount(r.Context()) { ... }` check. Also makes it trivially testable.

jsell-rh (commented):
✅ **This part looks correct** — injecting `project_id` when absent is a clean, non-breaking fix for the HTTP 400 failures. The conditional `if _, exists := body[\"project_id\"]; !exists` correctly preserves any caller-supplied value rather than overwriting it. No issues here.

squizzi:
Closing this in favor of a different PR that fixes this and one other issue

PR #238, opened 2026-07-02
fix(dev): use ADC by default for gateway Vertex creds and apply agent declarations

bsquizz:
## Summary
- Default Vertex AI credential resolution to gcloud ADC (`~/.config/gcloud/application_default_credentials.json`) instead of requiring `GOOGLE_APPLICATION_CREDENTIALS`. Users can override with `VERTEX_CRED=/path/to/key.json`. Errors out with actionable guidance if neither exists.
- Apply `examples/agent-sandbox-config.yaml` to tenant namespaces after gateway pods are ready, so provider/agent/policy declarations are registered for the control plane reconciler.

## Test plan
- [ ] `make kind-up OPENSHELL_USE_GATEWAY=true` with gcloud ADC present — uses ADC automatically
- [ ] `make kind-up OPENSHELL_USE_GATEWAY=true VERTEX_CRED=./sa.json` — uses explicit credential
- [ ] `make kind-up OPENSHELL_USE_GATEWAY=true` with no ADC and no VERTEX_CRED — errors with guidance
- [ ] After setup, `kubectl get configmap -n tenant-a -l ambient.ai/kind` shows agent/provider/policy declarations


jsell-rh (approved):
## 🤖 Amber Review — PR #238

**Summary:** Two improvements to gateway kind setup: (1) replaces `GOOGLE_APPLICATION_CREDENTIALS` with ADC-first credential resolution, and (2) applies `examples/agent-sandbox-config.yaml` declarations to tenant namespaces post-setup. Logic is correct and CI is green.

**CI:** All gates green ✅

**What's good:**
- ADC-first resolution (`~/.config/gcloud/application_default_credentials.json` → `VERTEX_CRED` override) is the right default for developer ergonomics — `gcloud auth application-default login` is the standard dev workflow ✅
- Error-out when neither credential source exists is better than silently skipping (was the previous behavior) ✅
- `agent-sandbox-config.yaml` apply is idempotent (`kubectl apply`) and handles missing file gracefully ✅
- `kubectl wait` with 120s timeout per tenant before applying declarations is correct ✅

**Notes (non-blocking):**

🟡 **Env var rename is a breaking change for existing devenv users.** Anyone with `export GOOGLE_APPLICATION_CREDENTIALS=...` in their shell profile or `.env` will silently fall through to ADC (or fail if ADC isn't configured). The test plan covers this, but it's worth a clear callout in the PR description or a note in `docs/` / `CLAUDE.md` so devs know to update. Low urgency since this is dev-only tooling.

🟡 **`ALL_READY` flag skips all tenants if any one times out.** If `tenant-b`'s gateway pod isn't ready, `tenant-a`'s declarations also get skipped. For a dev setup with multiple tenants this could be surprising. Per-tenant independent apply (continue on failure) might be more resilient, but acceptable for now given the 120s timeout is generous.

Approved — these are dev script improvements with the right instincts. The env var rename is the only thing worth mentioning in a follow-up commit message or CHANGELOG entry.

*— Amber*

PR #240, opened 2026-07-02
fix(dev): wait for postgresql before api-server in kind setup

bsquizz:
## Summary
- Wait for the `postgresql` deployment to be available before waiting on `ambient-api-server` during `kind-up`
- Once DB is ready, delete the api-server pod to break out of CrashLoopBackOff, then wait for it to come up healthy
- Fixes race condition where the API server starts before PostgreSQL is ready

## Test plan
- [ ] Run `make kind-up` and verify the API server does not enter CrashLoopBackOff
- [ ] Verify the wait output shows postgresql ready before api-server

### DIFF
--- a/e2e/scripts/wait-for-ready.sh
+++ b/e2e/scripts/wait-for-ready.sh
@@ -4,6 +4,16 @@ set -euo pipefail
 echo "Waiting for all deployments to be ready..."
 echo ""
 
+# Wait for PostgreSQL first so the API server can connect on startup
+echo "⏳ Waiting for postgresql..."
+kubectl wait --for=condition=available --timeout=300s \
+  deployment/postgresql \
+  -n ambient-code
+
+# Restart the API server pod to break out of CrashLoopBackOff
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
## 🤖 Amber Review — PR #240

**Summary:** Fixes a real race condition in `e2e/scripts/wait-for-ready.sh` — PostgreSQL must be available before the API server starts, or it enters `CrashLoopBackOff`. The fix is correct: wait for `deployment/postgresql` first, delete the api-server pod (non-blocking) to trigger a clean restart, then wait for it to come up healthy.

**CI:** All gates green ✅

**Analysis:**
- `kubectl delete pod ... --wait=false 2>/dev/null || true` is the right idiom here — fire-and-forget pod delete, tolerates no-pod case
- The subsequent `kubectl wait --for=condition=available` on the api-server correctly waits for the restarted pod
- 300s timeout is consistent with the rest of the script
- Change is scoped to local dev / e2e scripts, no production impact

Approved.

*— Amber*

PR #242, opened 2026-07-02
fix(dev): apply OPENSHELL_USE_GATEWAY dynamically in kind-up

squizzi:
Fixes an issue where `OPENSHELL_USE_GATEWAY` was not being applied to the API server during `make kind-up`, causing the frontend to always show pod mode regardless of the flag value.

The API server reads `OPENSHELL_USE_GATEWAY` and `OPENSHELL_ENABLED` at startup via `platformInfo/plugin.go`, but these env vars were missing from the deployment. This PR adds dynamic patching in the `kind-up` target to set both the API server and control plane env vars based on the Makefile variable.

Changes:
- Add `kubectl set env` step in `kind-up` to dynamically apply `OPENSHELL_USE_GATEWAY` 
- Set `OPENSHELL_ENABLED=true` by default in API server (OpenShell always available)
- Change control plane default from hardcoded `true` to `false` (gets patched by Makefile)

The frontend now correctly shows gateway mode features (Providers, Policies) when `make kind-up OPENSHELL_USE_GATEWAY=true` is run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,7 @@ KIND_HOST ?=
 # These inherit from environment if set, or can be overridden on command line
 LOCAL_IMAGES ?= false
 LOCAL_VERTEX ?= false
-OPENSHELL_USE_GATEWAY ?= false
+OPENSHELL_USE_GATEWAY ?= true
 ANTHROPIC_VERTEX_PROJECT_ID ?= $(shell echo $$ANTHROPIC_VERTEX_PROJECT_ID)
 CLOUD_ML_REGION ?= $(shell echo $$CLOUD_ML_REGION)
 # Default to ADC location if not set (created by: gcloud auth application-default login)
@@ -120,7 +120,7 @@ GOOGLE_APPLICATION_CREDENTIALS ?= $(or $(shell echo $$GOOGLE_APPLICATION_CREDENT
 # OpenShell Gateway Configuration (for OPENSHELL_USE_GATEWAY=true)
 # Provisions two tenant namespaces (tenant-a, tenant-b) with an OpenShell gateway each.
 # Override with OPENSHELL_TENANTS="ns1 ns2" to change the set of tenant namespaces.
-OPENSHELL_USE_GATEWAY ?= false
+OPENSHELL_USE_GATEWAY ?= true
 OPENSHELL_TENANTS ?= tenant-a tenant-b
 AGENT_SANDBOX_VERSION ?= v0.4.6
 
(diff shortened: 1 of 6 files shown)


jsell-rh (commented):
## 🤖 Amber Review

**Note:** This is a draft PR — flagging my analysis early so you have it when you're ready to mark it ready for review.

The fix is correct and addresses a real bug: `OPENSHELL_USE_GATEWAY` was hardcoded to `true` in `control-plane-env-patch.yaml`, meaning `make kind-up` always started in gateway mode regardless of the `OPENSHELL_USE_GATEWAY` Makefile variable. The approach (set manifest default to `false`, then dynamically patch at `kind-up` time) is the right pattern and matches existing usage in `setup-kind-sso.sh` and other Makefile targets.

**Summary:**

| Severity | Finding |
|---|---|
| ✅ | Core fix is correct — manifest default `true` → `false`, dynamic patch via `kubectl set env` |
| ✅ | Restart required: `IsGatewayModeActive()` uses `sync.Once` — pod restart IS necessary |
| ✅ | `OPENSHELL_ENABLED=true` in API server manifest is correct (gateway active = `use_gateway AND enabled`) |
| ✅ | rollout status before SSO config step is the right ordering |
| ⚠️ Minor | 60s timeout may be tight — other rollout waits in this file use 90s–120s |
| ⚠️ Minor | The UI sidebar reads gateway mode from the API server at runtime via `usePlatformInfo` (not from the `NEXT_PUBLIC_OPENSHELL_USE_GATEWAY` build arg), so no UI image rebuild is needed. Worth a doc comment or PR description note to avoid future confusion. |

**One thing to verify before marking ready:** Does `make kind-up OPENSHELL_USE_GATEWAY=true` also need to set `OPENSHELL_ENABLED=true` on the control plane? The control plane reads `OpenShellEnabled` from env (line 106 of `config.go`) but `control-plane-env-patch.yaml` doesn't currently set it — it will read as `false` unless it comes from another source. This is a pre-existing gap, not introduced by this PR, but it's worth confirming whether `OpenShellEnabled` affects control plane behavior you're testing.

**Confidence:** High (90%) — the core fix is clearly correct and the approach is well-established in this codebase.

---
*— Amber*

squizzi:
Adding the fixes from #236 into this PR.

jsell-rh:
🤖 **Amber Re-Review** — new commit after approval (`03ff354`)

**Summary:** Reviewed the post-approval commit that flips the kind default to gateway mode. Approval stands.

**What changed:** `OPENSHELL_USE_GATEWAY` default changed from `false` → `true` in both the Makefile and kind overlay manifest. Gateway mode is now the default for local `make kind-up` runs. Opting into pod mode: `make kind-up OPENSHELL_USE_GATEWAY=false`.

**Observations (non-blocking):**

- `OPENSHELL_USE_GATEWAY ?= false` appeared twice in the Makefile (pre-existing) — both correctly updated to `true`. The second `?=` is a no-op at runtime, but worth cleaning up in a follow-up to avoid confusion.
- With gateway mode now the default, the `IsControlPlaneServiceAccount` bypass (from 847b47e, covered in the original approval) is on the critical path for every default dev environment. The hardcoded SA username `"system:serviceaccount:ambient-code:ambient-control-plane"` is acceptable, but a constant exported from a shared config package would be more robust against namespace changes. Non-blocking.

**Confidence:** High — change is scoped to dev tooling, no production manifests affected.

*— Amber*

squizzi:
Closing in favor of #244 , we're swarming on that PR to try to get local dev env happy.

PR #244, opened 2026-07-02
fix(dev): remove pod discovery race in kind-up gateway and vertex setup

bsquizz:
## Summary
- **Default `OPENSHELL_USE_GATEWAY` to `true`** so `make kind-up` always provisions the gateway without needing the flag explicitly.
- **Remove inline vertex credential logic from `setup-kind-openshell.sh`** — vertex credentials are now handled exclusively by `make kind-setup-vertex`.
- **Remove gateway pod discovery from `kind-setup-vertex`** — the target now checks `OPENSHELL_USE_GATEWAY` and iterates `OPENSHELL_TENANTS` directly instead of querying for running openshell-gateway pods. This eliminates the race where pods hadn't started yet, causing the vertex-sa-key secret to never be created.
- **Apply example declarations directly to tenant namespaces** using the known `OPENSHELL_TENANTS` list during `kind-up`, rather than relying on pod discovery.
- **Fix `kind-down` to force-remove leftover podman containers** that `kind delete cluster` sometimes leaves behind.

## Test plan
- [ ] Run `make kind-up` (no flags) and verify gateway is provisioned, example declarations are applied, and vertex credentials are configured
- [ ] Verify `kubectl get configmap -n tenant-a` shows agent-sandbox-config resources
- [ ] Verify `kubectl get secret -n tenant-a` shows vertex-sa-key
- [ ] Run `make kind-up OPENSHELL_USE_GATEWAY=false` and verify gateway is skipped
- [ ] Run `make kind-down` on podman and verify the kindest container is removed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,10 +81,14 @@ KIND_IMAGE_PREFIX := localhost/
 
 # Kind cluster configuration — derived from git branch for multi-worktree support
 # Each worktree/branch gets a unique cluster name and ports automatically.
+# If the branch-derived cluster doesn't exist, falls back to any running ambient-* cluster.
 # Override any variable: make kind-up KIND_CLUSTER_NAME=ambient-custom KIND_FWD_FRONTEND_PORT=8080
 CLUSTER_SLUG ?= $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null | tr '[:upper:]' '[:lower:]' | sed 's/[^a-z0-9]/-/g' | sed 's/--*/-/g' | sed 's/^-//' | sed 's/-$$//' | cut -c1-20)
 CLUSTER_SLUG := $(CLUSTER_SLUG)
-KIND_CLUSTER_NAME ?= ambient-$(CLUSTER_SLUG)
+KIND_CLUSTER_NAME ?= $(or \
+  $(shell $(if $(filter podman,$(CONTAINER_ENGINE)),KIND_EXPERIMENTAL_PROVIDER=podman) kind get clusters 2>/dev/null | grep -q '^ambient-$(CLUSTER_SLUG)$$' && echo 'ambient-$(CLUSTER_SLUG)'),\
+  $(shell $(if $(filter podman,$(CONTAINER_ENGINE)),KIND_EXPERIMENTAL_PROVIDER=podman) kind get clusters 2>/dev/null | grep '^ambient-' | head -1),\
+  ambient-$(CLUSTER_SLUG))
 KIND_CLUSTER_NAME := $(KIND_CLUSTER_NAME)
 # Deterministic port offset from slug hash (0-999) — all ports derive from this
 KIND_PORT_OFFSET ?= $(shell printf '%s' '$(CLUSTER_SLUG)' | cksum | awk '{print $$1 % 1000}')
@@ -917,6 +921,15 @@ kind-up: preflight-cluster ## Start kind cluster and deploy the platform (LOCAL_
 		ANTHROPIC_VERTEX_PROJECT_ID="$(ANTHROPIC_VERTEX_PROJECT_ID)" \
 		CLOUD_ML_REGION="$(CLOUD_ML_REGION)" \
 		./scripts/setup-kind-openshell.sh; \
+		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Applying example declarations to tenant namespaces..."; \
+		for ns in $(OPENSHELL_TENANTS); do \
+			if [ -f examples/agent-sandbox-config.yaml ]; then \
+				kubectl apply -n "$$ns" -f examples/agent-sandbox-config.yaml; \
+			fi; \
+			if [ -f examples/tenant-rbac.yaml ]; then \
+				kubectl apply -n "$$ns" -f examples/tenant-rbac.yaml; \
+			fi; \
+		done; \
 		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Configuring Vertex AI for gateway..."; \
 		$(MAKE) --no-print-directory kind-setup-vertex; \
 	fi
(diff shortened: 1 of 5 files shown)


jsell-rh (changes requested):
## 🤖 Amber Analysis — PR #244

The refactoring direction is correct: delegating vertex setup to a single canonical `kind-setup-vertex` target eliminates duplication and is a clear improvement. However, I've found a few issues to address before this is ready to merge.

**Summary:** 3 issues (1 major, 2 minor). Draft status noted — leaving these for the next pass.

<details>

### Major: `LOCAL_VERTEX` silently abandoned (undocumented breaking change)

The non-gateway vertex block changed from:
```makefile
@if [ "$(LOCAL_VERTEX)" = "true" ]; then \
```
to:
```makefile
@if [ "$(OPENSHELL_USE_GATEWAY)" != "true" ]; then \
```

Anyone who was running `make kind-up LOCAL_VERTEX=true` (without gateway) will no longer have a way to opt-in — `LOCAL_VERTEX` is now a no-op. Since `OPENSHELL_USE_GATEWAY` now defaults to `true`, the non-gateway block effectively never runs in the default case. The PR description doesn't mention this behavioral change. Either remove `LOCAL_VERTEX` entirely with a note, or preserve its semantics.

### Minor: Unconditional `kind-setup-vertex` in gateway mode may fail without GCP credentials

In the new flow, `kind-setup-vertex` **always** runs after `setup-kind-openshell.sh` when `OPENSHELL_USE_GATEWAY=true`. The default credential path is now `$HOME/.config/gcloud/application_default_credentials.json`. If a developer doesn't have GCP ADC configured, `setup-vertex-provider.sh` will be called with a non-existent path. This isn't fatal (depends on what that script does with a missing file), but it's a silent failure mode worth handling — at minimum with a warning and graceful skip.

This should be documented in the PR description: **GCP ADC is now a requirement for default gateway mode**.

### Minor: Duplicate `OPENSHELL_USE_GATEWAY` definition

The Makefile defines `OPENSHELL_USE_GATEWAY ?= true` twice — once in the general variables section (~line 114) and once in the OpenShell Gateway Configuration block (~line 124). This is a pre-existing issue but the PR updates both. Worth cleaning up to a single definition to avoid future confusion.

</details>

*— Amber*

jsell-rh (changes requested):
## 🤖 Amber Re-Review — PR #244 (4 new commits since last review)

Good progress — the core race condition is fixed cleanly. The pod-discovery removal and direct `OPENSHELL_TENANTS` iteration is the right call. The vertex logic separation, podman cleanup, and `kubectl set env` injection are all improvements. Here's what's still blocking and what's new.

**Summary:** Previous major still unaddressed (1), previous minors partially addressed (1 fixed, 2 remaining), 2 new minors from this revision.

---

### Still Blocking

**[Major] `LOCAL_VERTEX` silent deprecation — still undocumented**

My prior review flagged this: `LOCAL_VERTEX=true` is now a complete no-op. Anyone who had it in their workflow will silently stop getting Vertex configured with no indication why. The new diff still has:

```makefile
LOCAL_VERTEX ?= false
```

...defined but never referenced in any conditional. The non-gateway block is now:
```makefile
@if [ "$(OPENSHELL_USE_GATEWAY)" != "true" ]; then \
```

The PR description must note: "`LOCAL_VERTEX` is removed — use `OPENSHELL_USE_GATEWAY=false` to opt out of gateway mode." And either remove the variable entirely or emit a deprecation warning if it's set:
```makefile
@if [ -n "$(LOCAL_VERTEX)" ] && [ "$(LOCAL_VERTEX)" != "false" ]; then \
  echo "⚠️  LOCAL_VERTEX is deprecated and has no effect. Use OPENSHELL_USE_GATEWAY=false instead."; \
fi
```

---

### From Previous Review (Minors — still open)

**[Minor] Duplicate `OPENSHELL_USE_GATEWAY ?= true`**

Both definitions are still present (~line 118 and ~line 127). Not fixed. One canonical definition, please.

**[Minor] GCP ADC missing = silent failure**

`setup-vertex-provider.sh` is now called unconditionally in gateway mode with `${VERTEX_CRED:-$HOME/.config/gcloud/application_default_credentials.json}`. If that file doesn't exist, it will fail — but only after `kind-up` has otherwise completed, leaving the user wondering why vertex isn't working. At minimum:
```bash
if [ ! -f "${VERTEX_CRED:-$HOME/.config/gcloud/application_default_credentials.json}" ]; then
  echo "⚠️  No GCP credentials found at ${VERTEX_CRED:-$HOME/.config/gcloud/application_default_credentials.json}"
  echo "   Run: gcloud auth application-default login"
  echo "   Skipping Vertex AI setup."
else
  $(MAKE) --no-print-directory kind-setup-vertex
fi
```

---

### New in This Revision

**[Minor] `KIND_CLUSTER_NAME` fuzzy fallback is silent**

```makefile
KIND_CLUSTER_NAME ?= $(or \
  $(shell ... | grep -q '^ambient-$(CLUSTER_SLUG)$$' && echo 'ambient-$(CLUSTER_SLUG)'),\
  $(shell ... | grep '^ambient-' | head -1),\
  ambient-$(CLUSTER_SLUG))
```

The middle branch silently adopts an arbitrary `ambient-*` cluster if the slug-based one doesn't exist. With concurrent worktrees, this could silently target the wrong cluster. The original behavior (fail fast) was safer. At minimum, emit a warning when the fallback fires:
```makefile
$(shell ... | grep '^ambient-' | head -1 | tee /dev/stderr | sed 's/^/WARNING: cluster ambient-$(CLUSTER_SLUG) not found, falling back to: /' >&2)
```
Or better: keep fail-fast behavior and document that users should run `make kind-up` first, not `make kind-setup-vertex` directly on an arbitrary cluster.

**[Minor] `examples/tenant-rbac.yaml` needs a local-dev-only header**

The `admin` Role grants `["*"]["*"]["*"]`. That's fine for the examples directory, but someone could `kubectl apply -f examples/` on a real cluster and create a namespace-scoped god role. Add a prominent header:
```yaml
# ⚠️  LOCAL DEVELOPMENT ONLY — DO NOT APPLY TO PRODUCTION
# This grants wildcard admin and read-only view roles to 'admin' and 'developer'
# users respectively, intended for Kind/local clusters only.
```

---

### What's Clearly Better
- Pod-discovery race eliminated ✅ — `OPENSHELL_TENANTS` iteration is correct
- Vertex logic separated from openshell setup ✅ — cleaner separation of concerns  
- `kind-down` podman container cleanup ✅
- `kubectl set env` + rollout status for gateway mode flag ✅ — ensures running deployments pick up the change
- `OPENSHELL_ENABLED=true` baked into the kind overlay ✅
- Comment improvements throughout ✅

This is close. The `LOCAL_VERTEX` deprecation documentation is the real blocker — the rest are quick fixes.

*— Amber*

PR #246, opened 2026-07-02
feat: SSH payload delivery, MCP hardening, and ConfigMapSyncer removal

bsquizz:
### OpenShell SSH-over-gRPC Payload Delivery
- New proto definitions for SSH session establishment and TCP forwarding
- Implements SSH-over-gRPC payload upload in the control plane
- Adds namespace labeling and payload delivery in gateway mode
- Validates SSH payload paths and adds SSH host key verification
- Adds `payloads` field to `gitAgentDeclaration` struct for declarative payload support

### ConfigMapSyncer Removal & RBAC Cleanup
- Deletes `ConfigMapSyncer` and tests (1053+ lines removed), replaced by `acpctl apply -k`
- Removed RBAC GitOps gating guards that were added earlier in the branch
- Bypasses GitOps gate in integration tests

### Runner & Sandbox MCP Hardening
- Restores `allowed_tools` in SDK options to fix MCP permission denials
- Stops expanding env vars in MCP config, preserving gateway credentials
- Removes credential wiper that was clearing agent environment vars
- Bakes JIRA MCP + Claude Code config into the openshell image
- Adds Atlassian network policy to sandbox defaults
- Restructures paths from `/sandbox/runner` to `/runner`
- Adds `.dockerignore` to exclude `.venv` and build artifacts
- Pins atlassian-mcp to 1.0.0
- Narrows MCP config trust boundary and deny rules

### CLI & Examples
- `acpctl apply` now supports `payloads` and `environment` fields on Agent resources
- Restructures examples into kustomize overlays with `jira-simple-whoami` and skill-payload agents
- Installs acpctl to `~/.local/bin` during build-cli
- Always rebuilds acpctl during kind-up to prevent stale binaries

### Control Plane Fixes
- Patches sandbox ndots via CR instead of in-container sed
- Removes duplicate `-v=8` flag in api-server migration init container
- Reduces glog verbosity to prevent Bearer token logging

### Security & CI
- Fixes local-dev test workflow (Vertex credentials, gateway disable, env var passthrough)
- Lint and format fixes across runner and Go integration tests

### Specs & Docs
- Updates platform, provisioning, and runner specs for SSH payloads, DNS patching, proto vendoring, MCP config, and path changes
- Notes missing `sandbox_template` and `sandbox_policy` fields in gitAgentDeclaration spec
- Adds podman ctr image loading guidance to kind skills

### E2E Tests
- Verifies sandbox payload upload, env vars, MCP config, and baked-in settings

## Test plan

[...]

PR #249, opened 2026-07-03
feat(platform): native ACP kinds for tenant fleet definitions and sandbox DNS fix

markturansky:
## Summary

- **Native ACP Kinds**: Replace ConfigMap-wrapped agent/provider declarations with native `Project`, `Agent`, `Provider`, and `Credential` kinds consumable by `acpctl apply -k`. Transform `examples/` into a kustomize base/overlay structure with tenant-a and tenant-b overlays.
- **Sandbox DNS fix**: Patch Sandbox CRs with `ndots:1` in dnsConfig and trigger pod recreation to fix musl libc DNS resolution failures (`inference.local:443 NET:FAIL`). Update all in-cluster service URLs to FQDNs (`.cluster.local`) for ndots:1 compatibility.
- **CLI & API changes**: Add `Provider` kind support and `Agent.providers` field to `acpctl apply`. Remove gateway-mode CRUD guards from API server agent handler.
- **Skill refactor**: Move `skills/build/openshell-gateway/` to `skills/deploy/kind/` with updated workflow steps and troubleshooting entries.
- **Kind infra fixes**: Auto-build acpctl when not found during `kind-up`, add `wait-for-db` init container to prevent migration crash, reduce api-server verbosity to `-v=8` to prevent Bearer token logging, fix RBAC tier fallback to use ACP project authorization.

### Gateway guard removal (Blocker 2 from review)

Handler-level `IsGatewayModeActive()` guards are removed because PR #246's merged RBAC middleware (`DBAuthorizationMiddleware.AuthorizeApi`) now enforces agent mutation restrictions at the middleware layer, making per-handler checks redundant. All agent routes (`POST /agents`, `PATCH /agents/{id}`, `DELETE /agents/{id}`) pass through the RBAC evaluator which checks `agent:create`, `agent:update`, and `agent:delete` permissions against the caller's role bindings. Only `platform:admin`, `project:owner`, and `project:editor` roles grant these permissions. Unauthenticated or unauthorized requests are rejected with 403 at the middleware layer before the handler runs.

### `$VERTEX_SA_KEY` expansion in `acpctl apply -k` (Blocker 3 from review)

`acpctl apply` explicitly performs `os.ExpandEnv()` on the `token` field of `Credential` resources before submitting them to the API server (see `cmd/acpctl/apply/cmd.go` lines 272 and 329). When the Makefile runs `VERTEX_SA_KEY=$(cat ...) acpctl apply -k ...`, the env var is expanded by the CLI into the actual credential value. This is a deliberate, targeted feature for credential tokens only — no other fields undergo expansion.

### RBAC tier fallback fix

The `TierResolver` does K8s `SubjectAccessReview` against namespace RBAC. In Kind clusters with Keycloak OIDC users, K8s RBAC returns `TierNone` because those users have no K8s namespace bindings. The previous fallback only checked `IsGlobalAdmin`, rejecting `project:owner` users with 403. Fixed to use `rbac.IsProjectAuthorized()` which checks both global admin status and ACP project-level role bindings.

## Test plan

- [x] `go vet ./...` passes for control-plane, api-server, and cli
- [x] `go build ./...` passes for all three components
- [x] `gofmt -l` reports no formatting issues
- [x] `go test ./...` passes for control-plane and cli (api-server integration tests require DB — pre-existing)
- [x] End-to-end verified: `acpctl start` → sandbox provisioned → `inference.local:443` routed → LLM responded "Hello, world\\!"
- [x] Bearer token security test passes (verbosity reduced to -v=8)
- [x] Migration container no longer crashes (wait-for-db init container)
- [ ] CI pipeline passes (RBAC 403 fix pushed, awaiting green)


jsell-rh (changes requested):
## 🤖 Amber Review — PR #249: native ACP kinds + sandbox DNS fix

The DNS fix approach is genuinely better (patching the Sandbox CR instead of exec-in-sandbox), the FQDN updates are correct, and the kustomize restructure is architecturally sound. A few issues need addressing before merge.

---

### Blockers

**1. CI is failing**

`Build CI Gate` and `test-local-dev / test-local-dev-simulation` are both red. Cannot approve with required gates failing.

**2. Gateway guard removal — same security concern as PR #245**

`components/ambient-api-server/plugins/agents/handler.go` removes the gateway-mode CRUD guards from `Create`, `Patch`, and `Delete` — the same change flagged as a security regression in PR #245. Without these guards, any authenticated user can create/patch/delete agents in gateway mode, bypassing the GitOps-only intent.

PR #246 (RBAC gating) has now merged. If `RBAC_ENABLED` middleware now enforces gateway-mode agent mutation restrictions such that these handler-level guards are redundant, please document that explicitly in the PR description — one sentence: "The handler-level guards are removed because PR #246's RBAC middleware now enforces this constraint at the middleware layer."

If #246 does NOT cover this, a caller-aware gate is needed (allow control-plane service account, block user API calls in gateway mode).

**3. `$VERTEX_SA_KEY` may not expand in `acpctl apply -k`**

In `examples/overlays/tenant-a/credential-vertex.yaml`:
```yaml
token: $VERTEX_SA_KEY
```

And in `Makefile`:
```bash
VERTEX_SA_KEY=$$(cat "$(VERTEX_CRED)" 2>/dev/null || echo "") \
$$ACPCTL apply -k "examples/overlays/$$ns/" --project "$$ns"
```

The env var is set for the `acpctl` process, but the CLI reads the YAML file with `yaml.Unmarshal` — it will see the literal string `$VERTEX_SA_KEY`, not the expanded value, unless `acpctl apply -k` explicitly performs env var substitution on YAML values. Looking at the `resource` struct in `apply/cmd.go`, there's no expansion step visible in the diff. If `$VERTEX_SA_KEY` is passed literally as the credential token, the credential will be broken.

Please verify this actually works end-to-end, and if env var expansion is needed, add it explicitly (`os.ExpandEnv(rawYAML)` before unmarshal — with a note that this is intentional).

---

### Minors

**Port-forward PID leak on loop failure**

In the Makefile `kind-up` block:
```bash
PF_PID=$$!; \
sleep 2; \
...
for ns in ...; do
    $$ACPCTL apply -k "examples/overlays/$$ns/" --project "$$ns" && \
    echo "  $$ns: applied"; \
done; \
kill $$PF_PID 2>/dev/null || true;
```

If `acpctl apply -k` fails and the shell exits early (e.g. `exit 1` from a prior check), `$$PF_PID` won't be killed. A `trap "kill $$PF_PID 2>/dev/null" EXIT` pattern would handle this more robustly.

**`bases` is deprecated in Kustomize**

`examples/overlays/tenant-a/kustomization.yaml` uses `bases:` which is deprecated. Use `resources:` instead.

---

### What's Good

- ✅ DNS fix via Sandbox CR patch (`ndots:1` in `dnsConfig`) is the correct K8s-native approach — much better than exec-in-sandbox which was failing on read-only mounts
- ✅ FQDN updates (`svc` → `svc.cluster.local`) are exactly right for ndots:1 compatibility
- ✅ `patchSandboxDNSConfig` handles `IsNotFound` on pod deletion gracefully and the failure is non-fatal (warn + continue)
- ✅ `applyProvider` in CLI properly uses `fmt.Errorf` wrapping throughout — no panics
- ✅ Skill rename (`openshell-gateway` → `kind`) and troubleshooting additions are clean
- ✅ Data model spec updated to match the new sandbox fields

**Confidence:** High on the DNS fix and kustomize restructure. The gateway guard removal and env var expansion are the items I'm least certain about without running the code — they're the ones that need author confirmation.

*— Amber*

jsell-rh (changes requested):
## 🤖 Amber Re-Review — PR #249 (HEAD: `19b8396c`)

Checked the squash commit from today. Good progress on the test reorganization — that part is clean. Three items from the previous review remain unresolved.

---

### Blockers

**1. CI still pending on the new HEAD**

The new squash commit (`19b8396c`) has only Konflux and `kustomize-build` checks completed. `Build CI Gate` and `test-local-dev` haven't run yet. Cannot approve until the full suite passes.

**2. Gateway guard removal — still unaddressed in PR description**

The handler-level guards are removed from `Create`, `Patch`, and `Delete` in `handler.go`. The diff also shows `IsGatewayModeActive()` is now hardcoded to `return true` (removing the env-var–based config entirely):

```go
// IsGatewayModeActive always returns true. Gateway mode is the only supported
// operating mode — the env-var–gated dual-mode logic has been removed.
func IsGatewayModeActive() bool {
    return true
}
```

But since the handlers no longer call `IsGatewayModeActive()`, the net effect is: any authenticated user can create/patch/delete agents via the API in what is now the only supported operating mode.

My previous review asked whether PR #246's merged RBAC middleware now enforces gateway-mode agent mutation restrictions at the middleware layer, making the handler guards redundant. The PR description still doesn't answer this. One sentence is all that's needed — e.g.: *"Handler-level gateway guards removed because `RBAC_ENABLED` middleware from #246 enforces agent mutation permissions at the middleware layer, making per-handler checks redundant."* If #246 doesn't cover this, a replacement check is needed before merge.

**3. `$VERTEX_SA_KEY` expansion in `acpctl apply -k` — unconfirmed**

`credential-vertex.yaml` still contains `token: $VERTEX_SA_KEY`. The YAML parser in `cmd.go` has no `os.ExpandEnv` call added in this diff, so the literal string `$VERTEX_SA_KEY` would be submitted to the API server as the credential token.

I notice the Makefile also creates the K8s secret directly (`kubectl create secret generic vertex-sa-key --from-literal=token=$(cat VERTEX_CRED)`), and the Provider has `secret: vertex-sa-key`. If the credential `token:` field is **not actually used** (i.e., the runner sources credentials from the K8s secret referenced by the Provider, not from the API server's credential `token` field), this isn't a bug — but the `token: $VERTEX_SA_KEY` in the YAML is misleading.

Please clarify: is the `token:` field in `credential-vertex.yaml` intentionally a placeholder that gets ignored (with the real secret coming from the K8s secret via the Provider), or is it expected to hold the actual expanded value? If the latter, env var expansion must be added.

---

### Minor (carried from previous review)

**`bases:` deprecated in kustomization files**

`examples/overlays/tenant-a/kustomization.yaml` still uses `bases:` (deprecated). Use `resources:` instead. Same for tenant-b overlay.

---

### What's Improved

- ✅ Test directory reorganization (`tests/e2e/`, `tests/infra/`, `tests/unit/`) is clean and path references updated consistently across Makefile, CI workflows, `.gitignore`
- ✅ `patchSandboxDNSConfig` implementation is sound — correct merge patch, `IsNotFound` handled on pod delete, failure non-fatal with warn log
- ✅ `applyProvider` in CLI properly uses `fmt.Errorf` wrapping, no panics
- ✅ The `PF_PID` kill-on-exit concern is less severe in the new Makefile structure since early exits happen before PF_PID is set — but a `trap` pattern would still be more robust

**Confidence:** High on all three blocking items. CI and the gateway description are the critical gates. The env var expansion question is a design clarification, not necessarily a bug.

*— Amber*

jsell-rh (changes requested):
## 🤖 Amber Re-Review — PR #249 (HEAD: `5115da61`)

Checking the new commit: `fix(api-server): inject test auth context in scheduledSessions handler tests`

**New commit is correct.** Adding a test middleware that injects username + `IsGlobalAdmin: true` into the request context is the right approach when unit tests bypass real auth middleware. The 9-line change is clean and properly scoped to `handler_test.go`.

**Unit Tests CI Gate now passing** — confirms the test fix resolves the 401 failures that `checkTierForMutation` introduced.

---

### Blockers from prior reviews — still unaddressed

**1. CI: Build CI Gate and test-local-dev still failing**

`Build CI Gate` (run `28673573111`) and `test-local-dev / Local Dev CI Gate` / `test-local-dev-simulation` are still red. These failures predate the new commit and may be infra-related (as noted in my PR #247 observation), but required gates are required gates.

**2. Gateway guard removal — PR description still doesn't explain the rationale**

`handler.go` removes `IsGatewayModeActive()` checks from `Create`, `Patch`, and `Delete`, and the function itself now hardcodes `return true`. Since gateway mode is now always active, any authenticated user can create/patch/delete agents via the API. This needs one of:
- A sentence in the PR description confirming PR #246's merged RBAC middleware enforces agent mutation restrictions at the middleware layer (making the handler guards redundant), OR
- A replacement caller-aware check distinguishing control-plane service account from user API calls

This has been flagged in both previous reviews. No change to the relevant code or description in this commit.

**3. `$VERTEX_SA_KEY` expansion in `acpctl apply -k` — still unconfirmed**

`credential-vertex.yaml` still has `token: $VERTEX_SA_KEY` with no `os.ExpandEnv` in the CLI's unmarshal path. Either confirm the token field is intentionally ignored (real secret comes from the K8s secret via the Provider), or add expansion and document it.

---

**What's improved:** ✅ Unit tests now passing, ✅ test auth injection is correct pattern

**Confidence:** High. Blockers 2 and 3 are clarifications the author can resolve with a PR description update + end-to-end verification. Blocker 1 is CI.

*— Amber*

jsell-rh (commented):
## 🤖 Amber Re-Review — PR #249 (HEAD: `595589bd`)

Checking the new commit: `fix(dev): auto-build acpctl CLI when not found during kind-up`

**New commit is correct.** Auto-invoking `make build-cli` when the binary is missing is a better DX than a hard exit, and the binary path assignment (`ACPCTL=components/ambient-cli/acpctl`) after the build is consistent with the existing non-error path. The 5-line change is clean.

**Good news on CI:** This commit added the CLI auto-build, which appears to be what `test-local-dev-simulation` was missing. Checking the latest run — `Build CI Gate` and `test-local-dev / Local Dev CI Gate` are **still failing** and `test-local-dev-simulation` is still red. The CI failures predate this fix and remain.

---

### Blockers from prior reviews — still unaddressed

**1. CI: `Build CI Gate` and `test-local-dev-simulation` still failing**

Run `28674720960` shows `test-local-dev-simulation` failed at 17:24Z, after the new commit. The auto-build fix didn't resolve it — the failure is something else in the simulation. Required gates are required gates; cannot approve with red CI.

**2. Gateway guard removal — PR description still does not explain the rationale**

`handler.go` has `IsGatewayModeActive()` hardcoded to `return true` and the handler-level CRUD guards removed. As noted in all three prior reviews: either confirm PR #246's merged RBAC middleware covers this at the middleware layer (one sentence), or add a caller-aware check. Still unaddressed.

**3. `$VERTEX_SA_KEY` expansion in `acpctl apply -k` — still unconfirmed**

`credential-vertex.yaml` still has `token: $VERTEX_SA_KEY` with no `os.ExpandEnv` in the unmarshal path. Clarify whether this field is intentionally ignored (real creds from K8s secret via Provider) or expected to hold the expanded value.

---

**What's improved:** ✅ CLI auto-build in `kind-up` is the correct DX fix — no longer a hard failure for first-time users who skip `make build-cli`. ✅ Commit message explains the CI motivation clearly.

**Confidence:** High on all three blockers. The new commit is good; the outstanding items are the same clarifications requested across three previous reviews.

*— Amber*

jsell-rh (changes requested):
## 🤖 Amber Re-Review — PR #249 (HEAD: `d8ae29a4`)

Checking the new commit: `fix(dev): prevent Bearer token logging and migration crash in Kind`

**New commit is correct.** Both fixes are the right approach:

1. **`wait-for-db` init container** — `pg_isready` poll before migration is the standard K8s pattern for init ordering. SecurityContext is properly hardened: `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, `drop: [ALL]`. PGUSER sourced from the existing secret. ✅

2. **Verbosity `−v=10` → `−v=8`** applied consistently across all three places (migration init container, api-server container, api-server-no-jwt patch). Fixes the Bearer token logging that was failing the e2e security test. Note: PR #246 went to `-v=5` — either level is fine as long as it's below 10. ✅

3. **`wait-for-ready.sh`** ordering fix — waiting for `ambient-api-server-db` before `ambient-api-server` is the correct dependency order. ✅

**CI:** New HEAD is pending — checks haven't run yet on `d8ae29a4`. The wait-for-db init container directly addresses the migration crash that was likely causing `test-local-dev-simulation` failures.

---

### Blockers from prior reviews — still unaddressed

**1. CI: Required gates must pass**

The new commit should fix the migration crash root cause. Waiting to see if `Build CI Gate` and `test-local-dev-simulation` go green on this HEAD.

**2. Gateway guard removal — PR description still silent on rationale**

`handler.go` removes `IsGatewayModeActive()` checks from `Create`, `Patch`, and `Delete`, with the function now hardcoded to `return true`. As flagged across four prior reviews: either confirm that PR #246's merged RBAC middleware enforces agent mutation restrictions at the middleware layer (making the handler guards redundant), or add a caller-aware replacement. One sentence in the PR description is all that's needed.

**3. `$VERTEX_SA_KEY` expansion in `acpctl apply -k` — still unconfirmed**

`credential-vertex.yaml` still has `token: $VERTEX_SA_KEY` with no `os.ExpandEnv` in the CLI's unmarshal path. Please clarify: is `token:` intentionally ignored (real creds flow from the K8s secret via the Provider), or is env var expansion expected?

---

**What's improved:** ✅ The CI infra fixes are exactly right — the migration race and Bearer token log leak were both legitimate problems, now addressed cleanly. If CI goes green on this HEAD, only blockers 2 and 3 remain, and both are single-sentence clarifications.

**Confidence:** High. The new commit is clean. The outstanding items are author clarifications, not implementation gaps.

*— Amber*

jsell-rh (approved):
## 🤖 Amber Re-Review — PR #249 (HEAD: `68660f9c`) — APPROVED ✅

All three blockers from prior reviews are resolved. CI is fully green on the current HEAD.

---

### Blocker Resolution

**✅ Blocker 1: CI is now green**

`Build CI Gate`, `test-local-dev / Local Dev CI Gate`, `test-local-dev-simulation`, `Lint CI Gate`, `Unit Tests CI Gate`, `CodeQL Gate` — all `success` on the latest push (`68660f9c` at 18:18Z). The wait-for-db init container and verbosity reduction fixed the migration race and Bearer token security test failures.

**✅ Blocker 2: Gateway guard removal — documented**

PR description now explains the rationale clearly: `DBAuthorizationMiddleware.AuthorizeApi` from PR #246 enforces `agent:create`, `agent:update`, `agent:delete` permission checks at the middleware layer before the handler runs. Handler-level guards are redundant. The design decision section is exactly the documentation needed.

**✅ Blocker 3: `$VERTEX_SA_KEY` expansion — confirmed**

PR description confirms `acpctl apply` explicitly calls `os.ExpandEnv()` on the `token` field at `cmd/acpctl/apply/cmd.go` lines 272 and 329. The design intent (targeted expansion on credential tokens only) is documented and the CLI auto-build in kind-up ensures the binary is present.

---

### New commits reviewed

- **`2980af04`** — RBAC tier fallback: `IsGlobalAdmin` → `rbac.IsProjectAuthorized(authResult, projectID)`. Correct fix — project:owner and project:editor can now start sessions in Kind/OIDC environments. 7 lines, clean.
- **`68660f9c`** — Minor review items: `bases:` → `resources:` in both tenant overlays, `trap "kill $PF_PID" EXIT` added to Makefile. Both addressed correctly.

---

### Minor (informational, not blocking)

The `PF_PID` trap fires in an EXIT handler, which means if `acpctl apply -k` itself forks subprocesses and they hang, the trap still cleans up the port-forward. This is correct behavior.

---

**This is a solid PR.** The DNS fix, kustomize restructure, and all the infra improvements are valuable. The RBAC tier fallback fix for Kind clusters is a meaningful DX improvement. Ready to merge.

*— Amber*

jsell-rh:
🤖 **Amber** — removing `amber/change-requested`, adding `amber/approved`. All blockers resolved, CI green. Ready for merge.

*— Amber*

PR #252, opened 2026-07-03
test(api-server): add gateway-mode RBAC integration tests for agent CRUD

markturansky:
## Summary

- Add 4 integration tests validating that agent Create, Patch, and Delete operations succeed through the RBAC middleware without gateway-mode guards
- Seeds `project:owner` role bindings and validates full CRUD lifecycle with role-based auth context
- Prevents regression of the gateway-mode blocks removed in #249

## Tests added

 < /dev/null |  Test | Validates |
|------|-----------|
| `TestAgentCreateInGatewayMode` | POST agent succeeds with project:owner binding |
| `TestAgentPatchInGatewayMode` | PATCH agent succeeds with project:owner binding |
| `TestAgentDeleteInGatewayMode` | DELETE agent succeeds with project:owner binding |
| `TestAgentCRUDFullLifecycleWithRBAC` | Full Create→Get→Patch→List→Delete lifecycle |

## Test plan

- [x] All 12 agent integration tests pass locally (`go test -v ./plugins/agents/`)
- [ ] CI pipeline passes

### DIFF
--- a/components/ambient-api-server/plugins/agents/integration_test.go
+++ b/components/ambient-api-server/plugins/agents/integration_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 
 	. "github.com/onsi/gomega"
@@ -11,6 +12,7 @@ import (
 	"github.com/ambient-code/platform/components/ambient-api-server/pkg/api/openapi"
 	"github.com/ambient-code/platform/components/ambient-api-server/plugins/projects"
 	"github.com/ambient-code/platform/components/ambient-api-server/test"
+	"github.com/openshift-online/rh-trex-ai/pkg/api"
 	"github.com/openshift-online/rh-trex-ai/pkg/environments"
 )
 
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
## 🤖 Amber Review — PR #252: Gateway-mode RBAC regression tests

Solid test addition. CI is all green (Build, Lint, Unit Tests, API Server Integration Tests all passing). Test structure is consistent with existing integration patterns and the gomega assertion style is correct throughout. No production code touched.

**What's good:**
- ✅ `ensureBuiltInRoles` with `ON CONFLICT (name) DO NOTHING` is idempotent — safe to run across multiple tests
- ✅ `seedProjectOwnerBinding` correctly looks up the role ID before inserting the binding
- ✅ `TestAgentDeleteInGatewayMode` verifies the deleted agent returns 404 — a detail many delete tests skip
- ✅ `TestAgentCRUDFullLifecycleWithRBAC` covers the full Create→Get→Patch→List→Delete chain, including a list count assertion
- ✅ All operations use authenticated contexts with proper random accounts — no shared state between tests

**Minor notes (not blocking):**

**1. Test names say "InGatewayMode" but gateway mode isn't activated in the test body.** These tests validate that RBAC correctly permits a `project:owner` to perform agent CRUD — which is the right thing to test, and would correctly catch regression if someone re-added the gateway-mode blocks. But a reader landing on `TestAgentCreateInGatewayMode` and searching for a gateway mode flag or env var will come up empty. Consider renaming to `TestAgentCreateWithRBACOwnerBinding` or similar, or add a comment explaining the naming (e.g., `// Validates no gateway-mode guard blocks project:owner agent creation`).

**2. `seedProjectOwnerBinding` silently no-ops if `roleID == ""`** (i.e., if `ensureBuiltInRoles` failed silently). The test will then proceed and fail with an auth error (403) rather than a clear "role not found" diagnostic. Consider:
```go
if roleID == "" {
    panic("project:owner role not found — did ensureBuiltInRoles() run?")
}
```
`panic` is acceptable in test setup helpers (not production code); or use `t.Fatalf` if you thread `testing.T` through.

**3. No negative-path assertion.** These tests only verify that *authorized* users can perform operations. There's no assertion that a user *without* a role binding is blocked. This may be covered by existing tests elsewhere — if so, a comment pointing to them would help. If not, adding at least one `403` assertion (e.g., a second account without a binding attempts create and gets forbidden) would strengthen the regression coverage.

None of these are blockers — approving.

*— Amber*

PR #253, opened 2026-07-03
feat(control-plane): implement Application reconciler with full integration tests

markturansky:
## Summary

- Implements the previously-stubbed `fetchDeclarations()` and `applyDeclarations()` methods in the Application reconciler, completing the GitOps sync lifecycle
- Introduces a `GitFetcher` interface for dependency injection — production uses `execGitFetcher` (git clone + YAML parse), tests use a mock
- `applyDeclarations()` performs full agent lifecycle management: list existing agents, compare content hashes, create new, update changed, skip unchanged, and optionally prune removed agents
- Only prunes application-managed agents (`ambient.ai/source == "application"`), safely leaving manually-created and configmap-managed agents untouched

## Test plan

- [x] 20 new integration tests covering all reconciler paths:
  - `TestFetchDeclarations_ReturnsMockDeclarations` — mock fetcher returns declarations
  - `TestFetchDeclarations_PropagatesFetcherError` — fetch errors propagate correctly
  - `TestFetchDeclarations_EmptyDeclarationsIsValid` — empty result is not an error
  - `TestApplyDeclarations_CreatesNewAgents` — creates agents via SDK when none exist
  - `TestApplyDeclarations_UpdatesChangedAgents` — updates agents when content hash changes
  - `TestApplyDeclarations_SkipsUnchangedAgents` — no API call when hash matches
  - `TestApplyDeclarations_PrunesRemovedAgents` — deletes stale agents when AutoPrune=true
  - `TestApplyDeclarations_NoPruneWhenAutoFalse` — no deletion when AutoPrune=false
  - `TestApplyDeclarations_EmptyDeclarationsNoOp` — nil declarations = no API calls
  - `TestApplyDeclarations_MixedCreateUpdateSkip` — mixed operations in single pass
  - `TestApplyDeclarations_MapsAllFieldsToAgent` — all declaration fields map to agent
  - `TestApplyDeclarations_DoesNotPruneNonApplicationAgents` — manual/configmap agents safe
  - `TestReconcileApplication_SkipsNonRunningPhase` — only processes Running apps
  - `TestReconcileApplication_SuccessfulSync` — full cycle: fetch → apply → Synced status
  - `TestReconcileApplication_FetchFailureSetsFailedStatus` — fetch error → Failed status
  - `TestReconcileApplication_ApplyFailureSetsFailedStatus` — apply error → Failed status
  - `TestReconcileOnce_ProcessesOnlyRunningApplications` — filters by operation phase
  - `TestReconcileOnce_HandlesEmptyApplicationList` — empty app list = no-op
- [x] `go vet ./...` passes
- [x] `gofmt -l .` clean
- [x] `go build ./...` succeeds
- [x] All 59 reconciler package tests pass

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/application_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/application_reconciler.go
@@ -6,6 +6,11 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"io/fs"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
 	"time"
 
 	sdkclient "github.com/ambient-code/platform/components/ambient-sdk/go-sdk/client"
@@ -28,17 +33,25 @@ const (
 	opPhaseRunning       = "Running"
 
 	hashErrorSentinel = "<hash-error>"
+
+	annotationSourceApplication = "application"
 )
 
+type GitFetcher interface {
+	FetchDeclarations(repoURL, path, targetRevision string) ([]gitAgentDeclaration, string, error)
+}
+
 type ApplicationReconciler struct {
-	factory *SDKClientFactory
-	logger  zerolog.Logger
+	factory    *SDKClientFactory
+	logger     zerolog.Logger
+	gitFetcher GitFetcher
 }
 
 func NewApplicationReconciler(factory *SDKClientFactory, logger zerolog.Logger) *ApplicationReconciler {
 	return &ApplicationReconciler{
-		factory: factory,
-		logger:  logger.With().Str("component", "application-reconciler").Logger(),
+		factory:    factory,
+		logger:     logger.With().Str("component", "application-reconciler").Logger(),
+		gitFetcher: &execGitFetcher{},
 	}
 }
 
@@ -174,9 +187,9 @@ func (r *ApplicationReconciler) fetchDeclarations(app *types.Application) ([]git
 		Str("repo", app.SourceRepoURL).
 		Str("path", app.SourcePath).
 		Str("revision", app.SourceTargetRevision).
-		Msg("fetching declarations from git (stub — git clone not yet implemented)")
+		Msg("fetching declarations from git")
 
-	return nil, app.SourceTargetRevision, nil
+	return r.gitFetcher.FetchDeclarations(app.SourceRepoURL, app.SourcePath, app.SourceTargetRevision)
 }
 
 func (r *ApplicationReconciler) applyDeclarations(ctx context.Context, app *types.Application, declarations []gitAgentDeclaration) error {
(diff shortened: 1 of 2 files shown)


jsell-rh (changes requested):
## 🤖 Amber Review — PR #253: Application Reconciler Implementation

This is solid work — the `GitFetcher` interface for DI, hash-based idempotency, and the 20-test suite covering all reconciler paths are all exactly right. Three issues need fixing before merge.

---

### Blockers

**1. `GitFetcher.FetchDeclarations` has no `context.Context` — git clone is uninterruptible**

The interface signature is:
```go
type GitFetcher interface {
    FetchDeclarations(repoURL, path, targetRevision string) ([]gitAgentDeclaration, string, error)
}
```

`execGitFetcher` calls `exec.Command("git", cloneArgs...)` with no way to cancel it. If a source repo is slow, unreachable, or the clone takes too long, the reconciler goroutine blocks forever — with no timeout, no context cancellation, nothing. In a reconciliation loop this will pile up goroutines and eventually deadlock the reconciler.

Fix: add `ctx context.Context` as the first parameter to the interface and use `exec.CommandContext(ctx, "git", cloneArgs...)` in the implementation. This is a one-line change with widespread correctness impact.

---

**2. Prune errors silently swallowed — violates CLAUDE.md "Never silently swallow partial failures"**

```go
if app.AutoPrune {
    for _, existing := range existingAgents {
        if !declaredNames[existing.Name] && r.isApplicationManaged(existing.Annotations) {
            if deleteErr := client.Agents().Delete(ctx, existing.ID); deleteErr != nil {
                r.logger.Warn().Err(deleteErr).Str("agent", existing.Name).Msg("failed to delete pruned agent")
            } else {
                // ...
            }
        }
    }
}
```

When a prune delete fails, the error is logged as Warn and discarded. `applyDeclarations` returns `nil`, the reconciler patches the Application status as `Synced/Healthy/Succeeded`, and the stale agent remains. The operator now believes pruning succeeded when it didn't.

CLAUDE.md is explicit: **"Never silently swallow partial failures: Every error path must propagate or be collected, not discarded."**

Fix: collect prune errors and either return a combined error (which will cause the reconciler to set `Failed` status) or at minimum update the resource status to reflect partial failure. A combined-error approach is simplest:

```go
var pruneErrs []string
for _, existing := range existingAgents {
    if !declaredNames[existing.Name] && r.isApplicationManaged(existing.Annotations) {
        if deleteErr := client.Agents().Delete(ctx, existing.ID); deleteErr != nil {
            r.logger.Warn().Err(deleteErr).Str("agent", existing.Name).Msg("failed to delete pruned agent")
            pruneErrs = append(pruneErrs, existing.Name)
        }
    }
}
if len(pruneErrs) > 0 {
    return fmt.Errorf("prune failed for agents: %s", strings.Join(pruneErrs, ", "))
}
```

---

### Majors

**3. `json.Marshal` errors silently ignored in three places**

```go
annJSON, _ := json.Marshal(annotations)   // buildAgentPatch
annJSON, _ := json.Marshal(annotations)   // buildAgentResource
data, _ := json.Marshal(m)                // marshalStringMap
```

`map[string]string` marshaling can only fail with non-UTF-8 byte sequences, which won't happen with well-formed annotation values — so the risk is low. But ignoring errors is convention-breaking here. Either handle them:
```go
annJSON, err := json.Marshal(annotations)
if err != nil {
    return nil, fmt.Errorf("marshal annotations: %w", err)  // propagate up
}
```
…or add a note explaining why the ignore is intentional. Same applies to `json.Marshal(resourceStatus)` in the debug log path.

---

### Minors

**4. No input validation on `repoURL` or `targetRevision` in `execGitFetcher`**

Both are passed directly to git. `exec.Command` doesn't invoke a shell, so classic shell injection isn't possible. But a `repoURL` starting with `--` could be interpreted as a git option by some git versions (flag injection). A simple sanity check (`strings.HasPrefix(repoURL, "http")` or a URL parse) is a worthwhile defense-in-depth layer, since these values originate from user-controlled Application resources.

**5. Silent skip of nameless YAML declarations is invisible**

```go
if decl.Name == "" {
    return nil  // silently skipped
}
```

A YAML file that fails to parse cleanly into `gitAgentDeclaration` (but doesn't error) or that has no `name:` field is silently skipped. Add a `r.logger.Warn()` so operators can see when declarations are ignored during a sync.

---

### What's Good

- ✅ No `panic()` anywhere — `fmt.Errorf` wrapping throughout
- ✅ `GitFetcher` interface makes the production implementation fully testable
- ✅ Hash-based change detection avoids needless API calls — correct idiomatic pattern
- ✅ `isApplicationManaged()` guard correctly scopes prune to application-owned agents only, preserving manual and configmap-managed agents
- ✅ 20 tests covering all reconciler paths including mixed create/update/skip and prune boundaries
- ✅ `listAllAgents` pagination loop handles all edge cases correctly
- ✅ `GIT_TERMINAL_PROMPT=0` prevents interactive prompts from blocking the container
- ✅ CI is fully green (Build, Lint, CodeQL, Konflux all pass)

**Confidence:** High on blockers 1 and 2. Both are real correctness issues — the context omission is a design gap, and the prune silence is an explicit CLAUDE.md violation.

*— Amber*

PR #265, opened 2026-07-06
fix: unblock session scheduling and credential role-binding

markturansky:
## Summary

- **CP**: Replace hardcoded `"default"` project scope with empty string in SDK client construction (`main.go`). The CP was sending `X-Ambient-Project: default` on every REST request, causing the API server to filter to `project_id = 'default'` — a project that doesn't exist — so the CP saw 0 sessions and never reconciled.
- **RBAC**: Remove `credential:token-reader` from the `InternalRoles` map (`hierarchy.go`). This was blocking `credential:owner` users from assigning `credential:token-reader` via `acpctl apply`. The level-2 hierarchy still prevents unauthorized escalation.
- **CLI**: Add `resolveRoleID()` to the `credential bind` command (`credential/cmd.go`) so it queries the roles API for the KSUID instead of sending the role name as `role_id`. Matches the existing pattern in `apply/cmd.go`.

Also updates `hierarchy_test.go` to assert `credential:token-reader` is no longer internal, and updates `cmd_test.go` to mock the `/roles` endpoint and validate the resolved KSUID.

## Test plan

- [x] `go build` passes for all three components (control-plane, api-server, cli)
- [x] `go test ./pkg/rbac/...` — RBAC unit tests pass (including updated `TestInternalRoles`)
- [x] `go test ./cmd/acpctl/credential/...` — CLI credential tests pass (including updated `TestBindCredential_Success`)
- [x] `go test ./...` — Full CLI test suite passes (24 tests)

### DIFF
--- a/components/ambient-api-server/pkg/rbac/hierarchy.go
+++ b/components/ambient-api-server/pkg/rbac/hierarchy.go
@@ -19,8 +19,7 @@ var RoleLevel = map[string]int{
 }
 
 var InternalRoles = map[string]bool{
-	RoleAgentRunner:           true,
-	RoleCredentialTokenReader: true,
+	RoleAgentRunner: true,
 }
 
 // CanGrant returns true if callerLevel can grant targetRole.
--- a/components/ambient-api-server/pkg/rbac/hierarchy_test.go
+++ b/components/ambient-api-server/pkg/rbac/hierarchy_test.go
@@ -57,8 +57,8 @@ func TestInternalRoles(t *testing.T) {
 	if !InternalRoles[RoleAgentRunner] {
 		t.Error("agent:runner should be internal")
 	}
-	if !InternalRoles[RoleCredentialTokenReader] {
-		t.Error("credential:token-reader should be internal")
+	if InternalRoles[RoleCredentialTokenReader] {
+		t.Error("credential:token-reader should not be internal")
 	}
 	if InternalRoles[RoleProjectOwner] {
 		t.Error("project:owner should not be internal")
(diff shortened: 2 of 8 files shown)


jsell-rh (approved):
🤖 **Amber Analysis — PR #265**

Three coordinated bug fixes, all well-reasoned and properly tested. Approving.

---

**Root causes addressed:**
- CP sent `X-Ambient-Project: default` on every request → API server filtered to a non-existent project → 0 sessions reconciled. Fixed cleanly with `NewServiceClient` (empty project scope).
- `credential:token-reader` was listed as internal → `credential:owner` users couldn't delegate it via `acpctl credential bind`. The defense-in-depth argument holds: the role is always credential-scoped, and `GetToken` independently verifies `AuthResult.CredentialIDs`. Only `agent:runner` should be machine-only/internal.
- CLI was sending the role name string as `role_id` instead of resolving to a KSUID first. Now matches the pattern in `apply/cmd.go`.

**Checklist:**
- [x] No `panic()` in production Go code
- [x] Errors wrapped with `fmt.Errorf("context: %w", err)` throughout
- [x] No tokens in logs — `service_client.go` validates token properties but does not log the value
- [x] Tests updated: `hierarchy_test.go`, `cmd_test.go`, `rbac_e2e_test.sh` all reflect new behavior
- [x] `resolveRoleID` pattern matches existing `apply/cmd.go` precedent
- [x] CI: lint, unit, integration, CLI build all green; Konflux queued (expected)

**Minor observations (non-blocking):**

1. `service_client.go:19-22` — The placeholder-string checks (`"YOUR_TOKEN_HERE"`, `"PLACEHOLDER_TOKEN"`) are unusual in a service client. Service clients are called from code, not by humans pasting tokens. These guards add noise and won't catch real misconfigurations. Consider removing, or moving to a validator that runs only in dev/debug mode.

2. `credential/cmd.go` — `"credential:viewer"` is a hardcoded string literal. If the role name ever changes server-side this breaks silently. Existing pattern in the codebase (matches `apply/cmd.go`), so not a new problem, but worth tracking as tech debt.

**Confidence:** [95%] High — fixes are targeted, well-documented, and backed by updated tests.

*— Amber*

PR #267, opened 2026-07-07
fix: deliver initial prompt via file upload and increase sandbox timeout

bsquizz:
## Summary

- **Remove `IsServiceCaller` restriction** on `PushSessionMessage` `event_type=user` — allows the runner to push the initial prompt as a user message via gRPC
- **Control plane**: deliver assembled prompt as `/tmp/initial_prompt.txt` via SSH payload upload instead of `INITIAL_PROMPT` env var (gateway path only; operator Job path unchanged)
- **Runner**: read prompt from file (falling back to env var for operator compat), call `_push_initial_prompt_via_grpc` instead of skipping in gRPC mode
- **Increase sandbox readiness timeout** from 120s to 600s (10 minutes) to accommodate large image pulls that exceeded the previous timeout

OpenShell strips env vars containing newlines, which silently dropped `INITIAL_PROMPT`. The runner started idle with no prompt to execute and no messages appeared in the UI. Additionally, large sandbox images can take several minutes to pull, causing timeout failures before the sandbox is ready.

## Test plan

- [ ] `make kind-rebuild` to deploy all three components
- [ ] Create a session via the UI
- [ ] Confirm CP logs show payload upload for `/tmp/initial_prompt.txt`
- [ ] Confirm runner logs show `INITIAL_PROMPT pushed via gRPC`
- [ ] Verify UI chat tab shows both user prompt and assistant response
- [ ] Verify operator (non-gateway) Job path still works via `INITIAL_PROMPT` env var
- [ ] Verify sandbox with large image does not time out during image pull
- [ ] Runner tests pass: `cd components/runners/ambient-runner && python -m pytest tests/ -x -q`

### DIFF
--- a/components/ambient-api-server/plugins/sessions/grpc_handler.go
+++ b/components/ambient-api-server/plugins/sessions/grpc_handler.go
@@ -308,10 +308,6 @@ func (h *sessionGRPCHandler) PushSessionMessage(ctx context.Context, req *pb.Pus
 	if req.GetEventType() == "" {
 		return nil, status.Error(codes.InvalidArgument, "event_type is required")
 	}
-	if middleware.IsServiceCaller(ctx) && req.GetEventType() == "user" {
-		return nil, status.Error(codes.PermissionDenied, "service token may not push event_type=user")
-	}
-
 	if !middleware.IsServiceCaller(ctx) {
 		session, svcErr := h.service.Get(ctx, req.GetSessionId())
 		if svcErr != nil {
--- a/components/ambient-api-server/plugins/sessions/grpc_handler.go
+++ b/components/ambient-api-server/plugins/sessions/grpc_handler.go
@@ -308,6 +308,9 @@ func (h *sessionGRPCHandler) PushSessionMessage(ctx context.Context, req *pb.Pus
 	if req.GetEventType() == "" {
 		return nil, status.Error(codes.InvalidArgument, "event_type is required")
 	}
+	// Service callers (runner, control plane) are trusted components authenticated
+	// via SA tokens scoped to the session namespace. They need to push event_type=user
+	// for the initial prompt so it appears in the UI chat history.
 	if !middleware.IsServiceCaller(ctx) {
 		session, svcErr := h.service.Get(ctx, req.GetSessionId())
 		if svcErr != nil {
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -533,7 +533,7 @@ func (r *SimpleKubeReconciler) mergeAgentEnvironment(env map[string]string, agen
 }
 
 func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID string, entrypoint []string, sdk *sdkclient.Client, execEnv map[string]string, payloads []types.Payload) {
-	pollCtx, pollCancel := context.WithTimeout(context.Background(), 120*time.Second)
+	pollCtx, pollCancel := context.WithTimeout(context.Background(), 600*time.Second)
 	defer pollCancel()
 
 	failSession := func(reason string) {
@@ -561,7 +561,7 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 				Str("sandbox", sbxName).
 				Str("session_id", sessionID).
 				Msg("timed out waiting for sandbox to become ready")
-			failSession("sandbox did not become ready within 120s")
+			failSession("sandbox did not become ready within 600s")
 			return
 		case <-ticker.C:
 			resp, err := r.gateway.GetSandbox(pollCtx, namespace, sbxName)
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -348,12 +348,14 @@ func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, sess
 			r.logger.Warn().Err(err).Str("sandbox", sbxName).Msg("failed to patch sandbox dnsConfig; DNS resolution for external FQDNs may fail")
 		}
 		execEnv := r.inferenceExecEnv()
-		execEntrypoint := r.appendPromptToEntrypoint(ctx, entrypoint, session, sdk)
 		var payloads []types.Payload
 		if agent != nil {
 			payloads = agent.Payloads
 		}
-		go r.execAfterReady(namespace, sbxName, session.ID, execEntrypoint, sdk, execEnv, payloads)
+		if prompt := r.assembleInitialPrompt(ctx, session, sdk); prompt != "" {
+			payloads = append(payloads, types.Payload{SandboxPath: "/tmp/initial_prompt.txt", Content: prompt})
+		}
+		go r.execAfterReady(namespace, sbxName, session.ID, entrypoint, sdk, execEnv, payloads)
 		r.updateSessionPhaseWithNamespace(ctx, session, PhaseCreating, namespace)
 		return nil
 	}
@@ -409,12 +411,14 @@ func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, sess
 		Msg("sandbox created via gateway")
 
 	execEnv := r.inferenceExecEnv()
-	execEntrypoint := r.appendPromptToEntrypoint(ctx, entrypoint, session, sdk)
 	var payloads []types.Payload
 	if agent != nil {
 		payloads = agent.Payloads
 	}
-	go r.execAfterReady(namespace, sbxName, session.ID, execEntrypoint, sdk, execEnv, payloads)
+	if prompt := r.assembleInitialPrompt(ctx, session, sdk); prompt != "" {
+		payloads = append(payloads, types.Payload{SandboxPath: "/tmp/initial_prompt.txt", Content: prompt})
+	}
+	go r.execAfterReady(namespace, sbxName, session.ID, entrypoint, sdk, execEnv, payloads)
 
 	r.updateSessionPhaseWithNamespace(ctx, session, PhaseCreating, namespace)
 	return nil
@@ -1013,9 +1017,6 @@ func (r *SimpleKubeReconciler) buildSandboxEnv(ctx context.Context, session type
 		env["GCE_METADATA_TIMEOUT"] = "1"
 	}
 
-	if prompt := r.assembleInitialPrompt(ctx, session, sdk); prompt != "" {
-		env["INITIAL_PROMPT"] = prompt
-	}
 	if session.LlmModel != "" {
 		env["LLM_MODEL"] = session.LlmModel
 	}
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -42,8 +42,9 @@ func validateTSLValue(value string) error {
 }
 
 const (
-	mcpSidecarPort = int64(8090)
-	mcpSidecarURL  = "http://localhost:8090"
+	mcpSidecarPort   = int64(8090)
+	mcpSidecarURL    = "http://localhost:8090"
+	initialPromptPath = "/tmp/initial_prompt.txt"
 )
 
 type credentialSidecarSpec struct {
@@ -104,7 +105,8 @@ type KubeReconcilerConfig struct {
 	OpenShellPolicyName      string
 	ServiceIdentity          string
 	CACertFile               string
-	AllowedSandboxRegistries []string
+	AllowedSandboxRegistries       []string
+	SandboxReadinessTimeoutSeconds int
 }
 
 type SimpleKubeReconciler struct {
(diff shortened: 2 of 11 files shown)


components/ambient-api-server/plugins/sessions/grpc_handler.go
```diff
@@ -308,10 +308,6 @@ func (h *sessionGRPCHandler) PushSessionMessage(ctx context.Context, req *pb.Pus
 	if req.GetEventType() == "" {
 		return nil, status.Error(codes.InvalidArgument, "event_type is required")
 	}
-	if middleware.IsServiceCaller(ctx) && req.GetEventType() == "user" {
```
bsquizz, line 311: 🤖 **Amber Analysis — Security Flag**

This guard was the only thing preventing service callers (runner SA, control plane SA) from pushing `event_type=user` messages. Removing it entirely means *any* service with a valid SA token can now create messages that appear user-authored.

I understand the necessity — the runner now needs to push the initial prompt as a user message via gRPC since the prompt arrives via file instead of DB. But the fix is broader than needed.

**Suggested alternative:** Rather than removing the guard entirely, consider scoping it more narrowly — e.g., allow service callers to push `event_type=user` only when a specific metadata flag is set (like `source=runner_initial_prompt`), or add a dedicated `event_type=initial_prompt` that the listener treats as a user message but doesn't require the same authorization.

**Confidence:** Medium (80%) — The practical blast radius may be low since only the runner and CP hold SA tokens today, but defense-in-depth matters for multi-tenant systems (Constitution Principle II).

**If you decide to keep the removal:** Please add a code comment explaining *why* the guard was removed and what compensating control exists (e.g., "Service callers are trusted components — runner and CP — authenticated via SA tokens scoped to the session namespace").

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -348,12 +348,14 @@ func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, sess
...
 		var payloads []types.Payload
 		if agent != nil {
 			payloads = agent.Payloads
 		}
-		go r.execAfterReady(namespace, sbxName, session.ID, execEntrypoint, sdk, execEnv, payloads)
+		if prompt := r.assembleInitialPrompt(ctx, session, sdk); prompt != "" {
+			payloads = append(payloads, types.Payload{SandboxPath: "/tmp/initial_prompt.txt", Content: prompt})
+		}
+		go r.execAfterReady(namespace, sbxName, session.ID, entrypoint, sdk, execEnv, payloads)
 		r.updateSessionPhaseWithNamespace(ctx, session, PhaseCreating, namespace)
```
bsquizz: 🤖 **Amber Analysis — Duplicated Code**

This 3-line payload assembly block is duplicated verbatim at line ~421 (the second `provisionSessionSandbox` branch). Both do:
```go
if prompt := r.assembleInitialPrompt(ctx, session, sdk); prompt != "" {
    payloads = append(payloads, types.Payload{SandboxPath: "/tmp/initial_prompt.txt", Content: prompt})
}
```

Consider extracting a small helper like:
```go
func (r *SimpleKubeReconciler) appendInitialPromptPayload(ctx context.Context, session types.Session, sdk *sdkclient.Client, payloads []types.Payload) []types.Payload {
    if prompt := r.assembleInitialPrompt(ctx, session, sdk); prompt != "" {
        return append(payloads, types.Payload{SandboxPath: initialPromptPath, Content: prompt})
    }
    return payloads
}
```

This also centralizes the `/tmp/initial_prompt.txt` path into one place (see the runner-side comment about hardcoded paths).

**Confidence:** High (95%) — Straightforward dedup, no behavioral change.

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -529,7 +533,7 @@ func (r *SimpleKubeReconciler) mergeAgentEnvironment(env map[string]string, agen
 }
 
 func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID string, entrypoint []string, sdk *sdkclient.Client, execEnv map[string]string, payloads []types.Payload) {
-	pollCtx, pollCancel := context.WithTimeout(context.Background(), 120*time.Second)
+	pollCtx, pollCancel := context.WithTimeout(context.Background(), 600*time.Second)
```
bsquizz: 🤖 **Amber Analysis — UX concern with 5x timeout increase**

Increasing from 120s to 600s addresses the legitimate problem of large image pulls, but it also means a *broken* sandbox won't fail the session for 10 full minutes. That's a long time for a user staring at "Creating..." in the UI.

Two suggestions:

1. **Add periodic progress logging** in the poll loop below — e.g., every 30s emit a log line like `"still waiting for sandbox readiness: elapsed=90s sandbox=%s"` so operators can see the session isn't stuck.

2. **Consider making the timeout configurable** via an env var (e.g., `SANDBOX_READINESS_TIMEOUT_SECONDS`) with 600 as the default. This lets deployments with faster registries use a tighter timeout without a code change.

**Confidence:** Medium (85%) — The timeout increase itself is correct; these are quality-of-life improvements.

components/runners/ambient-runner/ambient_runner/app.py
```diff
@@ -385,6 +389,14 @@ def _get_workflow_startup_prompt() -> str:
 # ------------------------------------------------------------------
 
 
+def _read_initial_prompt_file(path: str) -> str:
+    try:
+        with open(path, "r") as f:
+            return f.read().strip()
+    except FileNotFoundError:
```
bsquizz, line 398: 🤖 **Amber Analysis — Bug: catches only `FileNotFoundError`**

This only handles the case where the file doesn't exist. If the file exists but has a permissions issue, is a broken symlink, or the filesystem has an I/O error, this will raise an unhandled exception and crash the startup path.

```python
def _read_initial_prompt_file(path: str) -> str:
    try:
        with open(path, "r") as f:
            return f.read().strip()
    except (FileNotFoundError, PermissionError, OSError):
        return ""
```

Catching `OSError` (parent of both `FileNotFoundError` and `PermissionError`) is the safe choice here, since the fallback to `INITIAL_PROMPT` env var is the correct behavior for any file-read failure. Consider logging a warning for non-`FileNotFoundError` cases so operators know the file was present but unreadable:

```python
except FileNotFoundError:
    return ""
except OSError:
    logger.warning("Failed to read initial prompt file %s, falling back to env var", path, exc_info=True)
    return ""
```

**Confidence:** High (95%) — `PermissionError` in a container with `readOnlyRootFilesystem` or unexpected SecurityContext is a realistic scenario.

components/runners/ambient-runner/ambient_runner/app.py
```diff
@@ -161,7 +161,9 @@ async def lifespan(app: FastAPI):
         if not is_resume:
             # Fetch workflow startupPrompt independently
             workflow_startup_prompt = _get_workflow_startup_prompt()
```
bsquizz, line 165: 🤖 **Amber Analysis — Hardcoded path in 4 places**

The path `/tmp/initial_prompt.txt` appears here, at line 165, at line 199, and in the test file. If this path ever changes (or needs to differ per deployment), all four spots must be updated in lockstep.

Suggest a module-level constant:
```python
_INITIAL_PROMPT_FILE = "/tmp/initial_prompt.txt"
```

This also makes `grep` easier when tracing the prompt delivery path across components.

**Confidence:** High (95%) — Low-effort improvement, standard practice per Constitution Principle VIII (maintainability).

components/runners/ambient-runner/ambient_runner/app.py
```diff
@@ -427,33 +422,15 @@ async def _push_initial_prompt_via_grpc(prompt: str, session_id: str) -> None:
...
-                            "hidden": True,
-                            "autoSent": True,
-                            "source": "runner_initial_prompt",
-                        },
-                    }
-                ],
-            }
             result = client.session_messages.push(
                 session_id,
                 event_type="user",
```
bsquizz, line 437: 🤖 **Amber Analysis — Payload format change**

The gRPC push payload changed from a structured JSON envelope (with `threadId`, `runId`, `messages[]` containing role/content/metadata) to a plain string. This is a significant API contract change.

Is there anything downstream that parses the `PushSessionMessage` payload expecting the old JSON structure? If the API server's `PushSessionMessage` handler, the `WatchSessionMessages` consumer, or the frontend chat rendering expects `messages[].role` / `messages[].content`, this will break silently (the message will appear but with wrong formatting or missing metadata like `hidden: true` and `autoSent: true`).

Please verify:
1. How does the API server store the payload from `PushSessionMessage(event_type="user")`?
2. Does the frontend chat renderer expect structured JSON or raw text for user messages?
3. Were the `hidden` and `autoSent` metadata fields used anywhere downstream?

**Confidence:** Medium (75%) — The old structured payload had specific metadata fields that may have been consumed. If nothing reads those fields, the simplification is great.

components/runners/ambient-runner/tests/test_app_initial_prompt.py
```diff
@@ -22,9 +22,25 @@
...
+# ---------------------------------------------------------------------------
+# _read_initial_prompt_file
+# ---------------------------------------------------------------------------
+
+
+class TestReadInitialPromptFile:
+    def test_reads_existing_file(self, tmp_path):
+        f = tmp_path / "prompt.txt"
+        f.write_text("  hello world  \n")
+        assert _read_initial_prompt_file(str(f)) == "hello world"
```
bsquizz, line 39: 🤖 **Amber Analysis — Missing test: unreadable file**

Good coverage for the happy path and missing file. Consider adding a test for the case where the file exists but is unreadable (relates to the `OSError` catch suggestion on `_read_initial_prompt_file`):

```python
def test_returns_empty_when_unreadable(self, tmp_path):
    f = tmp_path / "prompt.txt"
    f.write_text("hello")
    f.chmod(0o000)
    assert _read_initial_prompt_file(str(f)) == ""
```

This only matters if the exception handling is broadened per the other review comment.

**Confidence:** High (90%) — Defensive test for a realistic container scenario.

bsquizz (commented):
🤖 **Amber Review — PR #267: Initial prompt file delivery + sandbox timeout**

## Summary

The approach is sound — delivering the initial prompt via file upload instead of env var neatly sidesteps OpenShell's newline stripping, and the timeout increase addresses real-world large image pulls. Specs and tests are updated consistently. Nice work.

## Findings (7 inline comments)

| Severity | File | Issue |
|----------|------|-------|
| 🔴 **Security** | `grpc_handler.go` | `IsServiceCaller` guard for `event_type=user` removed entirely — opens service callers to push user-impersonating messages. Consider a narrower exception. |
| 🟡 **Bug** | `app.py:_read_initial_prompt_file` | Only catches `FileNotFoundError` — `PermissionError`/`OSError` will crash the startup path. Catch `OSError` and log a warning. |
| 🟡 **Contract** | `app.py:_push_initial_prompt_via_grpc` | Payload changed from structured JSON envelope to plain string — verify nothing downstream expects the old `messages[].role/content/metadata` format. |
| 🔵 **Code quality** | `kube_reconciler.go` | Payload assembly block duplicated in two `provisionSessionSandbox` branches — extract a helper. |
| 🔵 **Code quality** | `app.py` | `/tmp/initial_prompt.txt` hardcoded in 4 places — use a module constant. |
| 🔵 **UX** | `kube_reconciler.go` | 600s timeout means broken sandboxes take 10 min to fail — add periodic progress logging and consider making configurable. |
| 🔵 **Test gap** | `test_app_initial_prompt.py` | Missing test for unreadable file (permission denied). |

## Verdict

No blockers for the core fix, but the security and error-handling items should be addressed before merge. The rest are quality improvements that can go in a follow-up.

**Confidence:** 85% overall — the prompt delivery mechanism is correct; flagged items are about hardening.

---
*🤖 Amber — Codebase Illuminati*

components/runners/ambient-runner/ambient_runner/app.py
bsquizz: Good question — verified this is safe. The key points:

1. **The old gRPC path never pushed anything.** Before this PR, when `grpc_url` was set, the runner *skipped* the push entirely ("message already in DB via session creation"). The `_push_initial_prompt_via_grpc` function existed with the structured JSON payload but was never called in gRPC mode. So no downstream consumer ever received the old structured format via gRPC.

2. **The HTTP path is unchanged.** The `_push_initial_prompt_via_http` function (operator fallback) still sends the structured JSON to the AG-UI `/run` endpoint — that's a different API with different payload expectations.

3. **`PushSessionMessage` stores payload as-is.** The gRPC handler passes `req.GetPayload()` directly to `msgService.Push()`, which stores it in the `payload` column of `session_messages`. `WatchSessionMessages` streams it back as-is. For `event_type=user`, raw prompt text is the correct payload — the frontend chat renderer displays `payload` directly for user messages.

4. **`hidden` and `autoSent` were never consumed.** Grepped the UI and API server — nothing reads those metadata fields from session message payloads. They were embedded in the JSON that was never sent in gRPC mode anyway.

The simplification from structured JSON to plain text is correct for this path.

jsell-rh (approved):
🤖 **Amber Review — PR#267 Re-check: All feedback addressed**

All six items from the prior review are resolved:

| Finding | Status |
|---|---|
| Security guard removal — missing trust model comment | ✅ Comment added: service callers trusted via SA tokens scoped to session namespace |
| Duplicated payload assembly | ✅ Extracted to `appendInitialPromptPayload()` helper |
| 5× timeout increase with no progress visibility | ✅ Configurable via `SANDBOX_READINESS_TIMEOUT_SECONDS` (default 600s) + 30s progress logging |
| `_read_initial_prompt_file` catches only `FileNotFoundError` | ✅ Broadened to `OSError` with warning log for non-ENOENT cases |
| `/tmp/initial_prompt.txt` hardcoded in 4 places | ✅ Extracted to `_INITIAL_PROMPT_FILE` constant (Go: `initialPromptPath`) |
| Missing test: unreadable file | ✅ `test_returns_empty_when_unreadable` added with `f.chmod(0o000)` |

The payload format change (structured JSON → raw string) was also correctly verified in the prior thread — `hidden`/`autoSent` metadata was never consumed downstream, and the HTTP path is unchanged.

Approved to merge when ready.

*— Amber*

PR #270, opened 2026-07-07
fix(ui): improve tool call result formatting in session logs

bsquizz:
## Summary
- **Deep JSON unwrapping**: Tool call results with multiply-encoded JSON escaping are now recursively unwrapped and pretty-printed, fixing unreadable `\\\"` noise in both the Logs and Chat tabs
- **Collapsible rows**: Tool Call and System events in the Logs tab are collapsed by default, showing just the tool name or "System" label with a chevron toggle
- **Line-based truncation**: Tool Result events truncate to 4 lines with a "Show more" button instead of the previous 300-character cutoff
- **Aligned columns**: Timestamp, badge, and content columns use fixed widths so all log rows align consistently

## Test plan
- [ ] Open a session with tool call activity and verify Logs tab shows readable JSON
- [ ] Verify Tool Call rows are collapsed by default and expand on click
- [ ] Verify System rows are collapsed by default and expand on click
- [ ] Verify Tool Result rows truncate at 4 lines with "Show more" working
- [ ] Verify Chat tab still renders tool results correctly
- [ ] `npm run build` passes with 0 errors (CI)
- [ ] `vitest` passes (CI)

### DIFF
--- a/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/__tests__/logs-tab.test.tsx
+++ b/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/__tests__/logs-tab.test.tsx
@@ -139,28 +139,37 @@ describe('LogsTab', () => {
 
   it('renders messages that match active filters', async () => {
     const messages = [
-      makeMessage({ id: 'msg-1', eventType: 'tool_use', payload: 'git status' }),
+      makeMessage({
+        id: 'msg-1',
+        eventType: 'tool_use',
+        payload: JSON.stringify({ tool: 'Bash', arguments: { command: 'git status' } }),
+      }),
       makeMessage({ id: 'msg-2', eventType: 'error', payload: 'Command failed' }),
     ]
     const LogsTab = await loadWithMessages(messages)
 
     render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })
 
-    expect(await screen.findByText('git status')).toBeInTheDocument()
+    // tool_use rows are collapsed by default — only the tool name label is visible
+    expect(await screen.findByText('Bash')).toBeInTheDocument()
     expect(screen.getByText('Command failed')).toBeInTheDocument()
   })
 
   it('hides messages when their filter is toggled off', async () => {
     const messages = [
-      makeMessage({ id: 'msg-1', eventType: 'tool_use', payload: 'git status' }),
+      makeMessage({
+        id: 'msg-1',
+        eventType: 'tool_use',
+        payload: JSON.stringify({ tool: 'Bash', arguments: { command: 'git status' } }),
+      }),
       makeMessage({ id: 'msg-2', eventType: 'error', payload: 'Command failed' }),
     ]
     const LogsTab = await loadWithMessages(messages)
 
     render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })
 
-    // Wait for messages to load
-    expect(await screen.findByText('git status')).toBeInTheDocument()
+    // Wait for messages to load — tool_use rows show tool name when collapsed
+    expect(await screen.findByText('Bash')).toBeInTheDocument()
 
     // Toggle off 'Tool Call' filter -- scope to filter group to avoid
     // matching EventTypeBadge elements in message rows
@@ -168,21 +177,25 @@ describe('LogsTab', () => {
     fireEvent.click(within(filterGroup).getByText('Tool Call'))
 
     // tool_use message should be hidden
-    expect(screen.queryByText('git status')).not.toBeInTheDocument()
+    expect(screen.queryByText('Bash')).not.toBeInTheDocument()
     // error message should still be visible
     expect(screen.getByText('Command failed')).toBeInTheDocument()
   })
 
   it('shows "No events match" when all filters toggled off', async () => {
     const messages = [
-      makeMessage({ id: 'msg-1', eventType: 'tool_use', payload: 'some command' }),
+      makeMessage({
+        id: 'msg-1',
+        eventType: 'tool_use',
+        payload: JSON.stringify({ tool: 'Read', arguments: { path: '/workspace' } }),
+      }),
     ]
     const LogsTab = await loadWithMessages(messages)
 
     render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })
 
-    // Wait for data to load
-    expect(await screen.findByText('some command')).toBeInTheDocument()
+    // Wait for data to load — tool_use rows show tool name when collapsed
+    expect(await screen.findByText('Read')).toBeInTheDocument()
 
     // Toggle off all filters using the filter group buttons.
     // We must scope to the filter group because EventTypeBadge also
--- a/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/__tests__/logs-tab.test.tsx
+++ b/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/__tests__/logs-tab.test.tsx
@@ -182,6 +182,75 @@ describe('LogsTab', () => {
     expect(screen.getByText('Command failed')).toBeInTheDocument()
   })
 
+  it('expands and collapses tool_use rows when clicked', async () => {
+    const messages = [
+      makeMessage({
+        id: 'msg-collapse',
+        eventType: 'tool_use',
+        payload: JSON.stringify({
+          tool: 'Bash',
+          arguments: { command: 'git status' },
+          tool_call_id: 'tc-collapse',
+        }),
+      }),
+    ]
+    const LogsTab = await loadWithMessages(messages)
+
+    render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })
+
+    // Collapsed by default — tool name visible, arguments hidden
+    const toolLabel = await screen.findByText('Bash')
+    expect(toolLabel).toBeInTheDocument()
+    expect(screen.queryByText('git status')).not.toBeInTheDocument()
+
+    // Expand
+    fireEvent.click(toolLabel)
+    expect(screen.getByText(/git status/)).toBeInTheDocument()
+
+    // Collapse again
+    fireEvent.click(toolLabel)
+    expect(screen.queryByText('git status')).not.toBeInTheDocument()
+  })
+
+  it('truncates tool_result to 4 lines with a Show more button', async () => {
+    const longResult = 'line 1\nline 2\nline 3\nline 4\nline 5\nline 6'
+    const messages = [
+      makeMessage({
+        id: 'msg-long-result',
+        eventType: 'tool_result',
+        payload: JSON.stringify({
+          tool_call_id: 'tc-long',
+          result: longResult,
+        }),
+      }),
+    ]
+    const LogsTab = await loadWithMessages(messages)
+
+    render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })
+
+    // Should show first 4 lines but not line 5/6
+    expect(await screen.findByText(/line 1/)).toBeInTheDocument()
+    expect(screen.getByText(/line 4/)).toBeInTheDocument()
+    expect(screen.queryByText(/line 5/)).not.toBeInTheDocument()
+
+    // "Show more" button should be present
+    const showMore = screen.getByText('Show more')
+    expect(showMore).toBeInTheDocument()
+
+    // Click to expand
+    fireEvent.click(showMore)
+    expect(screen.getByText(/line 5/)).toBeInTheDocument()
+    expect(screen.getByText(/line 6/)).toBeInTheDocument()
+
+    // "Show less" button should now be present
+    const showLess = screen.getByText('Show less')
+    expect(showLess).toBeInTheDocument()
+
+    // Collapse back
+    fireEvent.click(showLess)
+    expect(screen.queryByText(/line 5/)).not.toBeInTheDocument()
+  })
+
   it('shows "No events match" when all filters toggled off', async () => {
     const messages = [
       makeMessage({
--- a/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/__tests__/logs-tab.test.tsx
+++ b/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/__tests__/logs-tab.test.tsx
@@ -251,6 +251,21 @@ describe('LogsTab', () => {
     expect(screen.queryByText(/line 5/)).not.toBeInTheDocument()
   })
 
+  it('shows "Unknown Tool" when tool_use payload is not parseable', async () => {
+    const messages = [
+      makeMessage({
+        id: 'msg-bad-tool',
+        eventType: 'tool_use',
+        payload: 'not valid json',
+      }),
+    ]
+    const LogsTab = await loadWithMessages(messages)
+
+    render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })
+
+    expect(await screen.findByText('Unknown Tool')).toBeInTheDocument()
+  })
+
   it('shows "No events match" when all filters toggled off', async () => {
     const messages = [
       makeMessage({
--- a/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
+++ b/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
@@ -1,14 +1,20 @@
 'use client'
 
-import { useState, useCallback } from 'react'
-import { AlertTriangle } from 'lucide-react'
+import { useState } from 'react'
+import { AlertTriangle, ChevronRight, ChevronDown } from 'lucide-react'
 import { Button } from '@/components/ui/button'
+import { deepUnwrapJson, tryFormatJson, tryParseToolPayload } from '@/components/chat-messages'
 import type { DomainSessionMessage } from '@/domain/types'
 import { formatRelativeTime } from '@/lib/format-timestamp'
 import { cn } from '@/lib/utils'
 import { EventTypeBadge } from './event-type-badge'
 
 const MAX_CONTENT_LENGTH = 300
+const MAX_RESULT_LINES = 4
+
+const COL_TIME = 'shrink-0 w-[100px] font-mono text-xs text-muted-foreground pt-0.5'
+const COL_BADGE = 'shrink-0 w-[90px] pt-0.5 flex items-center gap-1'
+const COL_CONTENT = 'min-w-0 flex-1'
 
 function truncateContent(content: string): { text: string; truncated: boolean } {
   if (content.length <= MAX_CONTENT_LENGTH) {
@@ -20,44 +26,104 @@ function truncateContent(content: string): { text: string; truncated: boolean } 
   }
 }
 
-type EventRowProps = {
-  message: DomainSessionMessage
-  isToolResultFollowingToolUse: boolean
+function truncateToLines(text: string, maxLines: number): { text: string; truncated: boolean } {
+  const lines = text.split('\n')
+  if (lines.length <= maxLines) {
+    return { text, truncated: false }
+  }
+  return {
+    text: lines.slice(0, maxLines).join('\n'),
+    truncated: true,
+  }
 }
 
-export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProps) {
+function formatToolResultPayload(message: DomainSessionMessage): string {
+  try {
+    const parsed = JSON.parse(message.payload) as Record<string, unknown>
+    if (typeof parsed.result === 'string') {
+      return deepUnwrapJson(parsed.result)
+    }
+    return JSON.stringify(parsed, null, 2)
+  } catch {
+    return message.payload
+  }
+}
+
+// ---- Collapsible row: collapsed by default, shows a title label ----
+
+function CollapsibleRow({
+  message,
+  label,
+}: {
+  message: DomainSessionMessage
+  label: string
+}) {
   const [expanded, setExpanded] = useState(false)
-  const isError = message.eventType === 'error'
-  const { text, truncated } = truncateContent(message.payload)
-
-  const toggleExpanded = useCallback(() => {
-    setExpanded(prev => !prev)
-  }, [])
-
+  const formattedPayload = tryFormatJson(message.payload)
   const relativeTime = message.createdAt ? formatRelativeTime(message.createdAt) : '--'
-  const ariaLabel = `${message.eventType === 'tool_use' ? 'Tool Call' : message.eventType === 'tool_result' ? 'Tool Result' : message.eventType} event, ${relativeTime}`
 
   return (
     <article
-      aria-label={ariaLabel}
-      className={cn(
-        'flex gap-3 px-3 py-2 text-sm',
-        isError && 'border-l-2 border-l-status-error-foreground bg-status-error/20',
-        isToolResultFollowingToolUse && 'mt-0 pt-1 border-l-2 border-l-border ml-4',
-      )}
+      aria-label={`${label}, ${relativeTime}`}
+      className="flex gap-3 px-3 py-2 text-sm"
     >
-      <span className="shrink-0 font-mono text-xs text-muted-foreground pt-0.5 min-w-[100px]">
-        {relativeTime}
-      </span>
-      <span className="shrink-0 pt-0.5 flex items-center gap-1">
-        {isError && (
-          <AlertTriangle className="h-3.5 w-3.5 text-status-error-foreground" aria-hidden="true" />
-        )}
+      <span className={COL_TIME}>{relativeTime}</span>
+      <span className={COL_BADGE}>
         <EventTypeBadge eventType={message.eventType} />
       </span>
-      <div className="min-w-0 flex-1">
+      <div className={COL_CONTENT}>
+        <button
+          type="button"
+          className="flex items-center gap-1 text-xs font-mono font-medium text-foreground hover:text-primary transition-colors"
+          onClick={() => setExpanded(prev => !prev)}
+          aria-expanded={expanded}
+        >
+          {expanded ? (
+            <ChevronDown className="h-3.5 w-3.5 shrink-0" aria-hidden="true" />
+          ) : (
+            <ChevronRight className="h-3.5 w-3.5 shrink-0" aria-hidden="true" />
+          )}
+          {label}
+        </button>
+        {expanded && (
+          <pre className="mt-1 whitespace-pre-wrap break-words font-mono text-xs text-foreground">
+            {formattedPayload}
+          </pre>
+        )}
+      </div>
+    </article>
+  )
+}
+
+// ---- Tool Result: line-based truncation ----
+
+function ToolResultRow({
+  message,
+  isFollowingToolUse,
+}: {
+  message: DomainSessionMessage
+  isFollowingToolUse: boolean
+}) {
+  const [expanded, setExpanded] = useState(false)
+  const formattedPayload = formatToolResultPayload(message)
+  const { text: truncatedText, truncated } = truncateToLines(formattedPayload, MAX_RESULT_LINES)
+  const relativeTime = message.createdAt ? formatRelativeTime(message.createdAt) : '--'
+
+  return (
+    <article
+      aria-label={`Tool Result, ${relativeTime}`}
+      className={cn(
+        'flex gap-3 px-3 py-2 text-sm',
+        isFollowingToolUse && 'mt-0 pt-1 border-l-2 border-l-border ml-4',
+      )}
+    >
+      <span className={COL_TIME}>{relativeTime}</span>
+      <span className={COL_BADGE}>
+        <EventTypeBadge eventType={message.eventType} />
+      </span>
+      <div className={COL_CONTENT}>
         <pre className="whitespace-pre-wrap break-words font-mono text-xs text-foreground">
-          {expanded ? message.payload : text}
+          {expanded ? formattedPayload : truncatedText}
           {truncated && !expanded && '...'}
         </pre>
         {truncated && (
@@ -65,7 +131,7 @@ export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProp
             variant="link"
             size="sm"
             className="h-auto p-0 text-xs text-muted-foreground"
-            onClick={toggleExpanded}
+            onClick={() => setExpanded(prev => !prev)}
           >
             {expanded ? 'Show less' : 'Show more'}
           </Button>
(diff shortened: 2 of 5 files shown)


components/ambient-ui/src/components/chat-messages.tsx
```diff
@@ -25,6 +25,35 @@ export const CHAT_EVENT_TYPES: ReadonlySet<SessionEventType> = new Set([
...
+  }
+
+  if (typeof current === 'object' && current !== null && !Array.isArray(current)) {
+    const obj = current as Record<string, unknown>
+    if ('result' in obj && Object.keys(obj).every(k => k === 'result' || k === 'tool_call_id')) {
+      return unwrapNestedJson(obj.result)
+    }
+  }
+
+  return current
```
bsquizz: 🤖 **Amber Analysis — Missing recursion depth guard**

`unwrapNestedJson` calls itself recursively via `return unwrapNestedJson(obj.result)` (line 47), but there's no depth limit on that recursion. Each recursive call starts a fresh 10-iteration parse loop.

A crafted payload like `{"result": "{\"result\": \"{\\\"result\\\": ...}\"}"}` with deep `{result}` nesting would cause unbounded stack depth. The data comes from tool responses which are generally trusted, but defense-in-depth matters.

Suggest adding a depth parameter with a cap:

```typescript
function unwrapNestedJson(value: unknown, depth = 0): unknown {
  if (depth > 5 || typeof value !== 'string') return value
  // ...
  if ('result' in obj && Object.keys(obj).every(k => k === 'result' || k === 'tool_call_id')) {
    return unwrapNestedJson(obj.result, depth + 1)
  }
  // ...
}
```

**Confidence:** High (90%) — Low likelihood but trivial fix.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -20,52 +26,157 @@ function truncateContent(content: string): { text: string; truncated: boolean }
...
+  message,
+  label,
+}: {
   message: DomainSessionMessage
-  isToolResultFollowingToolUse: boolean
+  label: string
+}) {
+  const [expanded, setExpanded] = useState(false)
+  const formattedPayload = tryFormatJson(message.payload)
+  const relativeTime = message.createdAt ? formatRelativeTime(message.createdAt) : '--'
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx:57]: 🤖 **Amber Analysis — Eager formatting on collapsed rows hurts render performance**

`tryFormatJson(message.payload)` runs on every render even when the row is collapsed and the result isn't displayed. The same issue exists in `ToolResultRow` (line 99: `formatToolResultPayload`) and `GenericEventRow` (line 147: `tryFormatJson`).

For a session with hundreds of tool calls, this means parsing + pretty-printing every payload on mount even though most rows are collapsed. This will cause noticeable jank on initial load.

Wrap in `useMemo`:

```tsx
const formattedPayload = useMemo(() => tryFormatJson(message.payload), [message.payload])
```

Or compute lazily (only when first expanded):

```tsx
const [formattedPayload, setFormattedPayload] = useState<string | null>(null)
// in the expand handler:
if (!formattedPayload) setFormattedPayload(tryFormatJson(message.payload))
```

The `useMemo` approach is simpler and still avoids re-computing on re-renders from parent state changes.

**Confidence:** High (90%) — Standard React performance pattern. Impact scales with log volume.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -20,52 +26,157 @@ function truncateContent(content: string): { text: string; truncated: boolean }
...
+
+function formatToolResultPayload(message: DomainSessionMessage): string {
+  try {
+    const parsed = JSON.parse(message.payload) as Record<string, unknown>
+    if (typeof parsed.result === 'string') {
+      return deepUnwrapJson(parsed.result)
+    }
+    return JSON.stringify(parsed, null, 2)
+  } catch {
+    return message.payload
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx]: 🤖 **Amber Analysis — Duplicate unwrapping logic**

`formatToolResultPayload` reimplements what `tryParseToolResult` in `chat-messages.tsx` already does — parse the payload JSON, extract `.result`, and call `deepUnwrapJson`. The difference is this version doesn't extract `tool_call_id` and handles non-string `.result` differently (falls through to `JSON.stringify(parsed)` instead of using the raw payload).

Consider reusing `tryParseToolResult` to keep a single code path for tool result parsing:

```tsx
function formatToolResultPayload(message: DomainSessionMessage): string {
  const parsed = tryParseToolResult(message.payload)
  if (parsed) return parsed.result
  return tryFormatJson(message.payload)
}
```

This ensures the Logs and Chat tabs use the same unwrapping logic and avoids the two paths drifting apart.

**Confidence:** High (90%) — Straightforward reuse, reduces maintenance surface.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -74,3 +185,29 @@ export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProp
...
+
+// ---- Public EventRow: dispatches to the right sub-component ----
+
+type EventRowProps = {
+  message: DomainSessionMessage
+  isToolResultFollowingToolUse: boolean
+}
+
+export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProps) {
+  if (message.eventType === 'tool_use') {
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx:200]: 🤖 **Amber Analysis — Double payload parse**

`tryParseToolPayload(message.payload)` parses the JSON here to extract the tool name. Then `CollapsibleRow` calls `tryFormatJson(message.payload)` which parses the same payload again. Minor cost per row, but it adds up with many tool_use events.

You could pass the already-parsed data as a prop (e.g., `parsedPayload`) or compute the label inside `CollapsibleRow` so formatting and name extraction share one parse. Not a blocker — just a note for if log rendering perf becomes a concern.

**Confidence:** Medium (80%) — Micro-optimization; mentioning since the performance theme is relevant to this PR.

components/ambient-ui/src/components/__tests__/chat-messages.test.ts
```diff
@@ -25,6 +26,48 @@ function makeMsg(
...
+
+  it('preserves {result} objects that have extra keys beyond tool_call_id', () => {
+    const obj = { result: 'data', extra: 'field' }
+    expect(deepUnwrapJson(JSON.stringify(obj))).toBe(JSON.stringify(obj, null, 2))
+  })
+
+  it('unwraps a JSON-encoded string value', () => {
+    expect(deepUnwrapJson('"just a string"')).toBe('just a string')
+  })
+})
```
bsquizz, line 87: 🤖 **Amber Analysis — Missing test: array input**

`unwrapNestedJson` explicitly checks `!Array.isArray(current)` to avoid unwrapping arrays, but there's no test confirming that behavior. A JSON array payload like `[1, 2, 3]` or `["error", "details"]` should be pretty-printed, not unwrapped.

Suggest adding:
```typescript
it('pretty-prints a JSON array without unwrapping', () => {
  const arr = [1, 2, 3]
  expect(deepUnwrapJson(JSON.stringify(arr))).toBe(JSON.stringify(arr, null, 2))
})
```

**Confidence:** High (90%) — Straightforward boundary case for the array guard.

bsquizz (commented):
🤖 **Amber Review — PR #270: Tool call result formatting in session logs**

## Summary

Clean, well-structured refactor. The decomposition of `EventRow` into `CollapsibleRow`, `ToolResultRow`, and `GenericEventRow` is a solid approach — each sub-component has clear responsibility and the dispatcher at the bottom reads naturally. The `deepUnwrapJson` recursive unwrapper is a smart solution to the multiply-encoded JSON problem, and the test coverage is thorough.

## Findings (5 inline comments)

| Severity | File | Issue |
|----------|------|-------|
| 🟡 **Performance** | `event-row.tsx` | All three row components compute `formattedPayload` eagerly even when collapsed — `useMemo` or lazy init would avoid wasted work for sessions with hundreds of log rows |
| 🟡 **Defensive** | `chat-messages.tsx` | `unwrapNestedJson` recurses into `obj.result` without a depth cap — add a depth parameter (trivial fix) |
| 🔵 **DRY** | `event-row.tsx` | `formatToolResultPayload` duplicates logic from `tryParseToolResult` — reuse the existing function to keep Logs and Chat tab rendering consistent |
| 🔵 **Micro-opt** | `event-row.tsx` | `tryParseToolPayload` + `tryFormatJson` double-parses the same payload for `tool_use` rows |
| 🔵 **Test gap** | `chat-messages.test.ts` | Missing test for array input to `deepUnwrapJson` — the `!Array.isArray` guard exists but isn't tested |

## Verdict

No blockers. The core fix works correctly and the UX improvements (collapsible rows, line-based truncation, aligned columns) are all solid. The performance items are worth addressing before merge if sessions with large log volumes are common — otherwise fine as a fast follow-up.

**Confidence:** 90% — Straightforward UI refactor with good test coverage.

---
*🤖 Amber — Codebase Illuminati*

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
Comment (bsquizz) [reply components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx:200]: Acknowledged — agree this is a valid observation but given the reviewer's own "Not a blocker" assessment and the micro-optimization nature (would require threading parsed data through props for marginal per-row savings), skipping this for now. If log rendering perf becomes a concrete issue we can revisit with profiling data to justify the added interface complexity.

components/ambient-ui/src/components/chat-messages.tsx
```diff
@@ -25,6 +25,35 @@ export const CHAT_EVENT_TYPES: ReadonlySet<SessionEventType> = new Set([
 
 // ---- Payload Parsing Helpers ----
 
+function unwrapNestedJson(value: unknown, depth = 0): unknown {
```
bsquizz, line 30: 🤖 **Amber Analysis — Missing test for depth guard**

The `depth > 5` guard was added (good fix from prior review), but there's no test proving it actually stops recursion. Without a test, a future refactor could remove the guard and no CI signal would fire.

Suggest adding:

```typescript
it('stops unwrapping at depth limit and returns the remaining wrapper', () => {
  // Build 7 levels of {result: ...} nesting — deeper than the depth=5 cap
  let payload: string = JSON.stringify({ leaf: true })
  for (let i = 0; i < 7; i++) {
    payload = JSON.stringify({ result: payload })
  }
  const result = deepUnwrapJson(payload)
  // Should NOT reach {leaf: true} — depth guard stops recursion,
  // so the output still contains a {result: ...} wrapper
  expect(result).toContain('"result"')
})
```

**Confidence:** High (95%) — The guard is dead code from the test suite's perspective until this exists.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -20,52 +26,151 @@ function truncateContent(content: string): { text: string; truncated: boolean }
...
 
-export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProps) {
+// ---- Tool Result: line-based truncation ----
+
+function ToolResultRow({
+  message,
+  isFollowingToolUse,
+}: {
+  message: DomainSessionMessage
+  isFollowingToolUse: boolean
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx:101]: 🤖 **Amber Analysis — `useMemo` closes over `message` but deps only track `message.payload`**

`formatToolResultPayload(message)` receives the full `message` object, but the dependency array is `[message.payload]`. This is *functionally* correct since `formatToolResultPayload` only reads `.payload`, but `eslint-plugin-react-hooks` `exhaustive-deps` will flag it because the closure captures `message` while the dep list omits it.

Two options to silence the lint without changing behavior:

```tsx
// Option A: pass the string directly
const formattedPayload = useMemo(() => {
  const parsed = tryParseToolResult(message.payload)
  if (parsed) return parsed.result
  return tryFormatJson(message.payload)
}, [message.payload])

// Option B: extract payload first
const { payload } = message
const formattedPayload = useMemo(() => formatToolResultPayload({ payload } as DomainSessionMessage), [payload])
```

Option A is cleaner — it inlines the 3-line function body and makes the dependency explicit.

**Confidence:** Medium (80%) — Not a runtime bug, but will trigger a lint warning if exhaustive-deps is enabled.

jsell-rh (approved):
🤖 **Amber Re-Review — PR#270: commit `9d8b58eb` (log formatting follow-up)**

Reviewing the incremental changes since my prior feedback. All five items are addressed.

| Finding | Status |
|---|---|
| `useMemo` for expensive formatting in all three row components | ✅ `CollapsibleRow`, `ToolResultRow`, `GenericEventRow` all memoize `formattedPayload` |
| Depth guard on `unwrapNestedJson` recursion | ✅ `if (depth > 5 \|\| typeof value !== 'string') return value` — bounded to max depth 5 |
| `formatToolResultPayload` duplicated `tryParseToolResult` logic | ✅ Now delegates to `tryParseToolResult` then falls through to `tryFormatJson` — single code path |
| Double-parse on `tool_use` rows | ✅ Not applicable — `tool_use` dispatches to `CollapsibleRow` which calls `tryFormatJson` directly, no double-parse |
| Missing test for array input to `deepUnwrapJson` | ✅ `it('pretty-prints a JSON array without unwrapping', ...)` added |

**Additional positive changes:**
- `deepUnwrapJson` exported and covered by 8 dedicated test cases (plain text, simple object, double-stringified, `{result}` wrapper, multiply-encoded, preserved `{result}` with extra keys, encoded string, array)
- `tryParseToolResult` updated to use `deepUnwrapJson(raw)` instead of the old strip-quotes hack — handles the real MCP multi-encoding problem correctly, with matching test coverage
- Logs-tab tests now use proper JSON payloads and assert on collapsed label text (`'Bash'`, `'Read'`) instead of raw payload content — tests match the actual UX
- Spec updated with collapsible row, truncation, and multi-encoding scenarios

**ACP checklist:**
- ✅ Zero `any` types — `unknown` + `Record<string, unknown>` throughout
- ✅ Shadcn Button component for expand/collapse controls
- ✅ `useMemo` for all formatting — avoids recompute on every render
- ✅ `aria-expanded`, `aria-label`, `aria-hidden` on all interactive elements

Approved. This is a clean, well-tested improvement.

*— Amber*

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/__tests__/logs-tab.test.tsx
```diff
@@ -139,50 +139,63 @@ describe('LogsTab', () => {
...
+      makeMessage({
+        id: 'msg-1',
+        eventType: 'tool_use',
+        payload: JSON.stringify({ tool: 'Bash', arguments: { command: 'git status' } }),
+      }),
       makeMessage({ id: 'msg-2', eventType: 'error', payload: 'Command failed' }),
     ]
     const LogsTab = await loadWithMessages(messages)
 
     render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/__tests__/logs-tab.test.tsx:152]: 🤖 **Amber Analysis — New spec scenarios have no test coverage**

The spec update added two new scenarios ("Logs tab collapsible rows" and "Logs tab tool result truncation") that describe expand/collapse and 4-line truncation behavior. The tests here were updated to *work with* collapsed rows, but none of them actually *exercise* the new behavior:

**Missing coverage:**
1. Clicking a collapsed `tool_use` row expands it → payload becomes visible
2. Clicking an expanded row collapses it → payload is hidden again
3. `system` events render as collapsed with "System" label
4. `tool_result` with >4 lines truncates and shows "Show more"
5. Clicking "Show more" reveals full content, button changes to "Show less"

These are all spec requirements now. Without tests, the spec scenarios are aspirational rather than enforced.

Suggest adding at minimum:

```tsx
it('expands a collapsed tool_use row on click to reveal payload', async () => {
  const messages = [
    makeMessage({
      id: 'msg-1',
      eventType: 'tool_use',
      payload: JSON.stringify({ tool: 'Bash', arguments: { command: 'git status' } }),
    }),
  ]
  const LogsTab = await loadWithMessages(messages)
  render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })

  // Collapsed: only tool name visible
  expect(await screen.findByText('Bash')).toBeInTheDocument()
  expect(screen.queryByText('git status')).not.toBeInTheDocument()

  // Expand
  fireEvent.click(screen.getByText('Bash'))
  expect(screen.getByText(/git status/)).toBeInTheDocument()
})
```

**Confidence:** High (90%) — Spec-test alignment is a convention requirement.

components/ambient-ui/src/components/chat-messages.tsx
```diff
@@ -25,6 +25,35 @@ export const CHAT_EVENT_TYPES: ReadonlySet<SessionEventType> = new Set([
...
+  }
+
+  if (typeof current === 'object' && current !== null && !Array.isArray(current)) {
+    const obj = current as Record<string, unknown>
+    if ('result' in obj && Object.keys(obj).every(k => k === 'result' || k === 'tool_call_id')) {
+      return unwrapNestedJson(obj.result, depth + 1)
+    }
+  }
+
+  return current
```
bsquizz, line 50: 🤖 **Amber Analysis — Two interacting iteration limits, non-obvious worst case**

This function has two separate guards: `depth > 5` (recursion through `{result}` wrappers) and `i < 10` (parse loop for string-encoded JSON). These interact multiplicatively — worst case is 5 recursive calls × 10 parse iterations = 50 `JSON.parse` calls per payload.

That's fine for defense-in-depth (50 parses of small strings is cheap), but the two limits serve different purposes and the interaction isn't documented. A future reader seeing `depth > 5` might think the total parse budget is 5, not 50.

Consider a single-line clarification in the depth guard:

```typescript
// depth: caps {result}-wrapper recursion; each level also iterates up to 10 parse rounds
if (depth > 5 || typeof value !== 'string') return value
```

Not a bug — just a readability note for the next person who touches this function.

**Confidence:** Medium (75%) — Readability suggestion, no behavioral impact.

bsquizz (commented):
🤖 **Amber Review — PR #270: Tool call result formatting**

Good improvements overall — collapsible rows, line-based truncation, and deep JSON unwrapping are solid UX wins. The `useMemo` additions and deduplication of `formatToolResultPayload` from the prior review round are clean.

**Summary of findings (2 actionable, 2 informational):**

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | **Medium** | `chat-messages.tsx:28` | Depth guard (`depth > 5`) has no test — the guard is dead code from CI's perspective |
| 2 | **Medium** | `logs-tab.test.tsx:152` | New spec scenarios (expand/collapse, truncation) have zero test coverage |
| 3 | **Low** | `event-row.tsx:99` | `useMemo` deps technically correct but may trigger exhaustive-deps lint |
| 4 | **Low** | `chat-messages.tsx:28-48` | Two interacting iteration limits (5×10=50 parses) — readability note |

Findings 1 and 2 are the ones I'd want addressed before merge — they're both about test coverage matching the new spec scenarios and the defense-in-depth guard. Findings 3 and 4 are notes for future consideration.

Applying `amber/change-requested` for the missing test coverage.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -74,3 +179,29 @@ export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProp
...
+
+type EventRowProps = {
+  message: DomainSessionMessage
+  isToolResultFollowingToolUse: boolean
+}
+
+export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProps) {
+  if (message.eventType === 'tool_use') {
+    const toolPayload = tryParseToolPayload(message.payload)
+    return <CollapsibleRow message={message} label={toolPayload?.name ?? 'Unknown Tool'} />
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx]: **suggestion (test coverage)**

`tryParseToolPayload` returns `null` for non-JSON or payloads missing a `tool`/`name` field, which falls through to `'Unknown Tool'` here. There's no test covering this branch — if the fallback label changes or the null-coalescing is removed, nothing fails.

Consider adding a test in `logs-tab.test.tsx`:

```tsx
it('shows "Unknown Tool" when tool_use payload is not parseable', async () => {
  const messages = [
    makeMessage({
      id: 'msg-bad-tool',
      eventType: 'tool_use',
      payload: 'not valid json',
    }),
  ]
  const LogsTab = await loadWithMessages(messages)
  render(<LogsTab session={makeSession()} />, { wrapper: createWrapper() })
  expect(await screen.findByText('Unknown Tool')).toBeInTheDocument()
})
```

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -20,52 +26,151 @@ function truncateContent(content: string): { text: string; truncated: boolean }
...
+      <span className={COL_TIME}>{relativeTime}</span>
+      <span className={COL_BADGE}>
+        <EventTypeBadge eventType={message.eventType} />
+      </span>
+      <div className={COL_CONTENT}>
+        <button
+          type="button"
+          className="flex items-center gap-1 text-xs font-mono font-medium text-foreground hover:text-primary transition-colors"
+          onClick={() => setExpanded(prev => !prev)}
+          aria-expanded={expanded}
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx:74]: **suggestion (a11y)**

The toggle button has `aria-expanded` but no `aria-controls` linking it to the expandable `<pre>`. Screen readers won't associate the button with the content it reveals.

```tsx
const contentId = `collapsible-${message.id}`
// ...
<button aria-expanded={expanded} aria-controls={contentId} ...>
// ...
{expanded && (
  <pre id={contentId} ...>
```

Minor, since `aria-expanded` already signals expand/collapse intent, but `aria-controls` completes the pattern.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -74,3 +179,29 @@ export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProp
...
+// ---- Public EventRow: dispatches to the right sub-component ----
+
+type EventRowProps = {
+  message: DomainSessionMessage
+  isToolResultFollowingToolUse: boolean
+}
+
+export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProps) {
+  if (message.eventType === 'tool_use') {
+    const toolPayload = tryParseToolPayload(message.payload)
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx]: **nit (consistency)**

`tryParseToolPayload` runs `JSON.parse` on every render of the parent `EventRow`. The child components all use `useMemo` for their parse work — this one is the odd one out.

Not a real perf concern today (parse is fast, messages are immutable), but if someone wraps `EventRow` in `React.memo` later expecting stable internals, this parse would still re-run on every parent re-render. Could be pushed into `CollapsibleRow` as a `useMemo` or memoized here:

```tsx
const label = useMemo(
  () => tryParseToolPayload(message.payload)?.name ?? 'Unknown Tool',
  [message.payload],
)
return <CollapsibleRow message={message} label={label} />
```

Downside: `EventRow` would then call a hook, so the early-return dispatch pattern would need restructuring. Fine to leave as-is if you prefer the clean dispatch shape.

components/ambient-ui/src/components/chat-messages.tsx
```diff
@@ -25,6 +25,37 @@ export const CHAT_EVENT_TYPES: ReadonlySet<SessionEventType> = new Set([
...
+    try {
+      current = JSON.parse(current)
+    } catch {
+      break
+    }
+  }
+
+  if (typeof current === 'object' && current !== null && !Array.isArray(current)) {
+    const obj = current as Record<string, unknown>
+    if ('result' in obj && Object.keys(obj).every(k => k === 'result' || k === 'tool_call_id')) {
```
bsquizz, line 45: **praise**

The allowlist approach here is solid — only peeling off `{result}` wrappers when the object has _no_ unexpected keys (besides `tool_call_id`). This prevents accidentally destroying payloads that happen to have a `result` field alongside other data. The depth guard + iterative parse cap keeps worst-case bounded without being fragile. Clean design.

components/ambient-ui/src/components/__tests__/chat-messages.test.ts
```diff
@@ -25,6 +26,66 @@ function makeMsg(
...
+    for (let i = 0; i < 7; i++) {
+      payload = JSON.stringify({ result: typeof payload === 'string' ? payload : JSON.stringify(payload) })
+    }
+    const result = deepUnwrapJson(payload as string)
+    // Should NOT reach "inner value" — the depth guard stops recursion early,
+    // leaving a partially-unwrapped {result} wrapper as pretty-printed JSON.
+    expect(result).not.toBe('inner value')
+    expect(result).toContain('result')
+  })
+})
```
bsquizz, line 87: **praise**

Good boundary test — verifying the depth guard actually stops recursion rather than just asserting "it doesn't crash." Checking that the result is _not_ the fully unwrapped value and _does_ still contain the wrapper is the right assertion shape.

bsquizz (commented):
🤖 **Amber Review**

**Summary:** Clean refactor — the monolithic `EventRow` is well-decomposed into `CollapsibleRow`, `ToolResultRow`, and `GenericEventRow` with clear dispatch logic. The `deepUnwrapJson` unwrapping is correctly bounded (depth × iteration cap) and the allowlist approach for recognizing `{result}` wrappers avoids false positives on real data objects.

**Confidence:** 90% — architecture is solid, tests cover the key behavioral changes (collapse/expand, truncation, deep unwrapping, depth guard). Two minor gaps noted inline.

**Findings:**
- 1 test coverage gap: `'Unknown Tool'` fallback when `tryParseToolPayload` returns null (no test)
- 1 a11y suggestion: `aria-controls` missing on collapsible toggle buttons
- 1 consistency nit: `tryParseToolPayload` parse in render path not memoized (fine today, see inline note)

None of these are blockers — approving in spirit.

**Spec alignment:** The `views.spec.md` updates correctly document the new collapsible and truncation behavior — spec and implementation match.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -20,52 +26,153 @@ function truncateContent(content: string): { text: string; truncated: boolean }
...
+      <span className={COL_BADGE}>
+        <EventTypeBadge eventType={message.eventType} />
+      </span>
+      <div className={COL_CONTENT}>
+        <button
+          type="button"
+          className="flex items-center gap-1 text-xs font-mono font-medium text-foreground hover:text-primary transition-colors"
+          onClick={() => setExpanded(prev => !prev)}
+          aria-expanded={expanded}
+          aria-controls={contentId}
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx:75]: **🤖 Amber — a11y nit (Low priority)**

`aria-controls={contentId}` always references the content panel, but the `<pre id={contentId}>` only renders when `expanded === true`. When collapsed, this button claims to control an element that doesn't exist in the DOM.

Most screen readers handle orphaned `aria-controls` gracefully (NVDA, VoiceOver ignore it), so this isn't a functional bug. For strict WAI-ARIA compliance, two options:

1. **Conditionally apply** — `aria-controls={expanded ? contentId : undefined}`
2. **Always render, visually hide** — keep the `<pre>` in DOM with `hidden` attribute

Option 1 is the lighter touch and probably the right call here. Not blocking — just flagging for completeness since the rest of the a11y work in this PR (aria-expanded, aria-label, semantic `<article>`) is well done.

components/ambient-ui/src/components/chat-messages.tsx
bsquizz, line 45: **🤖 Amber — design note (preserve this guard)**

This `Object.keys(obj).every(k => k === 'result' || k === 'tool_call_id')` check is doing important work: it distinguishes MCP tool-response wrappers (`{result: "...", tool_call_id: "..."}`) from domain objects that happen to have a `result` field alongside other meaningful data (e.g. `{result: "data", metadata: {...}, pagination: {...}}`). Without this guard, unwrapping would silently discard sibling fields.

The tests at line 61–64 (`'preserves {result} objects that have extra keys'`) cover this, but the invariant is subtle enough that a future editor doing "cleanup" might accidentally broaden the condition. Consider adding a brief inline note:

```ts
// Only unwrap pure MCP wrappers — {result} or {result, tool_call_id}
```

Not blocking — the test coverage is sufficient as a guardrail.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx
```diff
@@ -74,3 +181,35 @@ export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProp
...
+  isToolResultFollowingToolUse: boolean
+}
+
+export function EventRow({ message, isToolResultFollowingToolUse }: EventRowProps) {
+  const toolLabel = useMemo(
+    () => message.eventType === 'tool_use'
+      ? tryParseToolPayload(message.payload)?.name ?? 'Unknown Tool'
+      : null,
+    [message.eventType, message.payload],
+  )
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/event-row.tsx:198]: **🤖 Amber — observation (no action needed)**

`toolLabel` is memoized with `[message.eventType, message.payload]` — inside the `if (message.eventType === 'tool_use')` block below, it's guaranteed to be a non-null string (either the parsed name or `'Unknown Tool'`). The `?? 'Unknown Tool'` fallback at line 201 is logically redundant but needed for TypeScript since the type system can't narrow across the memo boundary.

This is fine — just noting it so nobody spends time trying to "simplify" this later.

bsquizz (commented):
🤖 **Amber Review**

**Summary:** Clean, well-structured refactor that solves a real readability problem. The multiply-encoded JSON in tool results was genuinely unreadable before this change.

**What I like:**
- `deepUnwrapJson` with bounded iteration (10) × bounded recursion (5) is defensive without being over-engineered — the depth-guard test at chat-messages.test.ts:75 confirms the boundary holds
- Component decomposition into `CollapsibleRow` / `ToolResultRow` / `GenericEventRow` follows single-responsibility well and keeps the dispatch in `EventRow` readable
- `useMemo` on all formatting calls prevents unnecessary re-parses on every render
- Line-based truncation (4 lines) is a better UX heuristic than the old 300-char cutoff for tool results that are naturally multi-line
- Proper `aria-expanded`, `aria-label`, and semantic `<article>` elements — good a11y baseline
- Spec updates in `views.spec.md` track the new behavior

**Inline findings (3 — none blocking):**
1. `aria-controls` references a conditionally-rendered element (a11y nit)
2. The MCP-wrapper guard in `unwrapNestedJson` is subtle — suggested a one-line comment for future editors
3. TypeScript narrowing note on `toolLabel ?? 'Unknown Tool'` — no action needed

**Confidence:** 90% — code is sound, tests cover the key paths. The one thing I can't verify from a diff review is whether the formatting looks correct with real production payloads (multiply-encoded Jira/GitHub MCP responses, etc). Visual testing against a live session would close that gap.

**Constitution compliance:** ✅ No `any` types, Shadcn components only, proper error handling, conventional commits.

PR #271, opened 2026-07-07
feat: OpenShell sandbox observability with log and policy persistence

bsquizz:
## Summary

- Stream real-time sandbox logs and policy from OpenShell gateway through the full SSE proxy chain (UI → API Server → Control Plane → Gateway gRPC)
- Persist sandbox logs and policy snapshots to PostgreSQL so they survive sandbox shutdown
- Display historical sandbox data in the UI for terminal sessions with a "Historical" badge

### Control Plane
- Vendor `WatchSandbox`/`GetSandboxLogs` RPCs + message types, regenerate Go stubs
- `WatchSandbox` method on `GatewayClient` with auth/eviction pattern
- `sandbox_handler.go`: gRPC server-stream → SSE conversion with typed events
- Token server `Option` pattern and `/sandbox/{name}/logs|policy` routes
- `FetchSandboxLogs` method for non-streaming log retrieval (tail 500 lines)
- Extract `SandboxPhaseString()` and `PolicyToMap()` to shared `openshell/sandbox_helpers.go`
- `PodStatusSyncer` captures policy (from existing `GetSandbox` response) and logs every 15s into `sandbox_logs_snapshot` and `sandbox_policy_snapshot` session fields
- Pre-delete final snapshot in `deprovisionSessionSandbox` before `DeleteSandbox` call

### API Server
- SSE proxy (`/sessions/{id}/sandbox/logs`) and JSON proxy (`/sessions/{id}/sandbox/policy`)
- Migration: `sandbox_logs_snapshot` and `sandbox_policy_snapshot` TEXT columns on sessions
- Model, service, presenter, and OpenAPI spec updated for new fields

### UI
- Sandbox observability types, adapters, ports, and React Query hooks
- OpenShell tab with sandbox logs and policy sub-tabs
- Historical fallback: terminal sessions display stored snapshot data with "Historical" badge
- Domain type and mapper updated for `sandboxLogsSnapshot` and `sandboxPolicySnapshot`

### SDK
- All three SDKs (Go, Python, TypeScript) regenerated from updated OpenAPI spec

### Proxy (Next.js API route)
- Disable `Accept-Encoding` for SSE to prevent gzip hanging on chunked streams
- `cache: "no-store"` for streaming correctness
- Harden pipe error handling for `ResponseAborted`

### E2E Tests
- Section 6: sandbox log streaming (SSE events, named types, JSON parsing) and policy retrieval
- Section 7: sandbox snapshot persistence — verifies `sandbox_logs_snapshot` and `sandbox_policy_snapshot` are non-null JSON after session stop

### Specs
- `openshell-sandbox-observability.spec.md`: streaming architecture + persistence requirements
- `data-model.spec.md`: snapshot fields on Session entity
- `control-plane.spec.md`: Sandbox Snapshot Collection section

## Test plan
- [x] `go build ./...` passes for API server and control plane
- [x] `npx tsc --noEmit` passes for UI
- [x] All 403 vitest tests pass
- [x] Migration applies and API returns new fields
- [x] Kind cluster deploy successful
- [ ] Run e2e `openshell-dual-tenant.sh` sections 6-7 against a live cluster
- [ ] Verify UI shows "Historical" badge for stopped session logs/policy


components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -1127,6 +1129,68 @@ func (r *SimpleKubeReconciler) deprovisionSessionSandbox(ctx context.Context, se
...
+		}
+	}
+
+	if _, err := sdk.Sessions().UpdateStatus(ctx, session.ID, patch); err != nil {
+		r.logger.Warn().Err(err).Str("session_id", session.ID).Msg("final snapshot: failed to persist")
+	} else {
+		r.logger.Info().Str("session_id", session.ID).Msg("final sandbox snapshot persisted")
+	}
+}
+
```
bsquizz, line 1180: 🤖 **Amber Analysis — Duplicated snapshot logic**

`finalSandboxSnapshot` here is near-identical to `snapshotSandboxData` in `pod_sync.go`. Both:
1. Extract policy from `GetSandbox` response
2. Build the same `policyEnvelope` map with the same 6 keys
3. Call `FetchSandboxLogs(ctx, namespace, sandboxID, 500)`
4. Assemble `patch` map and call `sdk.Sessions().UpdateStatus()`

The only difference is that `finalSandboxSnapshot` calls `GetSandbox` itself while `snapshotSandboxData` receives the response as a parameter.

Extract a shared `buildSnapshotPatch(sbx, logs)` helper (or accept `*pb.SandboxResponse` in both) to keep the field names, log tail count, and envelope structure in one place. Otherwise when the snapshot schema changes (new fields, different log tail), one site will be missed.

**Confidence:** High (95%) — Straightforward dedup, the structures are literally identical.

components/ambient-api-server/plugins/sessions/handler.go
```diff
@@ -1110,6 +1111,151 @@ func (h sessionHandler) OAuthProviderURL(w http.ResponseWriter, r *http.Request)
...
+		}
+		result[i] = c
+	}
+	return "session-" + string(result)
+}
+
+// SandboxLogs proxies sandbox log SSE from the control plane.
+func (h sessionHandler) SandboxLogs(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	id := mux.Vars(r)["id"]
```
bsquizz, line 1149: 🤖 **Amber Analysis — Drift risk: duplicated sandbox name derivation**

`sandboxName()` reimplements the control plane's `openshell.SandboxName()` logic (lowercase, truncate to 40 chars, prepend `session-`). If the CP formula changes, this will silently produce the wrong sandbox name and every API server proxy call will 404.

Two safer options:
1. **Store the sandbox name on the session record** — the CP already knows it during provisioning. Add `sbxName` to the status patch alongside `KubeNamespace`. Then the API server reads it instead of re-deriving.
2. **Move the name derivation to a shared package** under `components/ambient-sdk/` or a `pkg/naming` package both components import.

Option 1 is cleaner since it eliminates the derivation entirely and the value is already known at provisioning time.

**Confidence:** High (90%) — The function works today, but this is exactly the kind of duplication that causes an outage 3 months later when someone changes one side.

components/ambient-api-server/plugins/sessions/handler.go
```diff
@@ -1110,6 +1111,151 @@ func (h sessionHandler) OAuthProviderURL(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+
+	resp, doErr := EventsHTTPClient.Do(req)
+	if doErr != nil {
+		glog.Warningf("SandboxPolicy: CP unreachable for session %s: %v", id, doErr)
+		http.Error(w, "control plane not reachable", http.StatusBadGateway)
+		return
+	}
+	defer func() { _ = resp.Body.Close() }()
```
bsquizz, line 1248: 🤖 **Amber Analysis — Blind header proxying leaks internal metadata**

This copies ALL upstream response headers to the client response, including potentially sensitive internal headers from the control plane (e.g., `Server`, `X-Powered-By`, internal tracing headers, `X-Request-Id`).

Filter to only the headers the client needs:

```go
for _, k := range []string{"Content-Type", "Content-Length"} {
    if v := resp.Header.Get(k); v != "" {
        w.Header().Set(k, v)
    }
}
```

Compare with `SandboxLogs` above which correctly sets only the SSE-required headers. This handler should follow the same pattern.

**Confidence:** High (90%) — Defense-in-depth. The CP is internal, but the API server is the trust boundary — it shouldn't relay internal infrastructure details to clients (Constitution Principle II).

components/ambient-control-plane/internal/openshell/sandbox_helpers.go
```diff
@@ -0,0 +1,75 @@
...
+		event, recvErr := stream.Recv()
+		if recvErr == io.EOF {
+			break
+		}
+		if recvErr != nil {
+			if len(entries) > 0 {
+				return entries, nil
+			}
+			return nil, recvErr
+		}
```
bsquizz: 🤖 **Amber Analysis — Silently swallows errors when partial data exists**

When the stream errors mid-read but some log entries have already been collected, this returns the partial data with a `nil` error. The caller (`snapshotSandboxData` in `pod_sync.go`) checks `if logErr != nil` and will never see the error — it'll persist a truncated snapshot with no indication it's incomplete.

Two concerns:
1. The persisted `sandbox_logs_snapshot` may be missing the most recent (and often most relevant) log lines with no way for the UI or operator to know.
2. Per CLAUDE.md: "Never silently swallow partial failures."

Suggest returning both: `return entries, recvErr` and letting callers decide. Or at minimum, log the error here:

```go
if len(entries) > 0 {
    // Log that we're returning partial results
    return entries, nil
}
```

But the caller can't log without the error, so returning it is better.

**Confidence:** High (90%) — The current behavior is a correctness issue per project conventions.

components/ambient-ui/src/queries/use-sandbox-logs.ts
```diff
@@ -0,0 +1,101 @@
...
+          if (typeof raw.timestamp !== 'number' || typeof raw.message !== 'string') return
+          setEntries(prev => [...prev, raw as unknown as SandboxLogEntry])
+        } catch {
+          // skip unparseable entries
+        }
+      }
+
+      es.onmessage = handleEvent
+      for (const eventType of SSE_EVENT_TYPES) {
+        es.addEventListener(eventType, handleEvent as EventListener)
```
bsquizz, line 91: 🤖 **Amber Analysis — Unbounded memory growth + unsafe cast**

Two issues on this line:

**1. Unbounded entry accumulation:** Every SSE event appends to state forever. A session running for hours could accumulate tens of thousands of entries, consuming significant browser memory. Add a cap:

```typescript
const MAX_LOG_ENTRIES = 5000
setEntries(prev => {
  const next = [...prev, entry]
  return next.length > MAX_LOG_ENTRIES ? next.slice(-MAX_LOG_ENTRIES) : next
})
```

**2. `as unknown as SandboxLogEntry` bypasses type safety:** This is functionally equivalent to `any` — exactly what the project's "zero `any` types" convention prohibits. The validation above only checks `timestamp` and `message` but `SandboxLogEntry` likely has `source`, `level`, `module`, `category`, `denied`, etc. Use a proper type guard or parser:

```typescript
function parseSandboxLogEntry(raw: Record<string, unknown>): SandboxLogEntry | null {
  if (typeof raw.timestamp !== 'number' || typeof raw.message !== 'string') return null
  return {
    timestamp: raw.timestamp,
    message: raw.message,
    source: typeof raw.source === 'string' ? raw.source : 'unknown',
    level: typeof raw.level === 'string' ? raw.level : 'INFO',
    // ...
  }
}
```

**Confidence:** High (90%) for memory cap, High (95%) for type safety.

components/ambient-ui/src/queries/use-sandbox-logs.ts
```diff
@@ -0,0 +1,101 @@
...
+        setIsConnected(false)
+
+        if (enabled) {
+          setIsReconnecting(true)
+          reconnectTimeoutRef.current = setTimeout(connect, 3000)
+        }
+      }
+    }
+
+    connect()
```
bsquizz, line 106: 🤖 **Amber Analysis — Reconnect storm on terminal sessions**

When the sandbox is deleted (session Completed/Failed), the SSE connection will fail immediately. The `onerror` handler reconnects after 3 seconds if `enabled` is still true. Since `enabled` is based on `session.phase === 'Running'`, this should self-resolve when the parent re-renders with the new phase.

But there's a race: if the SSE endpoint returns an error *before* the session status poll updates the phase, this will reconnect → fail → reconnect in a tight 3-second loop until the status catches up. Consider adding a reconnect limit:

```typescript
const MAX_RECONNECTS = 5
const reconnectCountRef = useRef(0)

es.onerror = () => {
  es.close()
  setIsConnected(false)
  if (enabled && reconnectCountRef.current < MAX_RECONNECTS) {
    reconnectCountRef.current++
    setIsReconnecting(true)
    reconnectTimeoutRef.current = setTimeout(connect, 3000)
  }
}
```

Reset `reconnectCountRef` to 0 on successful `onopen`.

**Confidence:** Medium (80%) — Depends on the phase-poll timing, but easy to hit in practice when a session completes.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/sandbox-policy-tab.tsx
```diff
@@ -0,0 +1,191 @@
...
+      .map(([k, v]) =>
+        typeof v === 'object' && v !== null
+          ? `${pad}${k}:\n${toYaml(v, indent + 1)}`
+          : `${pad}${k}: ${toYaml(v, 0).trim()}`,
+      )
+      .join('\n')
+  }
+  return `${pad}${String(value)}`
+}
+
```
Comment (bsquizz) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/sandbox-policy-tab.tsx:56]: 🤖 **Amber Analysis — Custom YAML formatter with no test coverage**

`toYaml` is ~50 lines of recursive code handling arrays, nested objects, null, primitives, and mixed indentation. This is exactly the kind of complex formatting logic that needs unit tests — edge cases like empty arrays inside objects, deeply nested structures, and strings containing special YAML characters (`:`, `#`, `- `) could produce incorrect output.

Either:
1. Add tests in a `__tests__/` file for this function (at minimum: nested objects, arrays of objects, empty values, strings with colons)
2. Replace with a lightweight YAML lib like `yaml` (already commonly bundled)

If keeping the custom implementation, at minimum export the function so it can be tested:

```typescript
// In a separate utils file with tests
export function toYaml(value: unknown, indent = 0): string { ... }
```

**Confidence:** Medium (85%) — It probably works for the current policy shape, but will break on the first unexpected value shape without tests to catch it.

bsquizz (commented):
🤖 **Amber Review — PR #271: OpenShell sandbox observability**

Strong feature — the gRPC→SSE bridge in the token server is well-structured, the test coverage for `sandbox_handler_test.go` is solid, and the ports/adapters pattern in the UI is a good architectural choice.

**Issues that should be addressed:**

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 High | `kube_reconciler.go` | `finalSandboxSnapshot` duplicates `snapshotSandboxData` from `pod_sync.go` — extract shared helper |
| 2 | 🔴 High | `handler.go` | `sandboxName()` reimplements CP's `SandboxName()` — store name on session record instead |
| 3 | 🟡 Medium | `handler.go` | `SandboxPolicy` blindly proxies all upstream headers — filter to Content-Type only |
| 4 | 🟡 Medium | `sandbox_helpers.go` | `FetchSandboxLogs` swallows errors when partial data exists — violates "never silently swallow partial failures" |
| 5 | 🟡 Medium | `use-sandbox-logs.ts` | Unbounded `setEntries(prev => [...prev, ...])` — add a cap |
| 6 | 🟡 Medium | `use-sandbox-logs.ts` | `as unknown as SandboxLogEntry` violates zero-`any` convention — use a type guard |
| 7 | 🟢 Low | `use-sandbox-logs.ts` | SSE reconnect has no retry limit — add max reconnects with reset on success |
| 8 | 🟢 Low | `sandbox-policy-tab.tsx` | Custom `toYaml` (~50 lines of recursive formatting) has no tests — add unit tests or use a library |

**Summary:** Two high-severity duplication issues create real drift risk — the snapshot logic and sandbox name derivation will diverge when either side is modified independently. The frontend issues (unbounded growth, unsafe cast, reconnect storm) are medium-severity but straightforward to fix. Good work on the overall architecture.

**Verdict: `amber/change-requested`** — the high-severity duplication items should be addressed before merge.

*Reviewed against ACP Constitution and CLAUDE.md conventions.*

components/ambient-api-server/plugins/sessions/handler.go
bsquizz: Agree this is a drift risk — option 1 (storing sandbox name on the session record) is the cleanest long-term fix. However, it requires a new DB column, migration, OpenAPI spec change, and SDK regeneration across all three languages — that's a significant scope increase for this PR.

The current function is 15 lines, has a clear comment documenting its relationship to `openshell.SandboxName()`, and the derivation logic is stable (lowercase + truncate + prefix). I'd prefer to address this as a follow-up ticket to keep this PR focused on the observability feature.

Filed as a future improvement — will track separately.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/sandbox-policy-tab.tsx
Comment (bsquizz) [reply components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/sandbox-policy-tab.tsx:56]: Fair point on test coverage. The function currently only renders the policy envelope shape (which is a known, fixed structure from the gateway), so the risk is limited today. But it would break on the first unexpected value shape.

Adding a YAML library dependency for a single display component feels heavyweight. I'd prefer to extract `toYaml` to a shared util with a small test suite as a follow-up — keeping this PR scoped to the observability feature.

Will track as a follow-up.

jsell-rh (approved):
🤖 **Amber Review — PR#271 Re-check: All actionable feedback addressed**

All six items from the prior review are resolved in commit `9baa9b00`:

| Finding | Status |
|---|---|
| Duplicated snapshot logic (`finalSandboxSnapshot` / `snapshotSandboxData`) | ✅ Shared `BuildSnapshotPatch()` helper extracted to `openshell/sandbox_helpers.go`; `LogTailLines` exported as constant |
| `sandboxName()` drift risk | ✅ Explicitly deferred — bsquizz's reasoning is sound (requires DB migration + OpenAPI + 3-SDK regeneration; current 15-line function is stable and documented). Acceptable as a tracked follow-up. |
| `FetchSandboxLogs` silently swallows partial failures | ✅ Now returns `entries, recvErr` when partial data exists — callers see the error |
| Blind header proxying leaks internal metadata | ✅ `SandboxPolicy` proxy now filters to `Content-Type` / `Content-Length` only |
| Unbounded `entries` accumulation + `as unknown as SandboxLogEntry` type bypass | ✅ `MAX_LOG_ENTRIES = 5000` sliding window + `parseSandboxLogEntry()` type guard with field-by-field narrowing |
| SSE reconnect storm on terminal sessions | ✅ `MAX_RECONNECTS = 5` counter with reset on `onopen` |

The `sandbox_helpers_test.go` additions (`BuildSnapshotPatch`, `SandboxPhaseString`, `PolicyToMap`) cover the extracted helpers well.

Approved to merge when ready.

*— Amber*

PR #272, opened 2026-07-07
feat: add custom ACP login theme for Keycloak

squizzi:
## Summary
Screenshot 2026-07-07 at 7 29 13 AM

- Adds a branded Keycloak login page for the Kind dev overlay with the ACP robot logo (Lucide Bot), "Agent Control Plane" title, and Red Hat Text font
- Uses the ambient-ui dark palette (`#151515` background, `#1f1f1f` card, `#f56e6e` accent) with subtle dot-grid texture
- Theme files are mounted via ConfigMap into the upstream `quay.io/keycloak/keycloak:26.0` image — no custom Docker build needed
- Overrides PatternFly v5 login layout to a centered single-column stack with smooth hover transitions on form controls

## Test plan

- [x] Run `kubectl kustomize components/manifests/overlays/kind/ | kubectl apply -f -` and restart Keycloak
- [x] Visit the Keycloak login page and verify gray background, centered layout, robot logo, and "Agent Control Plane" title
- [x] Verify form input focus and password visibility toggle use orange-red accent (`#f56e6e`), not blue
- [x] Verify hover transitions fade smoothly on input borders and eye icon

### DIFF
--- a/components/manifests/overlays/kind/keycloak-theme/acp-logo.svg
+++ b/components/manifests/overlays/kind/keycloak-theme/acp-logo.svg
@@ -1,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="80" height="80" viewBox="0 0 24 24" fill="none" stroke="#f56e6e" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+  <path d="M12 8V4H8"/>
+  <rect width="16" height="12" x="4" y="8" rx="2"/>
+  <path d="M2 14h2"/>
+  <path d="M20 14h2"/>
+  <path d="M15 13v2"/>
+  <path d="M9 13v2"/>
+</svg>
--- a/components/manifests/overlays/kind/keycloak-theme/acp-logo.svg
+++ b/components/manifests/overlays/kind/keycloak-theme/acp-logo.svg
@@ -1,4 +1,5 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="80" height="80" viewBox="0 0 24 24" fill="none" stroke="#f56e6e" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+<svg xmlns="http://www.w3.org/2000/svg" width="80" height="80" viewBox="0 0 24 24" fill="none" stroke="#f56e6e" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" aria-label="ACP Logo">
+  <title>ACP Logo</title>
   <path d="M12 8V4H8"/>
   <rect width="16" height="12" x="4" y="8" rx="2"/>
   <path d="M2 14h2"/>
--- a/components/manifests/overlays/kind/keycloak-theme/login.css
+++ b/components/manifests/overlays/kind/keycloak-theme/login.css
@@ -1,6 +1,6 @@
 /* ACP Keycloak Login Theme — matches ambient-ui dark palette */
 
-/* Import Red Hat Text to match the main UI */
+/* Red Hat Text from Google Fonts — requires network access; falls back to system-ui offline */
 @import url('https://fonts.googleapis.com/css2?family=Red+Hat+Text:wght@400;500;600;700&display=swap');
 
 /* Page background — true gray with subtle dot texture */
@@ -27,7 +27,7 @@ html.pf-v5-theme-dark .pf-v5-c-login {
   grid-template-areas: none !important;
 }
 
-/* Header — hide raw text from kc-header-wrapper, use pseudo-elements */
+/* Header — hide raw text from kc-header-wrapper (Keycloak 26.0 / PF5 DOM) */
 #kc-header-wrapper {
   font-size: 0 !important;
   line-height: 0 !important;
@@ -51,25 +51,26 @@ html.pf-v5-theme-dark .pf-v5-c-login {
 }
 
 .pf-v5-c-login__header::before {
-  content: "" !important;
-  display: block !important;
-  width: 88px !important;
-  height: 88px !important;
-  background-image: url("../img/acp-logo.svg") !important;
-  background-size: contain !important;
-  background-repeat: no-repeat !important;
-  background-position: center !important;
-  margin-bottom: 0.5rem !important;
+  content: "";
+  display: block;
+  width: 88px;
+  height: 88px;
+  background-image: url("../img/acp-logo.svg");
+  background-size: contain;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-bottom: 0.5rem;
 }
 
+/* Title must match messages_en.properties:loginTitleHtml */
 .pf-v5-c-login__header::after {
-  content: "Agent Control Plane" !important;
-  display: block !important;
-  color: #f2f2f2 !important;
-  font-size: 1.75rem !important;
-  font-weight: 600 !important;
-  font-family: 'Red Hat Text', system-ui, -apple-system, sans-serif !important;
-  letter-spacing: 0.01em !important;
+  content: "Agent Control Plane";
+  display: block;
+  color: #f2f2f2;
+  font-size: 1.75rem;
+  font-weight: 600;
+  font-family: 'Red Hat Text', system-ui, -apple-system, sans-serif;
+  letter-spacing: 0.01em;
 }
 
 /* Login card — centered with explicit margin auto */
@@ -87,11 +88,11 @@ html.pf-v5-theme-dark .pf-v5-c-login {
 }
 
 .pf-v5-c-login__main-body {
-  background-color: transparent !important;
+  background-color: transparent;
 }
 
 .pf-v5-c-login__main-header {
-  background-color: transparent !important;
+  background-color: transparent;
 }
 
 /* "Sign in to your account" heading */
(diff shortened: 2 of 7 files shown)


squizzi (commented):
## ACP Review — `squizzi/keycloak-signin-page-styling`

**Reviewed against**: CLAUDE.md conventions, `specs/standards/security/security.spec.md`, ACP review checklist

### Summary

Adds a custom Keycloak login theme for the Kind dev environment using ConfigMap-mounted theme files on the upstream `quay.io/keycloak/keycloak:26.0` image. All changes are scoped to `components/manifests/overlays/kind/` — no production code, no Go/TypeScript, no API changes.

### Checklist Results

| Check | Result | Notes |
|-------|--------|-------|
| No secrets/tokens in diff | ✅ Pass | No sensitive values introduced |
| Volume mounts `readOnly: true` | ✅ Pass | All 4 subPath mounts are read-only |
| ConfigMap `disableNameSuffixHash` | ✅ Pass | Matches existing `keycloak-realm-config` pattern |
| Image reference unchanged | ✅ Pass | Still uses upstream `quay.io/keycloak/keycloak:26.0` |
| Kind overlay scoped | ✅ Pass | No changes to base or other overlays |
| No `panic()` in Go | N/A | No Go code changed |
| No `any` in TypeScript | N/A | No TypeScript code changed |
| OwnerReferences | N/A | ConfigMap is standalone infra, not a controller child resource |
| Feature flag gating | N/A | Dev-only theming, not a user-facing feature |

### Color Palette Verification

All CSS color values match the ambient-ui dark theme palette from `globals.css`:
- `#151515` (background) ✅
- `#1f1f1f` (card) ✅
- `#292929` (secondary/input) ✅
- `#383838` (border) ✅
- `#f2f2f2` (foreground) ✅
- `#a3a3a3` (muted-foreground) ✅
- `#f56e6e` (primary/accent) ✅

### Minor Note

The CSS imports Red Hat Text from Google Fonts via external URL. In air-gapped Kind environments this will fail silently — fonts fall back to `system-ui, -apple-system, sans-serif` gracefully. Not a blocker for dev-only usage.

**Verdict: Approve** — clean, well-scoped change with no convention violations or security concerns.

components/manifests/overlays/kind/keycloak-theme/login.css
```diff
@@ -0,0 +1,273 @@
+/* ACP Keycloak Login Theme — matches ambient-ui dark palette */
+
+/* Import Red Hat Text to match the main UI */
+@import url('https://fonts.googleapis.com/css2?family=Red+Hat+Text:wght@400;500;600;700&display=swap');
```
squizzi, line 4: 🤖 **Amber Analysis**

**External network dependency in a local dev environment.** This `@import` fetches the Red Hat Text font from Google Fonts at runtime. In a Kind cluster (which may be air-gapped, behind a corporate proxy, or simply offline), this request will either fail silently (no font loads, falling back to `system-ui`) or add latency on every login page load.

**Recommended Action:** Either:
1. Bundle the font files as additional ConfigMap entries and use `@font-face` with local paths, or
2. Accept the fallback behavior and add a comment noting the external dependency so future devs aren't surprised when the font doesn't load offline.

Option 2 is fine for Kind dev — just make it explicit.

**Confidence:** High — this is a known pattern issue with container-based dev environments.

components/manifests/overlays/kind/keycloak-theme/login.css
```diff
@@ -0,0 +1,273 @@
...
+  height: 88px !important;
+  background-image: url("../img/acp-logo.svg") !important;
+  background-size: contain !important;
+  background-repeat: no-repeat !important;
+  background-position: center !important;
+  margin-bottom: 0.5rem !important;
+}
+
+.pf-v5-c-login__header::after {
+  content: "Agent Control Plane" !important;
```
squizzi: 🤖 **Amber Analysis**

**Duplicated title string.** "Agent Control Plane" is hardcoded here in the CSS `::after` pseudo-element AND separately in `messages_en.properties` (`loginTitleHtml=Agent Control Plane`). These two sources serve different purposes in Keycloak theming — the messages file controls the `<title>` and `#kc-page-title`, while this CSS pseudo-element creates a separate visual heading.

This means:
- If someone updates the product name in `messages_en.properties`, the CSS heading won't change (and vice versa)
- The page will show the title in two places if Keycloak renders both `#kc-page-title` and this `::after`

**Recommended Action:** Since the `::after` is the primary visual heading and `#kc-header-wrapper` is hidden, consider documenting this coupling with a comment referencing `messages_en.properties`. Or better, use a Freemarker template (`login.ftl`) to render the heading in the DOM so the title comes from one source.

**Confidence:** Medium — the dual-source won't break anything today, but will cause confusion during a rename.

components/manifests/overlays/kind/keycloak-theme/login.css
```diff
@@ -0,0 +1,273 @@
...
+
+/* Header — hide raw text from kc-header-wrapper, use pseudo-elements */
+#kc-header-wrapper {
+  font-size: 0 !important;
+  line-height: 0 !important;
+  color: transparent !important;
+  height: 0 !important;
+  overflow: hidden !important;
+  padding: 0 !important;
+  margin: 0 !important;
```
squizzi, line 41: 🤖 **Amber Analysis**

**Fragile DOM-hiding pattern.** This block hides `#kc-header-wrapper` by zeroing font-size, making text transparent, and collapsing height — rather than using a Freemarker template override. This works today against Keycloak 26.0's PatternFly v5 DOM, but is brittle:

- If Keycloak 26.x or 27.x changes the header wrapper's ID or structure, this selector stops matching
- The `font-size: 0` + `color: transparent` pattern can cause accessibility issues (screen readers may still announce invisible text)

Since this is Kind dev only, the blast radius is small. But consider adding a comment noting which Keycloak version this targets, so future upgrades know to re-verify the login page.

**Confidence:** Medium — not a bug today, but a maintenance concern on Keycloak upgrades.

components/manifests/overlays/kind/keycloak-theme/login.css
```diff
@@ -0,0 +1,273 @@
...
+  align-self: center !important;
+  background-color: #1f1f1f !important;
+  border: 1px solid #383838 !important;
+  border-radius: 0.625rem !important;
+  box-shadow: 0 4px 24px rgba(0, 0, 0, 0.4) !important;
+  grid-area: unset !important;
+}
+
+.pf-v5-c-login__main-body {
+  background-color: transparent !important;
```
squizzi: 🤖 **Amber Analysis**

**Pervasive `!important` overrides.** Nearly every declaration in this file uses `!important`. While I understand the motivation — PatternFly v5 has deep specificity chains — this creates a one-way door: any future theme adjustments also need `!important` or higher specificity, making iteration painful.

For properties that don't compete with PatternFly (e.g., `background-color: transparent` on `.pf-v5-c-login__main-body`, footer text-align), normal cascade would work fine without `!important`.

**Recommendation:** Not blocking, but consider a follow-up pass to remove `!important` from declarations where normal specificity already wins. The PatternFly CSS variable overrides (like `--pf-v5-c-form-control--BorderTopColor`) are actually the right approach — those don't need `!important` when set on the correct selector.

**Confidence:** Low — this is style guidance, not a bug. Acceptable as-is for a dev-only theme.

components/manifests/overlays/kind/keycloak-theme/acp-logo.svg
```diff
@@ -0,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="80" height="80" viewBox="0 0 24 24" fill="none" stroke="#f56e6e" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
```
squizzi: 🤖 **Amber Analysis**

**Minor: SVG has no accessibility metadata.** The SVG logo has no `<title>` element or `aria-label`. Since this SVG is loaded as a CSS `background-image` (via the `::before` pseudo-element), it's inherently invisible to screen readers — `background-image` content is never announced.

This is fine for a dev login page, but if this theme ever moves to a production context, consider either:
- Adding a `<title>ACP Logo</title>` inside the `<svg>` element
- Using an `` tag in a Freemarker template instead of `background-image`, with proper `alt` text

**Confidence:** Low — a11y best practice, not a functional issue for Kind dev.

components/manifests/overlays/kind/keycloak-theme/theme.properties
```diff
@@ -0,0 +1,3 @@
+parent=keycloak.v2
```
squizzi, line 2: 🤖 **Amber Analysis**

**Theme structure looks correct.** `parent=keycloak.v2` inherits the base theme and `styles=css/login.css` correctly references the relative path within the theme's `resources/` directory.

One observation: there's no `login.ftl` template override — the entire customization is CSS-only. This is a valid approach for visual-only changes and avoids the maintenance burden of tracking Freemarker template changes across Keycloak versions. Good tradeoff for a dev-only theme.

Worth documenting somewhere (README or inline comment) that this is intentionally CSS-only to avoid Keycloak upgrade friction.

squizzi (commented):
🤖 **Amber Review — Custom ACP Login Theme for Keycloak**

**Overall Assessment:** Solid dev-environment improvement. The ConfigMap-based theme injection approach is clean — no custom Docker build needed, and the Kustomize integration is well-structured. The visual design matches the ambient-ui dark palette correctly.

**Summary of Findings:**

| Severity | Finding | File |
|----------|---------|------|
| Medium | External Google Fonts dependency may fail in offline/air-gapped Kind clusters | `login.css:4` |
| Low | Title string "Agent Control Plane" duplicated between CSS and `messages_en.properties` | `login.css:66` |
| Low | Fragile DOM-hiding via `font-size: 0` on `#kc-header-wrapper` — ties to Keycloak 26.0 DOM structure | `login.css:31-38` |
| Info | Pervasive `!important` — acceptable for dev theme but limits future iteration | `login.css` throughout |
| Info | SVG logo has no a11y metadata (moot since it's a CSS background-image) | `acp-logo.svg` |
| Info | CSS-only theming (no `.ftl` templates) is the right tradeoff for dev — worth documenting | `theme.properties` |

**What's Good:**
- ConfigMap mount approach avoids custom image builds — clean separation of concerns
- `disableNameSuffixHash: true` prevents Kustomize from breaking the volume mount references
- Volume mounts correctly target Keycloak's expected theme directory structure
- PatternFly CSS variable overrides (not just raw property overrides) are the right pattern
- Scoped to Kind overlay only — no production impact

**Recommended:** Address the Google Fonts external dependency (even just a clarifying comment), then this is good to merge.

**Confidence:** 90% — no blocking issues, all findings are maintenance/resilience concerns appropriate for a dev-only theme.

components/manifests/overlays/kind/keycloak-theme/login.css
squizzi: Fixed in ad5b39c7. Updated the comment to explicitly note the external network dependency and graceful fallback: `/* Red Hat Text from Google Fonts — requires network access; falls back to system-ui offline */`

components/manifests/overlays/kind/keycloak-theme/login.css
squizzi: Fixed in ad5b39c7. Added a comment above the `::after` rule: `/* Title must match messages_en.properties:loginTitleHtml */` — so the coupling is explicit for future renames.

components/manifests/overlays/kind/keycloak-theme/login.css
squizzi: Fixed in ad5b39c7. Updated the comment to call out the Keycloak version: `/* Header — hide raw text from kc-header-wrapper (Keycloak 26.0 / PF5 DOM) */` — so future upgrades know to re-verify.

components/manifests/overlays/kind/keycloak-theme/login.css
squizzi: Fixed in ad5b39c7. Removed `!important` from:
- All CSS custom property overrides (`--pf-v5-*` variables) — these cascade naturally when set on the same selector
- Pseudo-element generated content (`::before`, `::after`) — PF5 doesn't define these
- `background-color: transparent` on containers PF5 doesn't style (`.pf-v5-c-login__main-body`, `.pf-v5-c-login__main-header`, `.pf-v5-c-login__main-footer`)
- `transition` properties — not competing with PF5
- Footer text properties (`text-align`, `color`, `margin-top`)

Kept `!important` on direct property overrides that compete with PF5's specificity chains (background-color, border-color, color on form controls and buttons).

components/manifests/overlays/kind/keycloak-theme/acp-logo.svg
squizzi: Fixed in ad5b39c7. Added `<title>ACP Logo</title>` and `aria-label="ACP Logo"` to the SVG element.

components/manifests/overlays/kind/keycloak-theme/theme.properties
squizzi: Fixed in ad5b39c7. Added a comment to `theme.properties`: `# CSS-only theme (no login.ftl) — avoids Keycloak upgrade friction`

squizzi (commented):
🤖 **Amber Review — No issues found**

Reviewed all 7 changed files covering the CSS-only Keycloak login theme.

**What I verified:**
- Volume mount paths map correctly to Keycloak's theme directory structure (`themes/acp/login/...`)
- CSS relative path `url("../img/acp-logo.svg")` resolves correctly from `resources/css/` → `resources/img/`
- ConfigMap generator keys in `kustomization.yaml` match `subPath` values in the deployment
- `parent=keycloak.v2` is consistent with the PF5 (`pf-v5-*`) CSS selectors throughout `login.css`
- Color palette is internally consistent across all 274 lines of CSS
- `loginTitleHtml` in `messages_en.properties` matches the CSS `content:` pseudo-element
- Font fallback chain (`'Red Hat Text', system-ui, -apple-system, sans-serif`) is present on all `font-family` declarations
- SVG has proper accessibility attributes

**Design notes (not issues):**
- CSS-only approach (no `login.ftl` template override) avoids maintenance burden on Keycloak upgrades — good call
- The `@import` for Google Fonts is render-blocking but the comment documents the tradeoff and system-ui fallback is in place; acceptable for a Kind dev overlay

PR #276, opened 2026-07-07
feat: add git repository payload support to control plane

squizzi:
## Summary

- Add git repo payload delivery to sandboxes via go-git clone + tar-over-SSH in the control plane, supporting network-restricted sandbox pods that cannot clone directly
- Implement ref resolution (branch → tag → commit SHA) with shallow clones for efficiency, `.git` exclusion from tar transfer, and mixed payload dispatch (inline content + repo URL in the same session)
- Fix `failSession` condition format (`status: "False"`, structured `reason`/`message`) so sandbox failures render correctly in the UI, and add `SandboxFailure` to the UI condition titles map
- Add `repo-clone-workspace` example agent with e2e test coverage that polls for clone delivery and verifies LICENSE, go.mod, and .git exclusion in the sandbox

## Test plan

- [ ] `cd components/ambient-control-plane && go vet ./... && go test ./internal/openshell/... ./internal/reconciler/...`
- [ ] `make kind-rebuild` then `make test-gateway-e2e` — verify section 10 (repo payload) passes
- [ ] Manual: create a session with `repo-clone-workspace` agent, verify cloned files at `/sandbox/workspace/`


squizzi (commented):
Reviewed the git repo payload support changes. One finding — goroutine leak in the tar streaming path. Details in inline comment.

components/ambient-control-plane/internal/openshell/ssh_upload.go
```diff
@@ -146,6 +159,194 @@ func writePayloadViaSSH(client *ssh.Client, p Payload) error {
...
+		return "", fmt.Errorf("clone %q (ref=%q): %w", repoURL, ref, err)
+	}
+
+	_ = repo
+	return tmpDir, nil
+}
+
+func isHexSHA(s string) bool {
+	if len(s) != 40 {
+		return false
```
squizzi, line 333: Goroutine leak when the caller doesn't read the pipe to completion. `tarDirectory` returns an `io.Reader` backed by an `io.Pipe`, and the goroutine blocks on `pw.Write()` if the reader side is abandoned. This happens when `writeRepoPayloadViaSSH` fails before or during `io.Copy` (e.g., SSH session start failure at line 310). Since `io.Pipe` has no finalizer-based cleanup, the goroutine, tar writer, and walk closure are leaked permanently. Under persistent SSH failures, these accumulate.

Severity depends on how often the SSH path fails in practice — likely low in healthy clusters, but could matter during gateway outages.

Fix: return an `io.ReadCloser` instead, where `Close()` calls `pr.CloseWithError(errCanceled)` to unblock the writer goroutine. Or accept a `context.Context` and select on cancellation inside the walk callback.

squizzi (commented):
Reviewed the git repo payload support changes (clone via go-git, tar streaming over SSH, `convertPayloads` dispatch, condition format fix, e2e tests). One finding — goroutine leak in the tar streaming error path. Details in inline comment.

jsell-rh (changes requested):
🤖 **Amber Review — PR #276: Git repo payload support**

**Summary:** Solid implementation of the git clone → tar → SSH delivery pipeline. The core approach is sound, test coverage is good, and the `convertPayloads` dispatch logic is clean. A few issues need attention before this is ready to merge.

**Confidence:** High (90%)

---

**Findings:**

**Critical — Temp dir leak on partial clone failure**

In `cloneRepo`, if the final full-clone succeeds but the subsequent `Worktree()` call fails, `os.RemoveAll(tmpDir)` is called and the function returns an error. But the caller `uploadRepoPayload` defers `os.RemoveAll(tmpDir)` on the returned path — which is empty string on error, so `os.RemoveAll("")` is a no-op. This is fine. However, in the branch/tag retry flow, intermediate temp dirs are cleaned up inline but with no guarantee if `MkdirTemp` races with a signal. The bigger issue: if `PlainCloneContext` returns without error but the worktree checkout fails, the tmpDir that was created is cleaned up with `os.RemoveAll(tmpDir)` but the function still returns `("", err)`, leaving `uploadRepoPayload`'s deferred `os.RemoveAll("")` as a no-op. This path is correct — but I'd recommend a single-temp-dir approach using `os.RemoveAll` in a `defer` from the start and re-using the same dir via `PlainInit` + `Fetch` to avoid the retry-create pattern entirely. (Minor: the current approach works but is harder to reason about.)

**Critical — Shell injection in `writeRepoPayloadViaSSH`**

```go
cmd := fmt.Sprintf("mkdir -p '%s' && tar xf - -C '%s'", targetPath, targetPath)
```

`targetPath` is validated by `validatePayloadPath` (which checks `^/[a-zA-Z0-9/_.\\-]+$`), so single-quotes cannot appear in it. However, this regex allows `-` which could collide with tar flags in edge cases. More importantly: the regex check happens inside `writeRepoPayloadViaSSH` but `validatePayloadPath` is also called inside `writePayloadViaSSH` — there are now two code paths and two validation points. If either is ever changed independently, the invariant breaks. Consider centralizing validation in `UploadPayloads` before dispatching.

**Major — No URL validation on `RepoURL` before clone**

`cloneRepo` accepts any URL string and passes it directly to `git.PlainCloneContext`. In a multi-tenant context this means an agent YAML author can supply any URL, including `file:///etc/passwd` (local filesystem reads), `git://` (deprecated, no auth), or a URL pointing to an internal network service. The `Payload.RepoURL` field should be validated at the `convertPayloads` boundary: require `https://` scheme and optionally allowlist known hosts.

```go
case p.RepoURL != "":
    u, err := url.Parse(p.RepoURL)
    if err != nil || u.Scheme != "https" {
        logger.Error().Str("repo_url", p.RepoURL).Msg("payload repo_url must be an https URL; skipping")
        continue
    }
    result = append(result, openshell.Payload{...})
```

**Major — Example agent clones the production repo**

`examples/base/agents/repo-clone-workspace.yaml` clones `https://github.com/openshift-online/agent-control-plane.git` at `ref: main`. This example will be deployed to real clusters and will make a live clone of the production repo every time a session starts with this agent. Use a small, stable public repo (like `https://github.com/octocat/Hello-World.git`) or a test fixture for the example.

**Minor — `_ = repo` is a code smell**

```go
_ = repo
return tmpDir, nil
```

`repo` is only used in the SHA-checkout branch. The blank assignment exists because the branch/tag fallback path assigns `repo` but only uses it conditionally. Refactor to eliminate the blank assignment — move the SHA checkout into a separate function or use a local variable scoped to that block.

**Minor — `defer f.Close()` inside loop**

In `tarDirectory`:
```go
f, err := os.Open(path)
defer f.Close()
```
`defer` inside a goroutine's function body is fine here since the goroutine returns after the walk, but `defer` inside `WalkDir` means the file handles won't be closed until the entire walk completes. For large repos this holds all file descriptors open simultaneously. Use an explicit `f.Close()` after `io.Copy` instead.

**No blockers on:** `convertPayloads` logic (clean), `failSession` condition fix (correct), `SandboxFailure` UI title (correct), test coverage for `TestConvertPayloads` and `TestTarDirectory` (good).

Please address the URL validation (SSRF risk) and example agent repo before merge.

— Amber

jsell-rh:
🤖 **Amber** — adding inline review comments shortly via the review above. Two items need resolution before this can be approved: (1) URL validation on `RepoURL` to prevent SSRF, (2) example agent should not clone the production repo. See review for full details.

jsell-rh:
🤖 **Amber** — Inline review comments noted above. Two items need resolution before this can be approved:\n\n1. **URL validation on `RepoURL`** (SSRF risk — `file://`, `git://`, internal endpoints allowed today)\n2. **Example agent clones production repo** — use a small stable test fixture instead\n\nSee the full review for details. Minor items (`_ = repo`, `defer` inside loop) can be addressed opportunistically.\n\n— Amber

jsell-rh:
Duplicate comment — disregard, full review posted above. — Amber

jsell-rh:
_(Deleting duplicate comment above — review was posted twice due to tooling. The full review is the formal review submitted via GitHub review UI, not these inline comments.)_\n\n— Amber

jsell-rh:
🤖 **Amber** — Labeling as `amber/change-requested`. Two blockers:\n1. **SSRF risk**: `RepoURL` accepts `file://`, `git://`, internal endpoints — add `https://`-only validation in `convertPayloads`\n2. **Example agent**: `repo-clone-workspace.yaml` clones the production ACP repo on every session start — use a small test fixture (e.g. `octocat/Hello-World`) instead\n\nMinor: `_ = repo` blank assignment, `defer f.Close()` inside WalkDir loop (file handle accumulation on large repos)\n\n— Amber

jsell-rh:
_(Ignore prior duplicate comment fragments — review is complete. Labeling PR now.)_

jsell-rh (approved):
🤖 **Amber Review — Approved**

The new commits fully address my prior concerns and the goroutine leak Squizzi self-identified. Good work closing all three in one pass.

**What changed:**
- `tarDirectory` → returns `io.ReadCloser` (pr.CloseWithError unblocks the goroutine on SSH failure — goroutine leak fixed ✅)
- `validateRepoURL()` enforces `https://` scheme — SSRF via `file://`/`git://` blocked ✅
- `copyFileToTar()` extracted — defer accumulation in WalkDir loop eliminated ✅
- Example agent switched to `octocat/Hello-World` ✅
- 9-case `TestValidateRepoURL` test coverage ✅

**ACP Checklist:**
- ✅ No tokens in logs (repo URLs are logged, which is fine)
- ✅ `convertPayloads()` handles content/repo_url mutual exclusivity with warning
- ✅ 5-minute context timeout for repo clone path
- ✅ OwnerRef pattern unaffected
- ✅ `go vet` passes (implied by test suite)

**One remaining minor nit** (non-blocking, follow-up fine):
`cloneRepo()` silently falls through branch → tag → SHA on any error from the first attempt, not just "ref not found". A network timeout on the branch clone will trigger a SHA retry against the same unreachable server. Consider checking for `plumbing.ErrReferenceNotFound` specifically before retrying. Low priority — the worst case is two extra failed clone attempts, not a correctness issue.

Confidence: High. Ship it.

— Amber

squizzi:
Significant implementation change, removing `amber/approved` label

jsell-rh (changes requested):
🤖 **Amber Review — PR#276: Add git repository payload support to control plane**

The overall approach is sound — cloning in-memory and streaming via tar-over-SSH is clever for network-restricted sandbox pods. The `validateRepoURL` scheme check, `.git` directory exclusion, and `validatePayloadPath` regex are good defensive touches.

However, there are two security issues that need to be resolved before this merges:

**Blocker — Unbounded in-memory clone (DoS risk):**
The control plane clones entire git repositories into heap memory (`memory.NewStorage()`). A session pointing at a multi-GB repo (e.g. a Linux kernel mirror, a large monorepo) will consume proportional heap on the control plane pod — potentially OOM-killing it and taking down all session reconciliation for the namespace. The SHA fallback path (full clone, depth=0) is especially dangerous.

Suggested mitigations (pick one or combine):
1. Cap clone depth/size: `CloneOptions.Tags = git.NoTags`, set a bytes-transferred limit, or use `git archive` via the SSH gateway instead of in-process cloning
2. Enforce a max compressed repo size before cloning (check `Content-Length` on the smart HTTP `/info/refs` response)
3. Stream the tar directly from a shallow clone without materialising the whole FS in memory (use `git.CloneContext` with a bounded in-memory store)

**Major — No URL allowlist (SSRF vector):**
Any authenticated user can create a session whose repo payload points at any HTTPS URL — including internal cluster services (e.g. `https://kubernetes.default.svc`), cloud metadata endpoints, or private artifact registries. The control plane makes the HTTP request, not the sandbox, so this bypasses sandbox network policies entirely.

Suggested mitigations:
1. Add a configurable `REPO_URL_ALLOWLIST` env var (comma-separated host patterns) defaulting to deny-all or a known-safe list
2. At minimum, block RFC1918 addresses and well-known cloud metadata endpoints (`169.254.169.254`, etc.) in `validateRepoURL`

**Critical — Context not propagated through `tarFilesystem` goroutine:**
The goroutine launched in `tarFilesystem` has no awareness of the parent context. If the session context is cancelled (timeout, pod restart), the goroutine continues reading the billy filesystem and writing to the pipe until it finishes, leaking the goroutine. Pass ctx through and check `ctx.Done()` in the walk loop.

**Minor — Retry loop loses intermediate errors:**
When branch → tag → SHA fallback occurs, each failed attempt overwrites `err`. If the final clone also fails, the returned error only describes the last attempt, making debugging hard. Consider accumulating: `var errs []string` and joining them.

**Minor — `OPENSHELL_USE_GATEWAY=true` in all CI runs:**
This is a meaningful blast radius increase — if the gateway pod isn't ready within the wait window, all CI runs fail. Confirm the gateway setup is reliable enough before making this the default.

Confidence: **High (90%)** on the security findings, **Medium (80%)** on the context-leak (depends on billy Walk implementation details).

*— Amber*

components/ambient-control-plane/internal/openshell/ssh_upload.go
```diff
@@ -146,6 +163,187 @@ func writePayloadViaSSH(client *ssh.Client, p Payload) error {
 	return nil
 }
 
+func validateRepoURL(repoURL string) error {
+	parsed, err := url.Parse(repoURL)
+	if err != nil {
+		return fmt.Errorf("invalid repo URL: %w", err)
+	}
+	if parsed.Scheme != "https" {
```
jsell-rh, line 191: 🤖 **Amber Analysis — SSRF: no block on internal cluster addresses**

`validateRepoURL` checks scheme and host presence but allows any HTTPS URL including:
- `https://kubernetes.default.svc` (K8s API server)
- `https://169.254.169.254` (cloud metadata endpoint)
- `https://10.x.x.x` (RFC1918 services)

The control plane makes the HTTP request (not the sandbox), so sandbox network policies don't protect against this. A tenant can use this to probe internal cluster services.

Add a blocklist for known-dangerous targets:
```go
func isPrivateHost(host string) bool {
    // block RFC1918, link-local, metadata endpoints
    blocked := []string{"169.254.", "10.", "172.16.", "192.168.", "kubernetes.default"}
    for _, b := range blocked {
        if strings.HasPrefix(host, b) || strings.Contains(host, b) {
            return true
        }
    }
    return false
}
```

Or better: resolve the hostname and reject if it maps to a private IP.

**Confidence:** High (90%) — standard SSRF pattern for any feature that fetches user-supplied URLs from a privileged process.

*— Amber*

components/ambient-control-plane/internal/openshell/ssh_upload.go
```diff
@@ -146,6 +163,187 @@ func writePayloadViaSSH(client *ssh.Client, p Payload) error {
...
+func isHexSHA(s string) bool {
+	if len(s) != 40 {
+		return false
+	}
+	for _, c := range s {
+		if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F')) {
+			return false
+		}
+	}
+	return true
```
jsell-rh, line 340: 🤖 **Amber Analysis — Context cancellation not respected in goroutine**

The goroutine launched here runs to completion regardless of context cancellation. If the session times out or the reconciler restarts, this goroutine continues walking the in-memory filesystem and writing to the pipe, leaking goroutine + memory until the walk finishes naturally.

Fix by checking `ctx.Done()` in the walk callback:
```go
err := billyutil.Walk(bfs, "/", func(path string, info os.FileInfo, walkErr error) error {
    select {
    case <-ctx.Done():
        return ctx.Err()
    default:
    }
    // ... rest of walk
})
```

`tarFilesystem` signature needs to accept `ctx context.Context` to enable this.

**Confidence:** Medium (80%) — goroutine leak is certain; severity depends on how often sessions are cancelled mid-clone.

*— Amber*

jsell-rh (commented):
Inline security findings to accompany the changes-requested review above.

jsell-rh (changes requested):
🤖 **Amber Re-Review — PR#276: Git repository payload support (commit f8165560)**

Good progress on commit f8165560 — two of the three issues from my last review are fully resolved. One blocker remains.

---

**✅ Fixed: SSRF protection**

`validateRepoURL` now resolves hostname DNS and blocks RFC1918, link-local, loopback, k8s.default, and GCP metadata endpoints. The test suite covers 16 cases including `127.0.0.1`, `169.254.169.254`, `10.x`, `172.16.x`, `192.168.x`, and the well-known service names. Solid.

**✅ Fixed: Context propagation through `tarFilesystem`**

The goroutine now checks `ctx.Done()` inside the walk loop and `TestTarFilesystemCancellation` verifies it actually errors. Good.

---

**🔴 Remaining Blocker: Unbounded in-memory full clone on SHA fallback**

In `cloneRepoFS`, the SHA fallback path (branch → tag → SHA) still does:

```go
cloneOpts.Depth = 0
repo, err = git.CloneContext(ctx, memory.NewStorage(), wt, cloneOpts)
```

`Depth: 0` is an unbounded full clone into heap memory. Any authenticated session user can point `repo_url` at a large public repo (Linux kernel, Chromium, etc.) and OOM the control plane pod. Because the control plane is the single reconciler for all sessions in the namespace, an OOM kill takes down everyone — multi-tenant blast radius.

The 5-minute context timeout provides a time bound but not a memory bound. A 4GB repo heap allocation on a 2Gi control plane pod will kill the pod before the timeout fires.

**Suggested fix (one-liner) — block SHA refs for now:**
```go
if isHexSHA(ref) {
    return nil, fmt.Errorf("direct SHA refs are not supported; use a branch or tag name")
}
```

SHA checkout at arbitrary commits is an edge case — ship it as a follow-up once there's a safe bounded implementation. This lets the branch/tag paths (always `Depth: 1`) merge safely now.

---

**⚠️ Advisory: DNS rebinding (non-blocking)**

`validateRepoURL` resolves DNS at validation time; the actual clone happens after. DNS rebinding (flip DNS to private IP after validation) can bypass the SSRF check. This is a known limitation of pre-clone validation and acceptable given attack complexity — worth a comment in the code acknowledging it.

---

**Summary:**

| Issue | Status |
|-------|--------|
| SSRF via internal IP/hostname | ✅ Fixed |
| Context cancellation in `tarFilesystem` | ✅ Fixed |
| Unbounded SHA full clone (OOM/DoS) | 🔴 **Blocker — still open** |
| DNS rebinding | ⚠️ Advisory |

Fix is a one-liner — happy to approve immediately after.

*— Amber*

jsell-rh (approved):
🤖 **Amber Review — PR#276 Re-check: All feedback addressed + significant hardening added**

All three items from the prior review are resolved, and the author went further with proactive hardening:

| Finding | Status |
|---|---|
| Goroutine leak in `tarDirectory` on SSH failure | ✅ Returns `io.ReadCloser`; caller calls `tarReader.Close()` which unblocks the writer goroutine via `pr.CloseWithError` |
| SSRF: no block on internal cluster addresses | ✅ `validateRepoURL()` enforces `https://` scheme, blocks `kubernetes.default` / `metadata.google.internal` by name, resolves hostname DNS and rejects RFC1918 / link-local / loopback / IPv6 private ranges |
| Context cancellation not respected in goroutine | ✅ `tarFilesystem` goroutine now checks `ctx.Done()` inside the walk callback; test coverage added (`TestTarFilesystemContextCancellation`) |

**Additional hardening beyond the review (all good):**
- `limitedFS` wraps `memfs` with a 1 GiB cumulative write limit, preventing OOM-kill of the control plane from large repos. `limitedFile` tracks bytes across `Create` and `OpenFile` paths. Tests cover under-limit, over-limit, and cumulative-across-files cases.
- SHA ref full-clone path (`Depth: 0`) is now bounded by `limitedFS`, so re-enabling it is safe.
- DNS rebinding risk is honestly documented as a comment — appropriate transparency.

**One minor note (non-blocking):** `limitedFS.OpenFile` wraps all open calls including read-only opens into `limitedFile`, meaning read `Write()` calls on an `O_RDONLY` file will still hit the counter. In practice this is harmless (reads don't call `Write`), but the counter tracks written bytes only if callers only write through `limitedFile`. The test suite validates the meaningful paths.

Approved to merge when ready.

*— Amber*

jsell-rh:
🤖 **Amber — Re-check after new commits (e8f99c8, 285f376, 9a785f2)**

Checked the three commits pushed since my prior approval. Summary:

- **e8f99c8** — merge from `main` (no logic change)
- **285f376** — conflict resolution for `kube_reconciler_test.go` and `skills/RECONCILE.md` (no logic change)
- **9a785f2** — `ruff`/`black` reformatting only: line-length wraps in Python runner files (`app.py`, `mlflow_autolog.py`, `mlflow_observability.py`), parenthesized context managers in tests, removal of unused import (`_INITIAL_PROMPT_FILE`), trailing whitespace removed from spec markdown — no logic changes

**No new concerns.** Prior approval stands.

*— Amber*

jsell-rh (approved):
🤖 **Amber Re-check — Post-approval commits reviewed**

Three commits landed after my Jul 7 22:36Z approval (`e8f99c8`, `285f376`, `9a785f2`). Reviewed:

- **Merge from `main`** (`e8f99c8`): Brings in mlflow runner formatting changes and new reconciler test cases (`TestConvertPayloads`, `TestConvertPayloads_RepoFieldsPreserved`, `TestPayloadStructFields`). The runner changes are pure `black` reformatting — no logic delta. The new tests are additive and good.
- **Conflict resolution** (`285f376`): `RECONCILE.md` and `kube_reconciler_test.go` merge conflict resolved. No logic changes, comment-only cleanup in test file.
- **Lint fixes** (`9a785f2`): Pre-push hook formatting pass. Clean.

**Approval stands.** No new issues introduced. Ready to merge when CI is green.

*— Amber*

jsell-rh (approved):
🤖 **Amber Re-check — Post-approval commit reviewed (497228ca)**

One commit landed after my Jul 8 08:33Z approval:

**`fix(control-plane): substitute NAMESPACE_PLACEHOLDER in gateway DNS names before validation`**

The fix is correct. `NAMESPACE_PLACEHOLDER` contains uppercase letters which fail RFC 1123 validation, causing the cascade described in the commit message (gateway never deploys → certgen skipped → TLS secrets missing → sandboxes deleted). Resolving the placeholder before building `GatewayConfig` is the right fix and matches the existing substitution pattern in `ApplyManifestToNamespace`.

- ✅ Fix is minimal and targeted (`strings.ReplaceAll` on `ServerDnsNames` before `GatewayConfig` construction)
- ✅ Test covers the substitution and validates the resolved name passes `gateway.ValidateDNSName`
- ✅ No new `panic()`, no tokens in logs, no API schema changes

**One minor nit (non-blocking):** `TestReconcileGateway_NamespacePlaceholderSubstitution` re-implements the substitution logic inline rather than calling the production function under test. This means if `reconcileGateway` changes how it resolves placeholders, the test won't catch it. A follow-up to extract the substitution into a pure helper function and test that directly would be worthwhile.

Approval stands. Ready to merge when CI is green.

*— Amber*

jsell-rh (commented):
🤖 **Amber Re-check — New commit `497228ca` (2026-07-08T15:55Z)**

**`fix(control-plane): substitute NAMESPACE_PLACEHOLDER in gateway DNS names before validation`**

One new commit since my Jul 8 07:05Z approval. Reviewed.

**What changed:** `gateway_reconciler.go` now resolves `NAMESPACE_PLACEHOLDER` in `ServerDnsNames` entries before constructing `GatewayConfig`. The substitution uses `gw.Name` (the gateway resource name, e.g. `"tenant-a"`) — this matches the intent expressed in the commit message and mirrors the existing substitution pattern in `ApplyManifestToNamespace`.

**Is the substitution key correct?**

`NAMESPACE_PLACEHOLDER` is replaced with `gw.Name` here, but in `ApplyManifestToNamespace` it's replaced with the *namespace* (project ID). These are semantically different: gateway name ≠ project namespace. If the DNS template is `openshell-gateway.NAMESPACE_PLACEHOLDER.svc.cluster.local`, the correct value should be the K8s *namespace* the gateway lives in, not the gateway's own name. If `gw.Name == project.ID` in practice (i.e. the gateway name is the same as the namespace), this is fine — but that coupling should be explicit. See the test: it uses `"tenant-a"` as the gateway name and expects `"openshell-gateway.tenant-a.svc.cluster.local"`, which implies gateway name ≈ namespace. This is worth a one-line comment in the code confirming the assumption.

**Test coverage:** `TestReconcileGateway_NamespacePlaceholderSubstitution` added — re-implements the substitution inline rather than calling the reconciler method directly (so it's testing the logic, not the call site), but for a simple `strings.ReplaceAll` this is acceptable. The test also calls `gateway.ValidateDNSName(dns)` to confirm the resolved name passes validation — good.

**Nit (non-blocking):** The test re-implements the substitution logic inline rather than calling a shared helper. If `reconcileGateway` is ever refactored to extract the substitution, the test won't catch regressions. Consider extracting `resolveGatewayDNSNames(names []string, gatewayName string) []string` as a pure, tested function — but this can be a follow-up.

**Overall:** The fix is correct, surgical, and addresses the real production regression (namespace validation cascade). No new concerns introduced.

**My prior approval stands.** No blocker from this commit.

*— Amber*

jsell-rh:
🤖 **Amber — Follow-up (commit `9e3aa11`)**

Reviewed `fix(control-plane): deploy gateways to project namespace instead of gateway name`.

This directly resolves the concern I flagged in my 17:35Z comment: `NAMESPACE_PLACEHOLDER` was being replaced with `gw.Name` (the gateway resource name) rather than the project's actual K8s namespace. The fix threads `projectID` from `reconcileProjectGateways` down into `reconcileGateway` as an explicit `namespace` parameter, then uses it for both the DNS name substitution and `NamespaceConfig.Name`.

- ✅ Both substitution sites updated (`resolvedDnsNames` and `NamespaceConfig.Name`)
- ✅ The gateway name / namespace coupling ambiguity is resolved — `namespace` is now explicit
- ✅ Matches the same fix applied on PR #279 (`eda843fa`) — consistent approach across both PRs

Approval stands. No new concerns.

— *Amber*

jsell-rh (commented):
🤖 **Amber Re-check — PR#276 — New commits since last review (2026-07-08T16:04Z)**

Two new commits since my last approval at `497228ca`:

| Commit | Summary |
|--------|---------|
| `7704460` | Merge `origin/main` — no logic changes |
| `9e3aa11` | **Fix: deploy gateways to project namespace instead of gateway name** |

---

### `9e3aa11` — Gateway namespace fix reviewed

This directly addresses the nit I flagged in my prior re-check: *"`NAMESPACE_PLACEHOLDER` is replaced with `gw.Name` here, but in `ApplyManifestToNamespace` it's replaced with the namespace (project ID). These are semantically different."*

The fix is correct and minimal:

- ✅ `reconcileProjectGateways` now passes `projectID` into `reconcileGateway` as the `namespace` parameter
- ✅ `NAMESPACE_PLACEHOLDER` resolves to `namespace` (= projectID, e.g. `"tenant-a"`) instead of `gw.Name` (e.g. `"openshell-gateway"`) — correct semantics
- ✅ `nsConfig.Name` is set to `namespace` — gateway resources now deploy into the actual project namespace
- ✅ 8-line surgical fix; no behavioral changes to other code paths
- ✅ This was a real production bug: gateways targeting a non-existent namespace (named after the gateway resource) caused "namespace not found" cascade failures

**Prior approval stands.** No new concerns. Ready to merge when CI is green.

*— Amber*

jsell-rh (approved):
🤖 **Amber Re-check — PR#276 — 2 new commits since last approval (2026-07-08T20:37Z)**

Two commits landed after my prior approval at `9e3aa11`:

| Commit | Summary |
|--------|---------|
| `ae251ae` | `fix(e2e): poll for session Running phase before checking payloads` |
| `53e0f4f` | `fix(control-plane): validate RSA PEM keys and make credential rotation non-fatal` |

---

### `ae251ae` — E2E polling fix ✅

Solid improvement:
- ✅ Replaces fixed `sleep 5` with a 60s polling loop (30 × 2s) for session `Running` phase — correct understanding of the CP lifecycle (payload upload is gated on sandbox READY + DNS verification + phase transition)
- ✅ Second poll (10 × 2s for file payload, 15 × 2s for repo payload) correctly decouples "session is running" from "file is present" — necessary since the upload happens just before exec starts
- ✅ Upgrades `skip` to `fail` for missing `repo-clone-workspace` agent and adds `acpctl apply` fallback — eliminates silent test gaps
- ✅ Adds agent to base `kustomization.yaml` — ensures it deploys with standard `make deploy`
- ✅ `${PAYLOAD_CONTENT:-}` and `${PHASE:-unknown}` are correct bash safety patterns for unset variables

No issues found.

---

### `53e0f4f` — RSA PEM validation + non-fatal rotation ✅

**`validateRSAPEM()`:**
- ✅ Handles both PKCS1 (`RSA PRIVATE KEY`) and PKCS8 (`PRIVATE KEY`) PEM formats — covers all real-world Vertex SA key formats
- ✅ `pem.Decode` → `ParsePKCS1PrivateKey` → `ParsePKCS8PrivateKey` → RSA type assertion — correct order, correct error propagation
- ✅ 6-case test suite covers PKCS1 valid, PKCS8 valid, truncated PEM, no PEM block, missing fields, invalid JSON — meaningful coverage

**Non-fatal credential rotation (`ensureVertexCredentialRefresh`):**
- ✅ Correct trade-off: returning `nil` on initial rotation failure lets CI with dummy credentials proceed; the `Warn` log with `provider` + `credential_key` fields gives operators enough context when real credentials fail

**CI + pre-commit:**
- ✅ `openssl genpkey -algorithm RSA -pkeyopt rsa_keygen_bits:2048` generates a properly-structured 2048-bit key — eliminates the truncated-stub root cause
- ✅ `provider_mapping_test.go` excluded from `detect-private-key` hook — correct, the file generates keys at runtime via `crypto/rand`

**One nit (non-blocking):** `contains()` in the test file re-implements `strings.Contains`. Safe to clean up in a follow-up.

---

**Summary:** Both commits are solid. The e2e polling fix correctly models the CP lifecycle, and the RSA validation is the right defense-in-depth for a security-critical path. **Approval stands.**

*— Amber*

jsell-rh:
🤖 **Amber — Re-check after new commits (`ae251ae`, `53e0f4f`)**

Two commits landed after my 19:04Z approval:

**`ae251ae`** — E2E polling fix: replaces fixed `sleep 5`/`sleep 3` with phase-polling loops (up to 30 × 2s). Correctly breaks on `Running`, `Succeeded`, or `Failed` so the test exits promptly even when sessions fail. `repo-clone-workspace` added to base kustomization and applied via `acpctl` if missing; missing agent is now a hard `fail` instead of a silent skip. ✅

**`53e0f4f`** — RSA PEM validation + non-fatal credential rotation:
- `validateRSAPEM` tries PKCS1 then PKCS8 and type-asserts the PKCS8 result — correct approach, handles both formats. ✅
- CI workflow now generates valid 2048-bit keys via `openssl genpkey` instead of a truncated stub. ✅
- Making `RotateProviderCredential` non-fatal is the right call for CI parity; the `Warn` log with provider/credential_key context gives enough signal for debugging. ✅
- `provider_mapping_test.go` excluded from `detect-private-key` hook — keys are runtime-generated, not committed. ✅

One minor note: `contains`/`searchSubstring` in the test file reimplements `strings.Contains` — the stdlib function would do the same job. Low priority, not a blocker.

**No new concerns.** Approval stands.

— *Amber*

jsell-rh:
🤖 **Amber — Re-check after new commit (`ceb0053`)**

`fix(ci): exclude vteam-catalog-lab-e2e workflow from detect-private-key hook` — 1-line diff adding `.github/workflows/vteam-catalog-lab-e2e.yml` to the `detect-private-key` exclude list in `.pre-commit-config.yaml`. Follows the identical pattern already in place for `test-local-dev.yml` and `provider_mapping_test.go`, and the key in question is a truncated, non-functional CI fixture (documented in the commit message). ✅

**No new concerns.** Approval stands.

— *Amber*

PR #278, opened 2026-07-07
feat: pretty hcmai E2E smoke test with live event streaming

markturansky:
## Summary

- Rewrites `hcmai-smoke.sh` from a thin wrapper into a self-contained pretty E2E demo script
- Orange + white color scheme with section banners, step headers showing exact `acpctl` commands
- Uses `acpctl session messages -F` (continuous follow) to live-stream AG-UI events in the background
- Polls `events-history --event-type RUN_FINISHED` to detect turn completion
- No cleanup — preserves all resources with replay instructions at the end
- Shows full session describe + event history at the end
- Masks OIDC client secret from command output

## Test plan

- [x] `./tests/e2e/hcmai-smoke.sh` — 18/18 passed on hcmai
- [x] Resources preserved and inspectable after run
- [x] No secrets leaked in terminal output

### DIFF
--- a/tests/e2e/hcmai-smoke.sh
+++ b/tests/e2e/hcmai-smoke.sh
@@ -1,67 +1,621 @@
 #!/usr/bin/env bash
-# hcmai-smoke.sh — E2E smoke test against the hcmai OpenShift cluster
+# hcmai-smoke.sh — Pretty E2E smoke test against the hcmai OpenShift cluster
 #
-# Wraps smoke-test-llm.sh with hcmai-specific OIDC + Vertex credentials.
-# Reads secrets from the ambient-code-gitops repo and the hcmai Keycloak
-# OIDC client embedded in the ambient-api namespace.
+# Validates the full golden path with colorful output:
+#   Login → Project → Provider → Credential → Agent → Session → LLM response
+#
+# Resources are KEPT after the run so you can inspect them.
+# Uses `acpctl session messages -f` to live-stream the LLM event stream.
 #
 # Usage:
-#   ./tests/e2e/hcmai-smoke.sh                    # full run with cleanup
-#   SKIP_CLEANUP=1 ./tests/e2e/hcmai-smoke.sh     # preserve resources
+#   ./tests/e2e/hcmai-smoke.sh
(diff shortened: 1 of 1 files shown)


jsell-rh (commented):
🤖 **Amber Review — PR #278**

**Summary:** Solid E2E smoke test rewrite. Test-only change, no production code impact. CI passing. A few minor points below.

**Confidence:** High (95%) — straightforward bash script, well-structured, good secrets hygiene.

---

### What's good

- Secrets hygiene is correct: `OIDC_CLIENT_SECRET` never logged, `VERTEX_SA_KEY` passed only via env var, client secret masked from terminal output
- `set -euo pipefail` present — good
- `RUN_ID`-suffixed resource names prevent test run collisions
- Background stream + explicit `kill`/`wait` pattern for the event stream is clean
- Token refresh logic in `refresh_oidc_token()` + `ensure_token()` is a solid pattern for long-running tests

### Minor issues

**1. Branch name mismatch**
Head branch is `fix/informer-initial-sync-deadlock` — unrelated to smoke test prettification. Not a blocker but creates confusing git history. Consider squashing onto a properly-named branch before merge.

**2. `json_field` shell-interpolates into Python**
```bash
json_field() {
  local json="$1" field="$2"
  echo "$json" | python3 -c "
...
    print(json.loads(raw[start:]).get('${field}',''))
..."
}
```
The `${field}` is shell-expanded inside the Python string. All callers use hardcoded strings (`"id"`, `"phase"`) so there's no injection risk here in practice, but it's a fragile pattern. A safer alternative is passing `field` as a positional arg to the Python invocation: `python3 -c "import sys,json; print(json.load(sys.stdin).get(sys.argv[1],''))" "$field"`.

**3. `curl -sk` disables TLS verification**
Intentional for the hcmai internal cluster, but worth a comment in the `api()` function so future readers know it's deliberate rather than accidental.

**4. No cleanup — accumulated resources**
Explicitly documented and intentional. Just a reminder that repeated runs on hcmai will accumulate `smoke-llm-*` projects and `vertex-smoke-*` credentials. The `USE_EXISTING_PROJECT` flag handles re-use but doesn't address stale credential growth. Not a blocker.

---

None of these are blockers for a test-only file. Approving with the branch name note as the only thing I'd like addressed before final merge.

— *Amber*

PR #285, opened 2026-07-08
fix(control-plane): verify sandbox ndots after ready instead of racing controller

bsquizz:
## Summary

- Fixes race condition where the agent-sandbox controller could create a pod with `ndots:5` before our CR patch landed, breaking inference routing via `inference.local`
- Instead of eagerly deleting the pod after patching the CR, now verifies the pod's actual `/etc/resolv.conf` after it reaches Ready
- If `ndots:5` is found, deletes the pod so the controller recreates it from the already-patched CR (up to 3 retries)

## Test plan

- [ ] `cd components/ambient-control-plane && go build ./...` passes
- [ ] Deploy to kind cluster with cold image pull (clear containerd cache)
- [ ] Create a session and verify CP logs show `"verified sandbox DNS config: ndots is correct"`
- [ ] Verify sandbox pod has `ndots:1` in `/etc/resolv.conf`
- [ ] If race is hit, verify logs show `"sandbox pod has ndots:5, deleting for recreation"` followed by successful recreation

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -42,8 +42,8 @@ func validateTSLValue(value string) error {
 }
 
 const (
-	mcpSidecarPort   = int64(8090)
-	mcpSidecarURL    = "http://localhost:8090"
+	mcpSidecarPort    = int64(8090)
+	mcpSidecarURL     = "http://localhost:8090"
 	initialPromptPath = "/tmp/initial_prompt.txt"
 )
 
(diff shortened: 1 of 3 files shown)


jsell-rh (changes requested):
## 🤖 Amber Analysis — PR #285

Good fix for a real race condition. The approach (verify after ready, delete+retry instead of eager-delete) is sound. Three issues need addressing before this leaves draft.

**Confidence: 90%** — Findings are based on code reading; the retry loop analysis is high confidence, the context concern is medium.

<details>
<summary>Details</summary>

### Critical: exec errors skip retry counter, allowing infinite loops

In `execAfterReady` (kube_reconciler.go ~L681):

```go
dnsOK, dnsErr := r.verifyAndFixDNSConfig(namespace, sandboxID, sbxName)
if dnsErr != nil {
    r.logger.Warn()...
    continue   // ← ndotsRetries NOT incremented
}
if !dnsOK {
    ndotsRetries++
    if ndotsRetries > maxNdotsRetries { ... fail ... }
    continue
}
```

If `ExecSandbox` consistently returns an error (e.g., gateway unreachable, pod not accepting exec), `dnsErr != nil` fires every tick, `ndotsRetries` stays 0 forever, and the session never fails. The pod is also never deleted. This loop will run until the outer session timeout (if one exists) or indefinitely.

**Fix:** increment `ndotsRetries` on exec errors too, or add a separate exec-error counter with its own threshold.

### Major: `verifyAndFixDNSConfig` creates a detached context

```go
func (r *SimpleKubeReconciler) verifyAndFixDNSConfig(namespace, sandboxID, sbxName string) (bool, error) {
    ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```

This creates a brand-new background context. If the session is cancelled (user abort, shutdown signal) while exec is in flight, the cancellation won't propagate into this function. Pass the caller's `ctx` in, with the 30s timeout applied on top:

```go
func (r *SimpleKubeReconciler) verifyAndFixDNSConfig(ctx context.Context, namespace, sandboxID, sbxName string) (bool, error) {
    ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```

### Minor: missing test coverage for new logic (Principle IV)

`verifyAndFixDNSConfig` contains the core logic of this fix — the ndots check, the delete path — and has zero test coverage. The mock interface stub added to `provider_reconciler_test.go` (the `FetchSandboxLogs` addition) is just interface compliance, not a test.

A table-driven test covering at minimum:
- `ndots:1` in resolv.conf → returns `(true, nil)`
- `ndots:5` in resolv.conf → deletes pod, returns `(false, nil)`
- exec error → returns `(false, err)`

would validate the retry logic and give reviewers confidence in the fix.

</details>

**Rollback:** `git revert` this commit and redeploy — the eager-delete approach from before this PR will return, but the race condition returns with it.

— Amber

jsell-rh (approved):
🤖 **Amber Review — PR #285 (Updated)**

The four commits since initial labeling fully address the concerns I'd flag on this fix. The race condition root cause is correctly identified and the solution is sound.

**What changed and why it's correct:**

The original approach (`patch CR → immediately delete pod`) had an unavoidable race against the agent-sandbox controller. The new approach (`patch CR → wait for READY → exec-verify resolv.conf → delete only if ndots:5`) eliminates the race by treating the pod's own state as the source of truth.

**Logic review:**
- `verifyAndFixDNSConfig` correctly derives its 30s timeout from the parent `pollCtx` (not a detached `context.Background()` — fixed in commit 3).
- `awaitingPodRestart` state machine correctly waits for the pod to leave READY before re-polling, preventing exec into a terminating pod.
- `ndotsRetries` is incremented in *both* the error and the `!dnsOK` paths — the infinite-loop fix from commit 3 is correct.

**Tests:**
Table-driven coverage for correct ndots, ndots:5 → delete, exec error, non-zero exit, and ndots:2 (not ndots:5). Solid.

**One minor nit (non-blocking):**
```go
const maxNdotsRetries = 3
if ndotsRetries > maxNdotsRetries {  // allows 4 attempts, named "max 3"
```
`>` allows one more attempt than the name implies. `>=` would be strictly correct. Fine to address or leave as-is — the behavior is safe either way.

**CI:** CodeQL ✅, Lint ✅, Unit Tests ✅, Build ✅, Konflux on-pull-request ✅. One test job `test-local-dev-simulation` is still in progress but the rest are green.

Removing `amber/change-requested` — this is ready for human review once the author marks it ready for review.

— *Amber*

PR #286, opened 2026-07-08
feat(events-api): API-driven Gateway reconciliation (Waves 10-13)

markturansky:
## Summary

Implements the full Gateway-as-API-Resource stack from PR #281's spec rewrite, replacing ConfigMap-based gateway provisioning with API-driven reconciliation across 4 waves:

- **Wave 10 (P11)**: Gateway API resource — OpenAPI spec, API server plugin (CRUD, migration, RBAC), Go/Python/TS SDK codegen
- **Wave 11 (P13+P15)**: Shared kustomize library (`go-sdk/kustomize`) for manifest loading + `acpctl apply` refactored to use shared library with Gateway kind support
- **Wave 12 (P12+P14+P16+P17)**: `GatewayReconciler` polling pattern (30s ticker, lists projects → gateways via SDK), ConfigMap-based provisioning eliminated, type definitions preserved
- **Wave 13 (P18+P19+P20+P21)**: Gateway example manifests, annotation-based failure tracking (`ambient.ai/reconcile-status`), platform-config.yaml removal from overlays

### Key changes

- New `GatewayReconciler` in control plane follows `ApplicationReconciler` polling pattern — SDK REST client lists all projects, then gateways per project
- ConfigMap functions (`LoadPlatformConfig`, `WatchPlatformConfig`, `initGatewayProvisioning`) removed; `NamespaceConfig`/`GatewayConfig` types retained for `ReconcileGateways`
- Gateway validation/reconcile failures reported via annotations (no status fields on Gateway type)
- Pre-existing `mockGateway` missing `FetchSandboxLogs` method fixed
- gofmt applied to 4 files with pre-existing formatting issues

## Test plan

- [x] `go build ./...` — clean (control plane)
- [x] `go vet ./...` — clean (control plane)
- [x] `gofmt -l` — clean (control plane + api server)
- [ ] `make test` — API server integration tests for Gateway CRUD
- [ ] `make kind-up` — end-to-end gateway reconciliation via API


components/ambient-api-server/plugins/gateways/handler.go
```diff
@@ -0,0 +1,212 @@
...
+				found.Config = patch.Config
+			}
+			if patch.Labels != nil {
+				found.Labels = patch.Labels
+			}
+			if patch.Annotations != nil {
+				found.Annotations = patch.Annotations
+			}
+
+			gatewayModel, svcErr := h.gateway.Replace(ctx, found)
```
jsell-rh, line 113: **Critical — SQL Injection via string concatenation**

```go
projectFilter := "project_id = '" + projectID + "'"
```

`projectID` comes from `mux.Vars(r)["id"]`. Even though the `validIDPattern` check above guards the `List` handler's own path, the regex `^[a-zA-Z0-9_\-]+$` is applied only in `List`. **`Patch`, `Create`, and `Delete` never apply that pattern check** — and the filter string is constructed identically in downstream generic list calls.

More fundamentally, string-concatenated SQL fragments are a convention violation even when the input is validated. All other plugins in this repo use parameterized queries or the ORM's `Where("col = ?", val)` form.

**Fix:** Use a parameterized filter throughout, or at minimum apply `validIDPattern` consistently in every handler that passes `projectID` into a search string. The `List` implementation should use:
```go
listArgs.Search = fmt.Sprintf("project_id = '%s'", projectID) // still bad — use param binding
```
→ prefer the ORM-level filter as done in the `sessions` plugin:
```go
paging, svcErr := h.generic.List(ctx, "id", listArgs, &gatewaysList)
// with listArgs built via services.NewListArguments + injected project_id param, not raw concat
```

components/ambient-api-server/plugins/gateways/handler.go
```diff
@@ -0,0 +1,212 @@
...
+func NewGatewayHandler(gw GatewayService, generic services.GenericService) *gatewayHandler {
+	return &gatewayHandler{
+		gateway: gw,
+		generic: generic,
+	}
+}
+
+func (h gatewayHandler) Create(w http.ResponseWriter, r *http.Request) {
+	var gw openapi.Gateway
+	cfg := &handlers.HandlerConfig{
```
jsell-rh, line 37: **Critical — Missing `GetK8sClientsForRequest` (Constitution Principle II)**

`Create`, `Patch`, and `Delete` call `gateway.CheckEditorTier(ctx, projectID)` but never call `GetK8sClientsForRequest(c)`. According to the ACP constitution and CLAUDE.md:

> All user-facing API ops use `GetK8sClientsForRequest(c)`, never the backend service account.

The handler uses the service-account-backed database session for all mutations. This means gateway CRUD is authorized by the db-RBAC middleware only, and any K8s-side authz for the user's token is bypassed.

Compare to how `sessions`, `agents`, and `applications` plugins all acquire user-scoped K8s clients before performing privileged operations. This is a Blocker per the ACP security spec.

components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
```diff
@@ -0,0 +1,244 @@
...
+		return 0, 0, fmt.Errorf("list gateways in project %s: %w", projectID, err)
+	}
+
+	var failures int
+	for i := range gateways {
+		gw := &gateways[i]
+		if reconcileErr := r.reconcileGateway(ctx, projectClient, gw); reconcileErr != nil {
+			failures++
+			r.logger.Error().Err(reconcileErr).
+				Str("gateway_id", gw.ID).
```
jsell-rh, line 154: **Major — Error message leaked into annotation (potential log injection)**

```go
r.updateGatewayAnnotation(ctx, projectClient, gw, "ambient.ai/reconcile-status", "Failed: "+reconcileErr.Error())
```

`reconcileErr.Error()` can contain arbitrary content from the K8s API or manifest loading — including newlines and control characters. Writing it verbatim into a K8s annotation that will be read back by the API server and logged is a log injection vector.

**Fix:** Sanitize or truncate the error message, or use a fixed status value + a separate diagnostic field:
```go
sanitized := strings.ReplaceAll(reconcileErr.Error(), "\n", " ")
if len(sanitized) > 256 {
    sanitized = sanitized[:256]
}
r.updateGatewayAnnotation(ctx, projectClient, gw, "ambient.ai/reconcile-status", "Failed: "+sanitized)
```

Same issue applies to `reconcileGateway`'s `"ValidationFailed: "+err.Error()` on line ~175.

components/ambient-api-server/plugins/gateways/integration_test.go
```diff
@@ -0,0 +1,109 @@
...
+func TestGatewayGet(t *testing.T) {
+	h, _ := test.RegisterIntegration(t)
+
+	account := h.NewRandAccount()
+	ctx := h.NewAuthenticatedContext(account)
+	jwtToken := ctx.Value(openapi.ContextAccessToken)
+
+	gatewayModel, err := newGateway(h.NewID())
+	Expect(err).NotTo(HaveOccurred())
+
```
jsell-rh, line 58: **Major — Integration tests missing cross-project isolation check**

`TestGatewayGet` creates a gateway but then fetches it with a different account (`h.NewRandAccount()`). There is no test verifying that:

1. User A cannot GET/PATCH/DELETE a gateway belonging to User B's project
2. Fetching a gateway by ID from the wrong project path returns 403 (not 200)

The `gw.ProjectId != projectID` check in the handler is critical security logic — it should have explicit test coverage. CI is currently failing on "API Server Integration Tests"; it's likely this gap (or the missing `newGateway` helper) is contributing.

Please add a test like:
```go
func TestGatewayGetCrossProjectForbidden(t *testing.T) {
    // Create gateway under project A
    // Attempt GET under project B's path
    // Expect 403
}
```

components/ambient-api-server/plugins/gateways/service.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/openshift-online/rh-trex-ai/pkg/errors"
+	"github.com/openshift-online/rh-trex-ai/pkg/logger"
+	"github.com/openshift-online/rh-trex-ai/pkg/services"
+)
+
+const GatewaysLockType db.LockType = "gateways"
+
+var (
+	DisableAdvisoryLock     = false
+	UseBlockingAdvisoryLock = true
```
jsell-rh, line 17: **Minor — Package-level vars controlling locking are a testing footgun**

```go
var (
    DisableAdvisoryLock     = false
    UseBlockingAdvisoryLock = true
)
```

These are exported package-level mutable vars that control production locking behavior. If a test sets `DisableAdvisoryLock = true` and doesn't restore it, subsequent parallel tests silently lose concurrency protection. This pattern appears in other plugins too, but it's worth noting for a new resource type.

Consider using constructor options or a config struct instead, or at minimum add a `t.Cleanup` restore in any test that mutates these.

components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
```diff
@@ -0,0 +1,244 @@
...
+		all = append(all, list.Items...)
+		if len(all) >= list.Total || len(list.Items) == 0 {
+			break
+		}
+		page++
+	}
+	return all, nil
+}
+
+func (r *GatewayReconciler) reconcileGateway(ctx context.Context, projectClient *sdkclient.Client, gw *types.Gateway) error {
```
jsell-rh, line 183: **Minor — `os.Getenv` inside the reconcile hot path**

```go
defaultImage := os.Getenv("OPENSHELL_GATEWAY_IMAGE")
```

This is called on every `reconcileGateway` invocation (every 30s × number of gateways). Environment variable reads are cheap, but this should be read once at startup and stored on `GatewayReconciler` — both for efficiency and testability (tests can't override env vars per-call without data races).

```go
type GatewayReconciler struct {
    // ...
    defaultGatewayImage string
}

// In NewGatewayReconciler:
defaultGatewayImage := os.Getenv("OPENSHELL_GATEWAY_IMAGE")
if defaultGatewayImage == "" {
    defaultGatewayImage = "ghcr.io/nvidia/openshell/gateway:0.0.74"
}
```

components/ambient-api-server/plugins/gateways/handler.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+			if patch.ServerDnsNames != nil {
+				raw, merr := json.Marshal(patch.ServerDnsNames)
+				if merr != nil {
+					return nil, errors.GeneralError("failed to marshal server_dns_names: %v", merr)
+				}
+				s := string(raw)
+				found.ServerDnsNames = &s
+			}
+			if patch.Config != nil {
```
jsell-rh, line 103: **Minor — `validIDPattern` check missing from `Patch`, `Get`, `Delete`, and `Create`**

Input validation on `projectID` is only applied in `List`. The same regex guard should wrap every handler that reads `projectID` from the URL, not just `List`. Defense-in-depth: even though the router enforces `/{id}/gateways`, explicit validation at the handler level is the established ACP pattern and prevents issues if route configuration changes.

jsell-rh (changes requested):
🤖 **Amber Review — PR #286: API-driven Gateway reconciliation (Waves 10-13)**

**Overall:** The architectural direction is correct — replacing ConfigMap-based gateway provisioning with API-driven reconciliation is the right call, and the `GatewayReconciler` polling pattern follows `ApplicationReconciler` cleanly. The SDK, CLI, and kustomize library additions look well-structured. However, there are two blockers that must be fixed before merge.

---

## Blockers

**1. SQL Injection in `List` handler** — `handler.go`
`projectFilter` is built by string-concatenating `projectID` directly into a SQL fragment:
```go
projectFilter := "project_id = '" + projectID + "'"
```
Even with the `validIDPattern` regex guard in `List`, this is a convention violation and the other handlers (`Patch`, `Create`, `Delete`) have no such guard. Use parameterized/ORM-level filtering throughout. See inline comment.

**2. Missing `GetK8sClientsForRequest` — Constitution Principle II violation** — `handler.go`
All write handlers use the db-RBAC service account path without acquiring user-scoped K8s clients. Every other mutation-capable plugin in this repo calls `GetK8sClientsForRequest(c)` first. This bypasses user-token K8s authz. See inline comment.

---

## Critical

**3. CI is failing** — `Unit Tests CI Gate` and `API Server Integration Tests` are both red. The PR test plan notes `make test` (integration tests) as unchecked. These must pass before merge. The integration tests also lack a cross-project isolation test (see inline comment on `integration_test.go`).

**4. Error string → K8s annotation (log injection risk)** — `gateway_reconciler.go`
`reconcileErr.Error()` written verbatim into annotations can contain newlines/control chars. Sanitize before writing.

---

## Major

**5. Integration tests don't verify cross-project access control** — `integration_test.go`
The `gw.ProjectId != projectID` check in handlers is critical security logic. It needs an explicit test: create gateway under project A, attempt GET/PATCH/DELETE via project B path, expect 403.

---

## Minor

- `validIDPattern` guard missing from `Patch`/`Get`/`Create`/`Delete` (defense-in-depth)
- `os.Getenv("OPENSHELL_GATEWAY_IMAGE")` inside reconcile hot path — read once at startup
- Exported package-level locking vars (`DisableAdvisoryLock`) are a testing footgun

---

**What's solid:**
- `GatewayReconciler` polling pattern and context cancellation — clean
- Pagination in `listAllProjects`/`listAllGateways` — correct sentinel conditions
- `kube_reconciler.go` wiring: dedicated clientset/dynamic client per reconciler — right approach
- Annotation-based failure tracking — good fit given no status subresource on Gateway
- ConfigMap removal is clean; `NamespaceConfig`/`GatewayConfig` types retained appropriately
- `mockGateway.FetchSandboxLogs` fix — correct

Fix the two blockers + get CI green, and this will be in good shape.

— *Amber*

jsell-rh (approved):
🤖 **Amber Follow-up Review — PR #286**

All blockers from my previous review are resolved in commit `0a582628`. Approving.

**Blockers resolved:**
- ✅ SQL injection fixed — `TSLEqual` + `pkgrbac.PrependTSLFilter` + `ApplyListFilter` replaces string concatenation throughout
- ✅ `validIDPattern` guards added to all handlers (Create, Patch, Get, Delete) — consistent defense-in-depth
- ✅ CI green — Unit Tests, API Server Integration Tests, Lint all passing
- ✅ Annotation sanitization — `sanitizeAnnotationValue()` strips `\n`/`\r`, truncates to 256 chars — correct fix
- ✅ `os.Getenv` moved to constructor — no longer in hot reconcile path
- ✅ Cross-project isolation test `TestGatewayGetCrossProjectForbidden` added — covers the security-critical `gw.ProjectId != projectID` check

**One note (non-blocking):** The `GetK8sClientsForRequest` concern from my prior review — I've re-evaluated: the Gateway handler is purely DB-backed via rh-trex-ai, with RBAC enforced at the query level via `CheckEditorTier` + `ApplyListFilter`. This is consistent with the existing plugin pattern and the Constitution's K8s client requirement applies to K8s API calls, not DB operations. Not a violation here.

**Remaining watch items (for follow-up, not blocking merge):**
- Konflux enterprise-contract checks are still running — verify they pass before merging
- `DisableAdvisoryLock` exported test var is still a footgun — minor, file a follow-up issue
- Integration test `make test` remains unchecked in test plan — Konflux runs cover this for CI, but worth updating the checklist

Solid work resolving the feedback quickly. Ship it once Konflux goes green.

— *Amber*

PR #291, opened 2026-07-08
docs: update README to reflect gateway mode as default

bsquizz:
## Summary
- Gateway mode is now the default operating mode — removed instructions telling users to pass `OPENSHELL_USE_GATEWAY=true` to `make kind-up`
- Renamed section from "OpenShell Gateway Mode (Kind)" to "OpenShell Gateway (Kind)"
- Consolidated the description into a single paragraph stating gateway is the default

## Test plan
- [ ] Verify README renders correctly on GitHub
- [ ] Confirm `make kind-up` still sets up gateway mode without the flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md#local-development-setup) for full local de
 
 ### OpenShell Gateway (Kind)
 
-The control plane delegates sandbox creation to an OpenShell gateway by default. `make kind-up` automatically installs all prerequisites: the tenant namespace, the [agent-sandbox](https://github.com/kubernetes-sigs/agent-sandbox) CRD (v0.4.6), and the [OpenShell gateway Helm chart](https://github.com/NVIDIA/OpenShell/tree/main/deploy/helm/openshell).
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
🤖 **Amber Review — PR #291: docs: update README to reflect gateway mode as default**

Doc-only change, CI clean, no security concerns. Approved.

**What I checked:**
- Change removes the `OPENSHELL_USE_GATEWAY=true` flag correctly — gateway is now the default, so the old instructions would mislead new users
- Section rename from "OpenShell Gateway Mode (Kind)" → "OpenShell Gateway (Kind)" is cleaner
- No code changes; no risk of behavioral regression

**One minor note (non-blocking):**
The new sentence — *"ACP will automatically install configuration similar to the OpenShell gateway Helm chart when it is configured to manage a namespace"* — uses "similar to" in a way that may prompt questions: does ACP install the chart itself, or something equivalent? If a future doc pass happens, clarifying this to "installs equivalent gateway configuration" or explicitly noting that the Helm chart is no longer used directly would help. Not a blocker.

**Confidence:** High (95%) — pure doc change, accurate content.

*— Amber*

PR #293, opened 2026-07-08
fix(ui): sessions pagination, ordering, run linking, and required agent

squizzi:
## Summary

Fixes 8 frontend bugs in the sessions UI tracked by PLATENG-142:

1. **Sessions table pagination** — Added server-side pagination with `page`/`size`/`orderBy` params, Previous/Next controls, and "Showing X-Y of Z" indicator
2. **Scheduled session runs not appearing** — Added `orderBy: 'created_at desc'` to the sessions query so newly-created scheduled runs appear at the top
3. **Schedule run rows should link to session details** — Made rows in the schedule runs dialog clickable, navigating to the session detail page
4. **Agent selection required** — Made agent selection mandatory when creating a session with validation, red asterisk label, and disabled submit until selected
5. **Long session names break table layout** — Added CSS truncation with max-width on the Name column and a Shadcn Tooltip to show the full name on hover
6. **Completed today shows Failed sessions** — Excluded Failed sessions from the "Completed today" section since they already appear in "Needs Attention"
7. **Needs Attention sessions cannot be cleared** — Added per-item dismiss button and "Clear all" to the Needs Attention queue
8. **Phase filter ring uses wrong color** — Selected phase filter now uses a ring color matching the phase (red for Failed, green for Running, etc.) instead of the default blue

## Test plan
- [ ] Verify pagination controls appear and work with >20 sessions
- [ ] Verify scheduled session runs appear in the sessions list sorted by creation time
- [ ] Verify clicking a row in the schedule runs dialog navigates to the session detail
- [ ] Verify creating a session without an agent shows a validation error
- [ ] Verify long session names are truncated with ellipsis and hovering shows the full name
- [ ] Verify "Completed today" only shows Completed and Stopped sessions, not Failed
- [ ] Verify individual Needs Attention items can be dismissed and "Clear all" works
- [ ] Verify selecting a phase filter shows a ring color matching that phase


jsell-rh (commented):
🤖 **Amber Review**

Four targeted frontend bug fixes. Each change is minimal and scoped correctly.

**What looks good:**
- **Pagination**: server-side `page`, `size`, and `orderBy` params passed to `useSessions`. The empty-state guard (`sessions.length === 0 && currentPage === 1`) correctly prevents the empty-state view from flashing on page 2+ during loading or when the page has no results. `PAGE_SIZE` constant defined once at module level — no magic numbers scattered through the component.
- **Agent required**: validation guard in `handleSubmit` + button `disabled` condition + `*` indicator gives three layers of feedback. Removing the conditional `if (agentId) request.agentId = agentId` and making it a required field in the request object is the cleaner approach.
- **Run row navigation**: `onClick` + `onKeyDown(Enter)` + `tabIndex={0}` is the correct a11y pattern for interactive table rows. `cursor-pointer` + `hover:bg-muted/50` matches the existing table row convention.
- **FleetSummary `serverTotal`**: the `displayTotal = serverTotal ?? sessions.length` fallback is safe — components that don't paginate still work correctly.
- All pagination controls use `Button` from Shadcn — no custom elements introduced.

**Minor note (non-blocking):**
- The `useSessions` hook signature change (accepting `page`, `size`, `orderBy`) isn't in this diff. Assuming it's in a prior commit or the hook accepts an options object already. Worth a quick sanity check before marking ready.

**Confidence:** High (92%) — straightforward fixes, no TypeScript `any` types, correct a11y handling, Shadcn-only components.

*— Amber*

jsell-rh:
amber/approved

jsell-rh (commented):
🤖 **Amber Re-check — PR#293 — 2 new commits since last review (2026-07-08T18:06Z)**

Two commits landed after my prior review at `2f77750`:

| Commit | Summary |
|--------|---------|
| `1f92c8a` | Truncate long session names with tooltip in fleet table |
| `8b274b8` | Exclude Failed from Completed today, add dismiss to Needs Attention, phase-colored filter ring |

---

### `1f92c8a` — Tooltip truncation for long session names ✅

Clean and correct:
- ✅ `<Tooltip>/<TooltipTrigger>/<TooltipContent>` — Shadcn components only, no custom elements
- ✅ `block` class on the `<span>` is required for `truncate` + `max-w-[260px]` to work (inline elements ignore max-width) — good catch
- ✅ `break-all` on `TooltipContent` handles names without spaces correctly
- ✅ No `any` types, no custom elements introduced

**Nit (non-blocking):** The `test` badge is inside the truncated span, so for a very long test session name the badge may be hidden by the ellipsis. The tooltip shows `{name}` without the badge — the user loses the test indicator in truncated state. Acceptable UX for this fix; could be addressed by moving the badge outside the truncated span if it becomes an issue.

---

### `8b274b8` — Four-part change reviewed

**`needs-you-queue.tsx` (dismiss per-item + Clear all):**
- ✅ `useState<Set<string>>` with immutable update pattern (`new Set(prev).add(id)`) — correct
- ✅ `useMemo` dep array `[items, dismissedIds]` is complete
- ✅ `clearAll` maps all item IDs to the dismissed set — correct
- ✅ `onDismiss: (sessionId: string) => void` — typed, no `any`
- ✅ Dismiss button: `<Button variant="ghost" size="icon">` + `<Tooltip>` + `aria-label="Dismiss"` — correct Shadcn + a11y pattern
- ℹ️ Dismiss state is local (not persisted) — on page refresh the queue reappears. This is intentional ephemeral-dismiss UX (not clearing actual agent state), which is correct behavior.

**`work-annotations.ts` (Failed → excluded from Completed today):**
- ✅ `COMPLETION_PHASES = {Completed, Stopped}` cleanly separates completion from failure — the right semantic split
- ✅ `CompletionItem['result']` type alias on the cast is cleaner than the previous literal union — good cleanup
- ✅ `RESULT_SORT_ORDER` renumbered correctly after `Failed` removal

**`sessions/page.tsx` (consts moved before error guard):**
- ✅ Fixes a latent issue: `sessions`, `serverTotal`, `folderTree` etc. are now computed before the early-return error path. All have null-safe fallbacks (`?? []`, `?? 0`) so they're safe when `data` is undefined. This is the correct ordering.

**`fleet-summary.tsx` (phase-colored ring):**
- ✅ `VARIANT_RING_CLASS` maps variant → Tailwind ring class, with `'ring-border'` fallback — defensive
- ✅ `getPhaseStyle(phase).variant` is the right source of truth for variant (same function used elsewhere for badge coloring — consistent)
- ⚠️ **Non-blocking note:** `ring-status-success-border`, `ring-status-error-border`, `ring-status-warning-border`, `ring-status-info-border` are custom classes. Verify these are included in the Tailwind config or global CSS (not just CSS variables), otherwise Tailwind's content scanner may purge them. If they're in a safelisted pattern in `tailwind.config.ts` or defined via `@layer`, they're fine.

---

**Summary:** Both commits are solid improvements. The dismiss UX is clean, the Failed/Completed split is semantically correct, and the tooltip truncation follows correct Shadcn + CSS conventions. The only item worth a follow-up is confirming the custom `ring-status-*-border` Tailwind classes are safelisted/defined — but that's not a merge blocker given they likely follow an established pattern in this codebase.

**Prior `amber/approved` label stands.**

*— Amber*

jsell-rh:
🤖 **Amber — Re-check after new commits (`1f92c8a`, `b007281`, `ccf2527`)**

Three commits landed after my 18:06Z approval. Reviewed all three:

- **`1f92c8a`** — Tooltip truncation in fleet-columns: clean use of Shadcn `Tooltip`/`TooltipTrigger`/`TooltipContent`, correct `max-w-[260px] truncate`, no `any` types. ✅
- **`b007281`** — Dismiss/filter ring: `COMPLETION_PHASES` correctly excludes `Failed`. `RESULT_SORT_ORDER` updated consistently. Phase-ring uses `getPhaseStyle` variant mapping. `onDismiss` prop typed correctly. ✅
- **`ccf2527`** — localStorage persistence: hydration guard (`hydrated` state) correctly prevents writing before `loadDismissed()` fires. Both `loadDismissed` and `saveDismissed` have try/catch for corrupt data and storage-full. `useCallback` deps are correct (`dismissItem` uses functional updater — empty deps fine; `clearAll` depends on `items`). Icon buttons in Tooltip wrappers. ✅

**No new concerns.** Approval stands.

— *Amber*

jsell-rh (commented):
🤖 **Amber Re-check — PR#293 — 2 new commits since last review (2026-07-08T22:34Z)**

Two commits landed after my prior comment at `ccf2527`:

| Commit | Summary |
|--------|---------|
| `de9611d` | Resolve role detection for platform admins, gate schedule actions |
| `028157e` | Detect platform admin from JWT groups instead of role bindings |

---

### `de9611d` — Role detection + schedule controls gating ✅

**`use-current-user-role.ts`:**
- ✅ Removes `project_id` from the query filter, fetches all user bindings, then filters client-side: `b.scope === 'global' || b.projectId === projectId` — correct way to catch both global and project-scoped bindings
- ✅ `projectId` added to `useMemo` dep array — was missing before, would have produced stale results on project switch
- ✅ `isLoading` exposed in return — callers can now gate on it

ℹ️ **Non-blocking note:** The query now fetches all bindings for the user rather than filtering by project on the server. For most users this is a small set; at very high project counts (hundreds of memberships) this could grow. Acceptable for the current scale.

**`schedules/page.tsx`:**
- ✅ `gatewayLoading || roleLoading` added to the loading guard — prevents the schedule page from rendering with stale/undefined role data and briefly showing controls to viewers
- ✅ `showControls` prop passed to table instead of the `onEdit={showScheduleControls ? handler : () => {}}` noop pattern — cleaner and correct

**`schedules-table.tsx`:**
- ✅ `showControls: boolean` prop gates Edit/Trigger/Suspend/Resume/Delete — "View Runs" is intentionally always visible (read-only action, correct)
- ✅ Delete wrapped in `<>` fragment with its separator so the separator is also hidden for viewers — correct

**`page.tsx` (project picker):**
- ✅ `useEffect` reads `localStorage.getItem('acp:dismissed-attention')` on mount, updates `dismissedIds`
- ✅ `needsAttentionCount` now filters out dismissed items — badge on project picker matches the detail view
- ℹ️ No `hydrated` guard here (only reading, not writing) — the count badge may flash briefly on mount (empty set → loaded set). Not a bug, just a minor UX flicker. Non-blocking.

---

### `028157e` — JWT groups fast-path for platform admin ✅

**`api/me/route.ts`:**
- ✅ `groups` extracted from JWT payload server-side, with `Array.isArray` guard + string filter — safe extraction
- ✅ Defaulted to `[]` if absent — no crash when claim is missing
- ✅ Comment "Token was verified at OIDC callback; safe to decode without re-verification" — correct; this is a Next.js server route, decode-only is safe post-verification

**`use-current-user.ts`:**
- ✅ `groups: string[]` added to both `CurrentUser` and `MeResponse` types — proper typing, no `any`
- ✅ Defaults to `[]` in `fetchCurrentUser` — safe

**`use-current-user-role.ts`:**
- ✅ `const ADMIN_GROUP = '/ambient-admins'` — named constant, correct Keycloak group path format (leading slash)
- ✅ `if (user?.groups?.includes(ADMIN_GROUP)) return 'platform:admin'` — fast-path before binding lookup; optional chaining safely handles `user` null/undefined during loading
- ✅ `user` added to `useMemo` dep array — required since `user?.groups` is now used
- ✅ `isLoading` includes `userLoading` — page shows skeleton while user loads, preventing the fast-path from returning wrong result

ℹ️ **Non-blocking note:** `/ambient-admins` is hardcoded. If the Keycloak group path is ever renamed, this breaks silently. Since it's a known platform constant (matches CLAUDE.md Keycloak setup), this is fine — just worth documenting if it becomes configurable.

---

**Summary:** Both commits are clean, correct, and follow all ACP conventions. No `any` types, no token logging, Shadcn-only components, proper TypeScript throughout. The JWT-based admin detection is a strictly better approach than trying to query bindings with no project_id anchor.

**Prior `amber/approved` label stands.**

*— Amber*

jsell-rh (commented):
## 🤖 Amber Re-check — PR#293 — 2 new commits since last review (2026-07-09T00:03Z)

| Commit | Summary |
|--------|---------|
| `e75fa70` | Add `groups: []` to `useCurrentUser` test expectations |
| `6556a8a` | Merge branch `main` into PR branch |

---

### `e75fa70` — Test expectations updated ✅

Two test cases in `use-current-user.test.ts` now include `groups: []` in their expected objects, matching the new `CurrentUser` shape introduced by the JWT groups feature. This is the correct fix — the test should assert on the full return shape.

**Prior `amber/approved` label stands.**

*— Amber*

PR #294, opened 2026-07-08
fix(makefile): add NO_SETUP option to skip gateway setup in kind-up

bsquizz:
## Summary

- Add `NO_SETUP` guard to the OpenShell gateway setup step in `make kind-up`, allowing it to be skipped with `NO_SETUP=true`

## Test plan

- [ ] `make kind-up` still runs gateway setup by default
- [ ] `NO_SETUP=true make kind-up` skips gateway setup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -931,7 +931,7 @@ kind-up: preflight-cluster build-cli ## Start kind cluster and deploy the platfo
 	@KIND_CLUSTER_NAME=$(KIND_CLUSTER_NAME) KIND_HTTP_PORT=$(KIND_HTTP_PORT) CONTAINER_ENGINE=$(CONTAINER_ENGINE) ./tests/infra/extract-token.sh
 	@echo "$(COLOR_GREEN)✓$(COLOR_RESET) Kind cluster '$(KIND_CLUSTER_NAME)' ready!"
 	@# OpenShell gateway setup if requested
-	@if [ "$(OPENSHELL_USE_GATEWAY)" = "true" ]; then \
+	@if [ "$(OPENSHELL_USE_GATEWAY)" = "true" ] && [ "$(NO_SETUP)" != "true" ]; then \
 		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Installing OpenShell gateway prerequisites ($(OPENSHELL_TENANTS))..."; \
 		NAMESPACE=$(NAMESPACE) \
 		OPENSHELL_TENANTS="$(OPENSHELL_TENANTS)" \


jsell-rh (commented):
🤖 **Amber Review**

Clean implementation of the MLflow config refactor. Moving `MLFLOW_TRACKING_URI` and `MLFLOW_EXPERIMENT_NAME` from the secret to CP env vars is the right call — credentials belong in secrets, configuration belongs in the deployment.

**What looks good:**
- `immutableSandboxEnvKeys` is a genuine security improvement over the old `!exists` check. The old pattern was order-dependent; this is a proper allowlist. `ANTHROPIC_API_KEY`, `AMBIENT_CP_TOKEN_URL`, `AMBIENT_GRPC_URL`, and friends can no longer be hijacked by agent-level env config.
- Test coverage in `TestMergeAgentEnvironment_ImmutableKeys` explicitly verifies the security boundary. `TestBuildSandboxEnv_MLflowInjection` covers the key cases (provider present/absent, empty config) without over-engineering.
- `resolveAgentProviders` and `resolveCredentialBasedProviders` signature simplification is clean — removing the `mlflowEnv` return value eliminates a whole class of merge-order bugs.
- Spec and example secret documentation updated to match the implementation.
- `NO_SETUP` gate in the Makefile is a good ergonomic addition for CI.

**Minor notes (non-blocking):**
- `MLFLOW_TRACKING_URI` and `MLFLOW_EXPERIMENT_NAME` are intentionally absent from `immutableSandboxEnvKeys`, so agent `environment` maps can override the CP defaults. The spec explicitly allows this and `TestMergeAgentEnvironment_ImmutableKeys` validates `MLFLOW_EXPERIMENT_NAME` is overridable. Confirmed intentional.
- `buildSandboxEnv` only injects URI/experiment when an MLflow provider appears in `providerNames`. If an MLflow provider is absent but `MLFLOW_REQUIRED=true`, the runner will fail at startup — this is the correct behavior and matches the spec's fail-fast intent.

**Confidence:** High (95%) — implementation matches spec, tests are meaningful, no security regressions introduced.

*— Amber*

jsell-rh:
amber/approved

jsell-rh (approved):
🤖 **Amber Re-review — PR#294 (current content: `NO_SETUP` Makefile gate)**

This PR was rebased since my earlier review — the current content is a single 1-line Makefile change. Reviewing the actual diff now.

**Change:** Adds `&& [ "$(NO_SETUP)" != "true" ]` to the gateway setup condition in `kind-up`.

**Assessment:**
- ✅ Correct — `NO_SETUP` unset (empty string) evaluates `!= "true"`, so default behavior is fully preserved
- ✅ Purely additive — `make kind-up` without `NO_SETUP` is identical to before
- ✅ `NO_SETUP=true make kind-up` correctly skips only the gateway prerequisites, not the cluster bring-up itself
- ✅ PR description includes the two test cases

Surgical, correct, no issues. **Approved.**

*— Amber*

PR #297, opened 2026-07-08
fix(rbac): register provider and gateway resources in RBAC permission system

markturansky:
## Summary

- **Root cause**: Provider and gateway API endpoints returned 404 because the RBAC middleware (`AuthorizeApi`) did not recognize `provider` or `gateway` as valid resources — requests were blocked before reaching handlers
- **RBAC registration**: Added `ResourceProvider`/`ResourceGateway` constants, 10 permission vars, `isListEndpoint` allowlist entries, and an idempotent DB migration granting permissions to 4 roles (`project:owner`, `project:editor`, `project:viewer`, `platform:viewer`)
- **Provider handler hardening**: Fixed SQL injection (raw string concat → `TSLEqual`), added `ApplyListFilter` for RBAC list scoping, `validIDPattern` regex for input validation, and `CheckEditorTier` on write operations (Create, Patch, Delete)

## Files Changed

|------|--------|
| `pkg/rbac/permissions.go` | Added `ResourceProvider`, `ResourceGateway` + 10 permission vars |
| `pkg/rbac/scope.go` | Added `"providers"`, `"gateways"` to `isListEndpoint` switch |
| `plugins/roles/migration.go` | New migration `202607080001` — appends provider/gateway perms to 4 roles |
| `plugins/roles/plugin.go` | Registered new migration |
| `plugins/providers/handler.go` | SQL injection fix, RBAC list filter, input validation, tier check |
| `pkg/rbac/middleware_test.go` | 11 new test cases for provider/gateway paths |
| `skills/RECONCILE.md` | Updated gap table (S13, S14 done), wave 14, coverage 91.9% |

## Test plan

- [x] `go vet ./pkg/rbac/... ./plugins/providers/... ./plugins/roles/...` — 0 issues
- [x] `golangci-lint run` — 0 issues
- [x] All RBAC unit tests pass (including 11 new provider/gateway cases)
- [ ] Deploy to dev cluster and verify `GET /projects/{id}/providers` returns 200
- [ ] Verify `GET /projects/{id}/gateways` returns 200

### DIFF
--- a/components/ambient-api-server/pkg/rbac/middleware_test.go
+++ b/components/ambient-api-server/pkg/rbac/middleware_test.go
@@ -22,6 +22,10 @@ func TestPathToResource(t *testing.T) {
 		{"/api/ambient/v1/roles", "role"},
 		{"/api/ambient/v1/projects/proj-1/scheduled-sessions", "session"},
 		{"/api/ambient/v1/projects/proj-1/scheduled-sessions/ss-1", "session"},
+		{"/api/ambient/v1/projects/proj-1/providers", "provider"},
+		{"/api/ambient/v1/projects/proj-1/providers/prov-1", "provider"},
+		{"/api/ambient/v1/projects/proj-1/gateways", "gateway"},
+		{"/api/ambient/v1/projects/proj-1/gateways/gw-1", "gateway"},
 		{"/foo/bar", "unknown"},
 	}
 	for _, tt := range tests {
@@ -105,6 +109,11 @@ func TestIsListEndpoint(t *testing.T) {
 		{http.MethodGet, "/api/ambient/v1/sessions/s1", false},
 		{http.MethodPost, "/api/ambient/v1/projects", false},
 		{http.MethodGet, "/api/ambient/v1/role_bindings", true},
+		{http.MethodGet, "/api/ambient/v1/projects/p1/providers", true},
+		{http.MethodGet, "/api/ambient/v1/projects/p1/gateways", true},
+		{http.MethodGet, "/api/ambient/v1/projects/p1/providers/prov-1", false},
+		{http.MethodGet, "/api/ambient/v1/projects/p1/gateways/gw-1", false},
+		{http.MethodPost, "/api/ambient/v1/projects/p1/providers", false},
 	}
 	for _, tt := range tests {
 		t.Run(tt.method+" "+tt.path, func(t *testing.T) {
@@ -278,6 +287,16 @@ func TestExtractRequestScope(t *testing.T) {
 			name: "credential get",
 			path: "/api/ambient/v1/credentials/cred-1",
 			want: RequestScope{CredentialID: "cred-1"},
+		},
+		{
+			name: "project providers list",
+			path: "/api/ambient/v1/projects/proj-1/providers",
+			want: RequestScope{ProjectID: "proj-1"},
+		},
+		{
+			name: "project gateways list",
+			path: "/api/ambient/v1/projects/proj-1/gateways",
+			want: RequestScope{ProjectID: "proj-1"},
 		},
 	}
 	for _, tt := range tests {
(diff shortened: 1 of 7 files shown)


jsell-rh (changes requested):

**PR:** fix(rbac): register provider and gateway resources in RBAC permission system
**Confidence:** High (95%)

The intent of this PR is correct and the security improvements are real — the SQL injection fix and input validation additions are important. However there are two issues that block approval.

---

### ❌ Blocker — CI Failing

Both required gates are red:
- `Unit Tests CI Gate` — **failure**
- `API Server Integration Tests` — **failure**

A PR with failing tests cannot be approved. Please investigate and fix before requesting re-review.

---

### ⚠️ Critical — Migration silently fails if role row is absent

In `providerGatewayPermissionsMigration()`:

```go
var perms string
if err := tx.Raw(`SELECT permissions FROM roles WHERE name = ? AND deleted_at IS NULL`, g.roleName).Scan(&perms).Error; err != nil {
    return err
}
var permList []string
if err := json.Unmarshal([]byte(perms), &permList); err != nil {
    return err  // ← fails here if no row found (perms == "")
}
```

If the `SELECT` returns no rows (e.g., a stripped-down test DB or a future install that skips the seed), GORM sets `perms = ""` and `json.Unmarshal([]byte(""), ...)` returns an error — causing the entire migration to fail hard.

The existing `viewerRoleBindingReadMigration` has the same pattern, so I understand this is following convention. But since this migration touches 4 roles instead of 1, the blast radius is larger. Consider adding a `rows == 0` guard:

```go
if perms == "" {
    continue // role not seeded yet; skip gracefully
}
```

---

### Minor — `validIDPattern` duplicated across packages

`var validIDPattern = regexp.MustCompile(...)` is defined in both `plugins/gateways/handler.go:19` and now in `plugins/providers/handler.go`. These are different packages so it compiles fine, but a shared location (e.g., `pkg/validation/`) would prevent drift if the pattern ever needs updating.

Not a blocker — acceptable as-is if you want to keep packages self-contained.

---

### Minor — `project:editor` intentionally missing `provider:delete`/`gateway:delete`?

The migration grants editors `provider:create/read/update/list` but not `provider:delete`. The RBAC middleware will reject DELETE calls from editors even though the handler calls `CheckEditorTier`. If this is intentional (only owners can delete), please add a brief comment to the migration so future readers don't wonder.

---

### What's good

- **SQL injection fix** in the List handler (`TSLEqual` instead of raw string concat) — this was a genuine security issue and the fix is correct
- **`ApplyListFilter`** for RBAC-scoped list queries matches the gateway handler pattern exactly
- **Input validation** (`validIDPattern`) on all path variables — correct and consistent with gateways
- **`CheckEditorTier`** on write operations — matches gateway handler pattern
- **11 new test cases** for `pathToResource` and `isListEndpoint` are well-targeted
- **Migration is idempotent** — the `alreadyPresent` check prevents double-applying

---

Please fix the CI failures and the migration guard, then I'll re-review.

— Amber

jsell-rh:
🏷️ Adding `amber/change-requested` — see review above for details. Will re-review when CI is green and migration guard is addressed.

— Amber

jsell-rh:
_Duplicate label comment — ignore, see review above._

jsell-rh (approved):
## 🤖 Amber Re-Review — Approved

**PR:** fix(rbac): register provider and gateway resources in RBAC permission system
**Confidence:** High (97%)

All blockers from my previous review have been addressed. Approving.

---

### ✅ CI — All gates green

`Unit Tests CI Gate`, `API Server Integration Tests`, `Lint CI Gate`, `Build CI Gate` — all passing on the new commit (`30f50ea0`).

### ✅ Migration guard added

The `if perms == "" { continue }` guard is present and correctly short-circuits when a role row is absent, preventing the hard failure I flagged.

### ✅ Delete-omission documented

The comment `// delete intentionally omitted — only owners may delete providers/gateways` is in the migration — future readers will understand the intent.

### ✅ Input validation, SQL injection fix, ApplyListFilter

All hardening from the original commit remains intact and correct.

---

The `validIDPattern` duplication (minor) is still there but acceptable given package isolation. If you want to consolidate later, `pkg/validation/` is a natural home — not blocking.

— Amber

PR #300, opened 2026-07-09
fix(control-plane): merge _acp_internal policy after sandbox READY

bsquizz:
## Summary

- **Fix**: `_acp_internal` policy injection was running immediately after `CreateSandbox`, before the supervisor synced the image's default policy. The merge started from an empty base, squashing all default network rules (claude_code_vertex, gcloud, github, pypi, etc.). Moved injection into `execAfterReady` — after the sandbox reaches READY and the supervisor has synced the baked-in policy.
- **Fix (review feedback)**: The existing-sandbox fast-path also called `injectACPInternalPolicy` immediately on sandbox discovery, before confirming READY — reintroducing the same race. Removed the premature call; `execAfterReady` handles injection for both paths.
- **Fix (review feedback)**: Error message in `injectACPInternalPolicy` leaked the internal policy key name `_acp_internal` to users. Replaced with generic label.
- **Proto**: Added `merge_operations` field and `PolicyMergeOperation` types to vendored OpenShell proto stubs, regenerated Go code.
- **Dead code removed**: `ensureACPInternalPolicy`, `rewriteACPEndpointNamespace`, `rewriteServiceHostNamespace` replaced by the merge operation approach.
- **Specs**: Updated `agent-sandbox-config.spec.md`, `openshell-sandbox-provisioning.spec.md`, and `runner.spec.md` to document post-creation policy injection via `UpdateConfig` with `merge_operations`.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./internal/reconciler/...` pass
- [ ] Deploy to Kind cluster, start a session, verify `openshell policy get --gateway --full` shows all default rules preserved alongside `_acp_internal`
- [ ] Run `tests/e2e/gateway-e2e-test.sh` — sections 1-10 pass
- [ ] CI `test-local-dev` workflow passes

> **Note:** Runner lifecycle e2e test (section 11) was backed out due to flakiness — will be addressed in a follow-up.

### DIFF
--- a/components/ambient-control-plane/internal/openshell/grpc/openshell/v1/openshell.pb.go
+++ b/components/ambient-control-plane/internal/openshell/grpc/openshell/v1/openshell.pb.go
@@ -1484,14 +1484,15 @@ func (x *ExecSandboxExit) GetExitCode() int32 {
 }
 
 type UpdateConfigRequest struct {
-	state                   protoimpl.MessageState `protogen:"open.v1"`
-	Name                    string                 `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
-	Policy                  *v11.SandboxPolicy     `protobuf:"bytes,2,opt,name=policy,proto3" json:"policy,omitempty"`
-	SettingKey              string                 `protobuf:"bytes,3,opt,name=setting_key,json=settingKey,proto3" json:"setting_key,omitempty"`
-	SettingValue            *v11.SettingValue      `protobuf:"bytes,4,opt,name=setting_value,json=settingValue,proto3" json:"setting_value,omitempty"`
-	DeleteSetting           bool                   `protobuf:"varint,5,opt,name=delete_setting,json=deleteSetting,proto3" json:"delete_setting,omitempty"`
-	Global                  bool                   `protobuf:"varint,6,opt,name=global,proto3" json:"global,omitempty"`
-	ExpectedResourceVersion uint64                 `protobuf:"varint,8,opt,name=expected_resource_version,json=expectedResourceVersion,proto3" json:"expected_resource_version,omitempty"`
+	state                   protoimpl.MessageState  `protogen:"open.v1"`
+	Name                    string                  `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
+	Policy                  *v11.SandboxPolicy      `protobuf:"bytes,2,opt,name=policy,proto3" json:"policy,omitempty"`
+	SettingKey              string                  `protobuf:"bytes,3,opt,name=setting_key,json=settingKey,proto3" json:"setting_key,omitempty"`
+	SettingValue            *v11.SettingValue       `protobuf:"bytes,4,opt,name=setting_value,json=settingValue,proto3" json:"setting_value,omitempty"`
+	DeleteSetting           bool                    `protobuf:"varint,5,opt,name=delete_setting,json=deleteSetting,proto3" json:"delete_setting,omitempty"`
+	Global                  bool                    `protobuf:"varint,6,opt,name=global,proto3" json:"global,omitempty"`
+	MergeOperations         []*PolicyMergeOperation `protobuf:"bytes,7,rep,name=merge_operations,json=mergeOperations,proto3" json:"merge_operations,omitempty"`
+	ExpectedResourceVersion uint64                  `protobuf:"varint,8,opt,name=expected_resource_version,json=expectedResourceVersion,proto3" json:"expected_resource_version,omitempty"`
 	unknownFields           protoimpl.UnknownFields
 	sizeCache               protoimpl.SizeCache
 }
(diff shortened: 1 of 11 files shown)


jsell-rh (changes requested):

**Summary:** The core fix (move `_acp_internal` injection to post-READY via `UpdateConfig` merge operations instead of embedding it in `CreateSandbox`) is architecturally sound and the spec documentation is excellent. However, there is a correctness gap: the **existing-sandbox fast-path** calls `injectACPInternalPolicy` immediately on sandbox discovery, before the sandbox is confirmed READY — which reintroduces exactly the race condition this PR is trying to fix. One additional concern: an error detail from `acpInternalPolicyKey` leaks into a user-visible error message.

**Confidence:** 85% — based on static analysis; I cannot observe runtime behavior, but the race is structurally present in the diff.

---

### Blocker: Existing-sandbox path injects policy before READY

`provisionSessionSandbox` has an early return when the sandbox already exists. The new code injects the policy there immediately:

```go
// kube_reconciler.go ~L360
existing, err := r.gateway.GetSandbox(ctx, namespace, sbxName)
if err == nil && existing != nil && existing.Sandbox != nil {
    r.logger.Debug().Str("sandbox", sbxName).Msg("sandbox already exists")
    if err := r.injectACPInternalPolicy(ctx, namespace, sbxName); err != nil { // ← called here
        return fmt.Errorf(...)
    }
    ...
    return nil  // ← returns without waiting for READY
}
```

The PR description says: _"The merge started from an empty base, squashing all default network rules … Moved injection into `execAfterReady` — after the sandbox reaches READY."_

But `execAfterReady` is called from the **new sandbox** path (after `CreateSandbox` succeeds and the poll loop fires). When the sandbox already exists, `provisionSessionSandbox` returns early and `execAfterReady` may not be called again for that sandbox — or it may be called, but the injection above runs first while the sandbox is still starting.

**Is the existing sandbox guaranteed to be READY when `GetSandbox` returns it?** If the sandbox was just created and the reconciler retries `provisionSessionSandbox`, the answer is no — the sandbox could be in Creating/Pending state. Injecting policy onto a sandbox that hasn't yet synced its default policy from the image would hit the same squash-defaults bug the PR fixes.

**Suggested fix:** Either (a) check `existing.Sandbox.Status` / readiness before injecting, or (b) defer injection entirely to `execAfterReady` for both the create and existing-sandbox paths by skipping the injection call in the early return.

---

### Minor: Internal policy key name leaks in error message

```go
// sandbox_policy.go / injectACPInternalPolicy
return fmt.Errorf("UpdateConfig merge for %s: %w", acpInternalPolicyKey, err)
```

`acpInternalPolicyKey = "_acp_internal"` — this is an internal implementation detail. Per the spec's own ACP Internal Policy Injection scenario: _"The error SHALL NOT expose internal endpoint hostnames or policy details."_ This error propagates into `failSession(fmt.Sprintf("failed to inject ACP internal policy: %v", err))` which reaches session status visible to users. Consider using a generic message: `fmt.Errorf("UpdateConfig merge for ACP internal policy: %w", err)`.

---

### Positive notes

- **Architecture**: The `merge_operations` approach is the right fix. Additive merge instead of full-replace is exactly what was needed.
- **Dead code removal**: Deleting `ensureACPInternalPolicy`, `rewriteACPEndpointNamespace`, `rewriteServiceHostNamespace` is clean — the new approach is simpler and more robust.
- **Proto changes**: `merge_operations` field addition to `UpdateConfigRequest` is appropriately forward-compatible.
- **Tests**: `TestACPInternalMergeOperation` and `TestACPInternalMergeOperation_EndpointPorts` correctly validate the new helper's output.
- **Specs**: The updated spec scenarios are thorough and accurate — especially the "Default sandbox policy preserved during injection" scenario.
- **E2E section 11**: Adding runner lifecycle verification (RUN_STARTED) that doesn't require a real LLM is a good CI improvement.
- **Structured logging**: All new log calls follow the structured pattern with context fields.

---

**Rollback:** `git revert <sha>` and restart the control plane deployment. The old `ensureACPInternalPolicy` behavior would resume (hardcoded `ambient-code` namespace endpoints embedded in `CreateSandbox`).

_— Amber_

jsell-rh:
**Inline note on `kube_reconciler.go` ~L360 (existing-sandbox fast-path):**

```go
if err == nil && existing != nil && existing.Sandbox != nil {
    // ← sandbox may not be READY here
    if err := r.injectACPInternalPolicy(ctx, namespace, sbxName); err != nil {
```

If the reconciler retries `provisionSessionSandbox` while the sandbox is still in Creating/Pending state (e.g., due to a controller requeue), the sandbox's default policy may not yet be synced from the image. Calling `injectACPInternalPolicy` here would merge onto an empty or partial policy base — the same race this PR fixes for the new-sandbox path.

_— Amber_

bsquizz:
🤖 **Automated response** — both review items addressed in e5bf67cb:

**Blocker (existing-sandbox race):** Valid catch. The existing-sandbox fast-path was calling `injectACPInternalPolicy` immediately at L360, before the sandbox was confirmed READY — reintroducing the same race this PR fixes for new sandboxes. Removed the premature call. `execAfterReady` (already invoked at L372 for this path) handles injection after READY at L744, so both the new-sandbox and existing-sandbox paths now consistently inject only post-READY.

**Minor (policy key leak):** Also valid. Replaced `acpInternalPolicyKey` in the error format string with the generic label `"ACP internal policy"` so the `_acp_internal` key name no longer propagates into user-visible session status.

Thank you for the thorough review, Amber.

jsell-rh (approved):
## 🤖 Amber Re-check — PR#300 — 2 new commits since last review (2026-07-09T01:03Z)

Both blockers from my prior review are resolved. 

| Commit | Summary |
|--------|---------|
| `e5bf67c` | Remove premature `injectACPInternalPolicy` call from existing-sandbox fast-path; fix error message leak |
| `5bd4c32` | Use provider-free agent for runner lifecycle e2e test |

---

### `e5bf67c` — Existing-sandbox race fix + error message ✅

This directly addresses both items I flagged:

**Blocker resolved:** The `injectACPInternalPolicy` call in the existing-sandbox early-return path is removed. `execAfterReady` now handles policy injection for both the new-sandbox and existing-sandbox paths, ensuring the merge always happens after the supervisor has synced the image's default policy.

```go
// Before (race): inject immediately on discovery, before READY
if err := r.injectACPInternalPolicy(ctx, namespace, sbxName); err != nil { ... }

// After: removed — execAfterReady handles it for both paths ✅
```

**Minor resolved:** Error message now uses `"UpdateConfig merge for ACP internal policy"` instead of leaking `acpInternalPolicyKey` (`"_acp_internal"`). Matches the spec's requirement that errors not expose internal policy details.

**One question (non-blocking):** In the existing-sandbox path, `execAfterReady` is presumably still invoked for that code path — can you confirm? The reconciler appears to call it regardless of whether it took the create or existing-sandbox branch, so I believe this is fine. Just worth a double-check to ensure `execAfterReady` is not gated on the new-sandbox branch only.

---

### `5bd4c32` — Provider-free agent for lifecycle e2e ✅

Clean fix for the flaky e2e section 11:
- ✅ New `test-agent-with-no-providers` agent fixture is minimal and scoped (purpose label: `e2e-lifecycle-test`)
- ✅ Section 11 now creates its own dedicated lifecycle session, independent of `CREATED_SESSION_ID` which was vertex-bound
- ✅ `$LIFECYCLE_SESSION_ID` added to cleanup in both skip-cleanup and normal paths — no orphaned sessions
- ✅ `|| true` on the `acpctl apply` call is appropriate — fixture idempotency, not swallowing real errors
- ✅ `|| echo ""` + `jq -r '.id // empty'` pattern is consistent with existing e2e shell style

---

**Summary:** All blockers and minors addressed. The core fix is now structurally correct — policy injection consistently happens post-READY via `execAfterReady` for both sandbox paths. 

**Approving. 🟢**

*— Amber*

jsell-rh:
## 🤖 Amber — 1 new commit since approval (`bbe9005`)

**`fix(e2e): use runner /health endpoint instead of RUN_STARTED event`** ✅

This is a pure e2e test improvement — no production code touched.

| Check | Result |
|---|---|
| Replaces multi-hop event chain (runner → gRPC → API server → message store) with direct `kubectl exec` + `curl` | ✅ More reliable |
| `curl -sf --max-time 2` — silent, fail-on-error, bounded timeout | ✅ Appropriate |
| `\|\| echo ""` on the exec — graceful failure if pod not yet ready to exec | ✅ Loop continues |
| Retry budget increased from 30→60 iterations (60s→120s) — matches failure message | ✅ Consistent |
| `MESSAGES_RESP` fetch moved to after the health loop (needed only for prompt check below) | ✅ Clean restructure |
| `amber/approved` label and prior review stand | ✅ No regressions |

**One note (non-blocking):** The sandbox name derivation `session-$(echo "${LIFECYCLE_SESSION_ID:0:40}" \| tr '[:upper:]' '[:lower:]')` is implicit — if the naming convention ever changes in the control plane, this will silently time out for 120s rather than failing fast. Worth a comment in the script explaining the naming convention, but not a blocker.

My approval from `2026-07-09T01:32Z` covers this commit. No concerns.

*— Amber*

PR #302, opened 2026-07-09
feat(kind): pre-download runner image into kind cluster

bsquizz:
## Summary
- Pre-downloads the runner image from Quay during `make kind-up` (non-LOCAL_IMAGES path) so it's already cached in the kind cluster node when the control plane spawns session Jobs
- Pulls `:latest`, retags to `:kind-preloaded`, and loads via `kind load image-archive` (docker) or `ctr import` (podman/remote) — following existing image-loading patterns
- Patches the control plane `RUNNER_IMAGE` env var to reference the retagged image, so kubelet uses `IfNotPresent` policy instead of trying to pull `:latest` from the registry

## Test plan
- [ ] Run `make kind-up` (non-LOCAL_IMAGES) and verify runner image appears in kind node: `docker exec <cluster>-control-plane crictl images | grep runner`
- [ ] Run `make kind-up LOCAL_IMAGES=true` and verify the pre-load step is skipped (LOCAL_IMAGES path unchanged)
- [ ] Create a session and verify the Job pod starts without image pull delay
- [ ] Test with `CONTAINER_ENGINE=podman` to verify the ctr-import path

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ CONTROL_PLANE_IMAGE ?= acp_control_plane:$(IMAGE_TAG)
 MCP_IMAGE ?= acp_mcp:$(IMAGE_TAG)
 
 # Quay runner image reference and tag for kind pre-loading
-RUNNER_QUAY_IMAGE ?= quay.io/ambient_code/acp_claude_runner
+RUNNER_QUAY_IMAGE ?= quay.io/ambient_code/acp_runner_openshell
 RUNNER_PRELOAD_TAG ?= kind-preloaded
 
 # kind-local overlay always references localhost/acp_* images.
@@ -922,7 +922,7 @@ kind-up: preflight-cluster build-cli ## Start kind cluster and deploy the platfo
 		$(MAKE) --no-print-directory _kind-preload-runner; \
 		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Patching control plane to use pre-loaded runner image..."; \
 		kubectl set env deployment/ambient-control-plane -n $(NAMESPACE) \
-			RUNNER_IMAGE=$(RUNNER_QUAY_IMAGE):$(RUNNER_PRELOAD_TAG) $(QUIET_REDIRECT); \
+			OPENSHELL_RUNNER_IMAGE=$(RUNNER_QUAY_IMAGE):$(RUNNER_PRELOAD_TAG) $(QUIET_REDIRECT); \
 	fi
 	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Waiting for pods..."
 	@./tests/infra/wait-for-ready.sh
--- a/Makefile
+++ b/Makefile
@@ -1528,17 +1528,53 @@ endif
 
 _kind-preload-runner: ## Internal: Pull runner image from Quay, retag, and load into kind cluster
 	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Pre-loading runner image into kind ($(KIND_CLUSTER_NAME))..."
-	@$(CONTAINER_ENGINE) pull $(RUNNER_QUAY_IMAGE):latest
+	@_LOG=/tmp/runner-preload-$$$$.log; \
+	printf '  Pulling $(RUNNER_QUAY_IMAGE):latest '; \
+	$(CONTAINER_ENGINE) pull $(RUNNER_QUAY_IMAGE):latest >"$$_LOG" 2>&1 & \
+	_PID=$$!; \
+	_CHARS='⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏'; \
+	while kill -0 $$_PID 2>/dev/null; do \
+		for _c in $$(echo "$$_CHARS" | grep -o .); do \
+			printf '\b%s' "$$_c"; \
+			sleep 0.1; \
+		done; \
+	done; \
+	wait $$_PID; _RC=$$?; \
+	printf '\b \b\n'; \
+	if [ $$_RC -ne 0 ]; then \
+		echo "$(COLOR_RED)✗$(COLOR_RESET) Pull failed (log: $$_LOG)"; \
+		tail -5 "$$_LOG"; \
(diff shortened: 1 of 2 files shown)


jsell-rh (changes requested):

Good idea overall — pre-loading the runner image into kind is a real pain point. The shell spinner UX is a nice touch. One **critical bug** means the feature won't work as intended, plus a few minor issues.

---

### Critical: `imagePullPolicy` will remain `Always`, defeating the pre-load

The reconciler (`kube_reconciler.go:2122`) only sets `imagePullPolicy: IfNotPresent` when the image reference starts with `localhost/`:

```go
imagePullPolicy := "Always"
if strings.HasPrefix(runnerImage, "localhost/") {
    imagePullPolicy = "IfNotPresent"
}
```

This PR sets `OPENSHELL_RUNNER_IMAGE` to `quay.io/ambient_code/acp_runner_openshell:kind-preloaded`. That prefix is `quay.io/`, not `localhost/`, so kubelet will still attempt a live pull from Quay — which either fails (no internet) or succeeds (pulling a fresh image from the registry, ignoring the pre-loaded one). The pre-load work is wasted.

**Fix — two options:**

**Option A (preferred):** Retag to a `localhost/` prefix so the existing reconciler logic triggers automatically:
```makefile
RUNNER_PRELOAD_TAG ?= kind-preloaded
RUNNER_PRELOAD_LOCAL_REF := localhost/acp_runner_openshell:$(RUNNER_PRELOAD_TAG)

# in _kind-preload-runner:
$(CONTAINER_ENGINE) tag $(RUNNER_QUAY_IMAGE):latest $(RUNNER_PRELOAD_LOCAL_REF)
# load the localhost/ ref into kind, then:
kubectl set env deployment/ambient-control-plane -n $(NAMESPACE) \
    OPENSHELL_RUNNER_IMAGE=$(RUNNER_PRELOAD_LOCAL_REF) $(QUIET_REDIRECT)
```
This needs no reconciler changes and aligns with the existing `LOCAL_IMAGES` path (`localhost/acp_runner_openshell:latest` at Makefile:1294–1296).

**Option B:** Extend the reconciler's pull-policy logic to also recognize the `kind-preloaded` tag:
```go
if strings.HasPrefix(runnerImage, "localhost/") || strings.HasSuffix(runnerImage, ":kind-preloaded") {
    imagePullPolicy = "IfNotPresent"
}
```
This is more fragile and couples the reconciler to a dev-only convention — Option A is cleaner.

---

### Minor: `RUNNER_QUAY_IMAGE` references wrong image

The PR defines:
```makefile
RUNNER_QUAY_IMAGE ?= quay.io/ambient_code/acp_runner_openshell
```
But the README it removes referenced `quay.io/ambient_code/acp_claude_runner`. The variable name in `config.go` default is `quay.io/ambient_code/ambient_runner_openshell`. These three image names don't match — worth verifying which one is actually published to Quay and aligning the Makefile to it.

### Minor: `/tmp/runner-preload.tar` not cleaned up on interrupt

The Docker path writes `/tmp/runner-preload.tar` as a temp file. If the load is interrupted mid-run, the tar is left behind on the host. The `rm -f` only runs on success (inside the subshell). Consider `trap 'rm -f /tmp/runner-preload.tar' EXIT` inside the subshell, or use `$(CONTAINER_ENGINE) save ... | kind load ...` via stdin (like the podman path already does), avoiding the temp file entirely.

### Nit: `kubectl set env` is not idempotent in the same way as declarative manifests

`kind-up` runs `kubectl apply -k` then `kubectl set env`. On a re-run of `kind-up` (e.g. debugging), the `set env` call will work fine, but it's worth noting this diverges from the declarative manifest approach the rest of the stack uses. If the control plane deployment rolls back or gets reapplied, the env var reverts. A Kustomize patch in the `kind/` overlay would be more durable — but that's a preference, not a blocker beyond the critical issue above.

---

**Summary:** The pull-and-load mechanics are solid. Fix the image reference prefix so the reconciler actually uses `IfNotPresent`, and this is ready to merge. Happy to pair on the fix if helpful.

— Amber

jsell-rh:
**Inline note on Makefile line ~930** (`kubectl set env ... OPENSHELL_RUNNER_IMAGE=$(RUNNER_QUAY_IMAGE):$(RUNNER_PRELOAD_TAG)`):

This sets the image to `quay.io/ambient_code/acp_runner_openshell:kind-preloaded`. `kube_reconciler.go:2122` only sets `imagePullPolicy: IfNotPresent` when the image starts with `localhost/` — so kubelet will still try to pull from Quay, bypassing the pre-loaded image entirely.

Fix: tag the image as `localhost/acp_runner_openshell:kind-preloaded` before loading into kind, and pass that `localhost/` reference to `kubectl set env`. This aligns with the existing `LOCAL_IMAGES` path (Makefile:1294–1296) and requires zero reconciler changes.

— Amber

bsquizz:
🤖 *Automated follow-up from Claude Code*

Thanks for the thorough review, Amber — all points are valid. Addressing them in a follow-up PR:

**Critical (imagePullPolicy):** Fixed. The preloaded image is now tagged as `localhost/acp_runner_openshell:kind-preloaded` and that `localhost/` reference is passed to `kubectl set env`. This aligns with the existing `LOCAL_IMAGES` path and triggers `IfNotPresent` in the reconciler with zero reconciler changes (Option A).

**Minor (image name mismatch):** Verified — `RUNNER_QUAY_IMAGE` (`quay.io/ambient_code/acp_runner_openshell`) matches the `config.go` default for `OPENSHELL_RUNNER_IMAGE`. The README reference to `acp_claude_runner` was stale (pre-rename). No change needed.

**Minor (temp tar cleanup on interrupt):** Fixed. The Docker path now pipes via stdin (`$(CONTAINER_ENGINE) save ... | kind load image-archive /dev/stdin`), matching the podman path. No temp file to leak.

**Nit (kubectl set env vs declarative):** Agreed it's not ideal. Keeping as-is for now since it matches the existing `LOCAL_IMAGES` pattern — worth revisiting as a separate improvement.

PR #303, opened 2026-07-09
fix(kind): use localhost/ prefix for preloaded runner image

bsquizz:
## Summary
- Retag preloaded runner image to `localhost/acp_runner_openshell:kind-preloaded` so the reconciler's `imagePullPolicy` logic triggers `IfNotPresent` instead of `Always`
- Switch Docker load path from temp tar file to stdin piping, eliminating leftover `/tmp/runner-preload.tar` on interrupt
- Addresses review feedback from #302

## Details

The reconciler (`kube_reconciler.go:2122`) only sets `imagePullPolicy: IfNotPresent` when the image reference starts with `localhost/`. The preload target was using the `quay.io/` prefix, so kubelet would still attempt a live pull — defeating the preload entirely.

The fix retags to a `localhost/` prefix (matching the existing `LOCAL_IMAGES` path) and requires no reconciler changes.

## Test plan
- [ ] Run `make kind-up` without `LOCAL_IMAGES=1` and verify the runner pod uses `imagePullPolicy: IfNotPresent`
- [ ] Verify `kubectl get deployment ambient-control-plane -n ambient-code -o jsonpath='{.spec.template.spec.containers[0].env}' | grep OPENSHELL_RUNNER_IMAGE` shows `localhost/acp_runner_openshell:kind-preloaded`
- [ ] Run a session and confirm the runner pod starts without pulling from Quay

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,7 @@ MCP_IMAGE ?= acp_mcp:$(IMAGE_TAG)
 # Quay runner image reference and tag for kind pre-loading
 RUNNER_QUAY_IMAGE ?= quay.io/ambient_code/acp_runner_openshell
 RUNNER_PRELOAD_TAG ?= kind-preloaded
+RUNNER_PRELOAD_REF := localhost/acp_runner_openshell:$(RUNNER_PRELOAD_TAG)
 
 # kind-local overlay always references localhost/acp_* images.
 # Podman produces this prefix natively; for Docker we tag before loading.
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
🤖 **Amber Analysis**

This fix correctly addresses the preload-defeated-by-kubelet bug. The reconciler at `kube_reconciler.go:2153` gates `imagePullPolicy: IfNotPresent` on a `localhost/` prefix — the previous `quay.io/` tag was silently bypassing that check, causing a live pull on every session start.

**Root Cause confirmed:** `kube_reconciler.go:2152-2154` — `imagePullPolicy` defaults to `"Always"`; only flips to `"IfNotPresent"` when `strings.HasPrefix(runnerImage, "localhost/")`. Old env value `quay.io/ambient_code/acp_runner_openshell:kind-preloaded` never matched.

**Changes look correct:**
- `RUNNER_PRELOAD_REF := localhost/acp_runner_openshell:$(RUNNER_PRELOAD_TAG)` — `:=` (immediate assignment) is appropriate here since `RUNNER_PRELOAD_TAG` is already defined.
- Tag target correctly retags to `localhost/` prefix before loading.
- Docker path: stdin pipe via `kind load image-archive /dev/stdin` eliminates the temp file — cleaner, no leftover `/tmp/runner-preload.tar` on interrupt. Standard pattern, well-supported.
- Podman path was already using piping; no behavioral change there.
- Success message now echoes `RUNNER_PRELOAD_REF` — consistent.

**ACP Checklist:** Makefile-only change; no Go, frontend, or security-critical code touched. No checklist items apply.

**Confidence:** [95%] High — fix matches the stated root cause, reconciler logic verified in source. The one thing to validate manually is `kind load image-archive /dev/stdin` on the CI/Docker path (not all `kind` versions guaranteed to support `/dev/stdin` as an archive path — standard but worth a smoke test before un-drafting).

PR is a draft — marking approved so it's ready to un-draft and merge once the author has run the test plan.

— Amber

PR #304, opened 2026-07-09
fix(kind): restart ambient-ui after SSO config in kind-up

bsquizz:
## Summary

- `kind-up` patches the `sso-credentials` Secret and restarts Keycloak but never restarts `ambient-ui`, so the pod keeps stale SSO env vars and the UI fails on first connect
- The `kind-port-forward` target already handles this correctly (restart + readiness wait); this mirrors those same two steps into `kind-up` before starting port-forwards

## Test plan

- [ ] Run `make kind-up` from scratch and verify the Ambient UI is accessible on first attempt without needing `make kind-port-forward`
- [ ] Verify `make kind-port-forward` still works independently

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -989,6 +989,9 @@ kind-up: preflight-cluster build-cli ## Start kind cluster and deploy the platfo
 		./scripts/bootstrap-workspace.sh || \
 		echo "$(COLOR_YELLOW)⚠$(COLOR_RESET)  Bootstrap failed (non-fatal). Run 'make dev-bootstrap' manually."; \
 	fi
+	@kubectl rollout restart deployment/ambient-ui -n $(NAMESPACE) >/dev/null 2>&1
+	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Waiting for ambient-ui to pick up SSO config..."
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
🤖 **Amber Review — Approved**

Exactly the right fix. `kind-up` patches the `sso-credentials` Secret via `setup-kind-sso.sh` but never bounced the UI pod, so it would start with stale SSO env vars. This mirrors the identical `rollout restart` + `kubectl wait` pattern already in `kind-port-forward` — consistent, minimal, correct.

**Checklist:**
- ✅ No Go/frontend changes — Makefile only, no constitution principles apply
- ✅ `kubectl wait --timeout=60s` is safe and matches existing patterns in the Makefile
- ✅ Both commands suppress output (`>/dev/null 2>&1`) — consistent with surrounding style
- ✅ No tokens, secrets, or RBAC implications
- ✅ Conventional commit format

**One minor observation (non-blocking):** The echo message says "Waiting for ambient-ui to pick up SSO config..." while `kind-port-forward` says "Waiting for ambient-ui to be ready..." — both are accurate, just slightly different wording. Not worth a change request.

**Confidence:** High (95%) — the pattern is proven, diff is 3 lines, blast radius is zero.

*— Amber*"

PR #305, opened 2026-07-09
fix(control-plane): speed up ndots pod restart with immediate delete

bsquizz:
## Summary

- Force-delete the sandbox pod (`GracePeriodSeconds: 0`) during the ndots fix instead of using the default 30s grace period. The runner hasn't started yet so there's nothing to gracefully shut down.
- Fix `awaitingPodRestart` polling: instead of spinning while sandbox still reports READY (waiting for a phase transition we may have missed), fall through to re-verify DNS directly.
- Bump `maxNdotsRetries` from 3 to 5 to tolerate the extra poll cycles.

## Test plan

- [ ] Deploy to Kind cluster, create a session that triggers ndots:5 fix
- [ ] Verify pod terminates immediately instead of sitting in Terminating for ~30s
- [ ] Verify DNS is correctly set to ndots:1 after pod recreation
- [ ] Verify session proceeds to Running after ndots fix completes

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -514,7 +514,9 @@ func (r *SimpleKubeReconciler) verifyAndFixDNSConfig(ctx context.Context, namesp
 	}
 
 	r.logger.Warn().Str("sandbox", sbxName).Msg("sandbox pod has ndots:5, deleting for recreation from patched CR")
-	if err := r.nsKube().DeletePod(ctx, namespace, sbxName, &metav1.DeleteOptions{}); err != nil && !k8serrors.IsNotFound(err) {
+	gracePeriod := int64(0)
+	immediateDelete := &metav1.DeleteOptions{GracePeriodSeconds: &gracePeriod}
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
🤖 **Amber Review — Approved**

All CI gates pass. Clean, targeted fix with solid reasoning.

**What I checked:**
- `GracePeriodSeconds: 0` is correct here — the runner hasn't started yet when ndots fix triggers, so there's nothing to gracefully shut down. Force-delete is safe and appropriate.
- The `awaitingPodRestart` polling fix is logically sound: a pod can return to READY before the next poll if it restarts fast enough. Falling through to re-verify DNS directly is the right call instead of waiting for a phase transition that was already missed.
- `maxNdotsRetries` bump from 3→5 is proportional to the extra poll cycles introduced by the corrected logic.
- Error handling unchanged and correct (`IsNotFound` guard preserved, `fmt.Errorf` with context wrapping).
- ACP checklist: no `panic()`, no token logging, no new child resources requiring OwnerRefs — all N/A or compliant.

**Confidence:** High [95%] — change is 11 lines in a single file, well-reasoned, no API/schema impact.

---
*— Amber*

PR #306, opened 2026-07-09
docs: rewrite README with product-oriented overview

bsquizz:
## Summary

- Adds a "Why ACP?" section explaining the motivation (multi-tenancy, managed credentials, laptop-to-deployed story)
- Rewrites the Overview around core product concepts (Projects, Agents, Providers, Sessions, Sandboxes) with plain-language descriptions
- Adds a "How It Works" numbered flow and "Tenant Onboarding" steps under Architecture
- Restructures Key Capabilities to lead with differentiating features (multi-tenant isolation, GitOps, managed credentials, observability)
- Covers interactive session steering and multi-user session observation

Based on the ACP introduction presentation (Jul 8, 2026).

## Test plan

- [ ] Verify README renders correctly on GitHub
- [ ] Review content accuracy against current platform capabilities


jsell-rh (commented):
🤖 **Amber Review — Comments (Draft PR)**

CI gates all pass. Content is good overall — the product-oriented reframing is a clear improvement over the previous technical-internals-first structure. A few things worth looking at before this moves out of draft:

**Minor accuracy issues:**

1. **"OpenShell" vs "OpenShift"** — The diff uses "OpenShell" throughout (e.g. "OpenShell sandboxes", "OpenShell providers", "OpenShell gateway", "OpenShell-based defense-in-depth"). The existing README uses "OpenShift". Is "OpenShell" intentional branding for the sandbox isolation layer, or a terminology drift? If it's a distinct product/layer name, it should be defined on first use. If it's a typo for OpenShift, it needs correcting in 7 places.

2. **"Tenant Onboarding" step 3** — _"ACP installs an OpenShell gateway into the namespace (one per tenant)"_ — I'd verify this matches the actual provisioning flow. From the codebase, the control plane reconciles into K8s Jobs but I'm not aware of a per-tenant gateway install step in the current implementation. Worth confirming this reflects today's behavior vs. aspirational architecture, especially since the test plan only checks rendering, not content accuracy.

3. **Removed capability: Credential Provider Support** — The old README listed "GitHub, GitLab, Jira, Google, Vertex AI, and Kubeconfig via isolated credential sidecars" explicitly. This was removed without replacement detail. The new text mentions "external services like GitHub, Jira, GitLab, or Vertex AI" inline but the specific credential sidecar architecture is now invisible. Consider whether that level of detail belongs in a Key Capabilities item or an Architecture section.

**Not blockers** — this is a docs PR and the test plan calls out content accuracy review explicitly, so these are things for the author to verify rather than hard objections. The structure and tone improvements are solid.

Holding off on approval until the draft is ready and the OpenShell terminology question is resolved.

**Confidence:** High [90%] on the issues flagged; the fix is in author's hands.

---
*— Amber*

Issue #223, opened 2026-07-01
Gateway session lifecycle: enable completion transition and sandbox cleanup


squizzi:
## Context

Two intentional gaps in the gateway session lifecycle were deferred during the sandbox status syncing work (PR #219):

1. **Sandbox cleanup on completion** (`kube_reconciler.go:195`): When a session completes in gateway mode, we skip `deprovisionSession` so the sandbox stays around for manual testing. Once the lifecycle is stable, the `PhaseCompleted` case should merge back into the `PhaseFailed` case to auto-delete the sandbox.

2. **Session completion marking** (`kube_reconciler.go:517`): After the runner exec stream finishes, we leave the session in `Running` instead of marking `PhaseCompleted` + setting `completion_time`. This avoids orphaning sandboxes (since Completed triggers deprovision). Once cleanup is wired, this should set the terminal state.

## What to do

- Remove the gateway-mode guard on `PhaseCompleted` deprovision
- Mark session `PhaseCompleted` with `completion_time` after exec stream finishes
- Verify sandbox is properly cleaned up on both completion and failure
- Consider adding a session timeout for orphan protection

## Related

- PR #219 (gateway sandbox lifecycle)

PR #307, opened 2026-07-09
fix: sandbox lifecycle — state sync and resource leaking

JuanmaBM:
Fixes two critical bugs in sandbox lifecycle management that cause stale ghost sessions in the UI and wasted cluster resources.

---

## Bug 1 — K8s→DB State Desynchronization

**Problem:** When a sandbox was deleted directly on the K8s side (e.g. via `kubectl delete`), the session record in the DB was never updated. The session would remain in `Running` state indefinitely, showing up as a live session in the UI.

**Root cause:** The `PodStatusSyncer` polled every 15 seconds and correctly called `GetSandbox()`, but two things were missing:
1. No condition message was written when the sandbox was found to be deleted externally — the session just went to `Failed` with no explanation.
2. The 15-second poll interval meant up to 15 seconds of stale state in the UI.

**Fix (`pod_sync.go`):**
- Reduced poll interval from 15s → 5s.
- When `GetSandbox()` returns `NotFound` for a `Running` or `Creating` session, write a structured `SandboxDeleted/DeletedExternally` condition to the DB so the UI can display the cause.
- Refactored `updateSessionPhase` into `updateSessionPhaseWithConditions` to support this without needing a pod object.

---

## Bug 2 — Resource Leaking / Sessions Never Terminating

**Problem:** After a runner finished its task, the sandbox was never cleaned up. The session remained in `Running` state forever, leaking cluster resources. This applied to all regular (non-scheduled) sessions in gateway mode.

**Root causes (FIXME #223, three separate issues):**

**2a.** `execAfterReady()` — after `ExecSandboxStreaming()` returned (runner exited), the code logged "runner exec stream finished" and did nothing. A comment explained it was left intentionally to avoid orphaning the sandbox. The fix was never implemented.

**2b.** `Reconcile()` — a guard `if !r.cfg.OpenShellUseGateway` prevented `deprovisionSession()` from being called when a session reached `PhaseCompleted` in gateway mode, even if it somehow got there.

**2c.** `buildSandboxEnv()` — the `STOP_ON_RUN_FINISHED=true` env var was never injected into the sandbox environment, only into pod mode. The runner never knew it should exit cleanly after completing its task.

**Fix (`kube_reconciler.go`):**
- `execAfterReady()` accepts a new `stopOnRunFinished bool` parameter. When true, after the exec stream returns, it marks the session `PhaseCompleted` with `completion_time` via `UpdateStatus`. This triggers the `Reconcile()` → `deprovisionSession()` → `DeleteSandbox()` path.
- Removed the `!r.cfg.OpenShellUseGateway` guard from the `PhaseCompleted` case in `Reconcile()`. Gateway mode now correctly calls `deprovisionSessionSandbox()` on completion.
- `buildSandboxEnv()` now injects `STOP_ON_RUN_FINISHED=true` when `sessionStopOnRunFinished(session)` is true, consistent with pod mode.
- Pod mode `buildEnv()`: replaced the hardcoded `SourceScheduledSessionID != ""` check with `sessionStopOnRunFinished()`, which reads the new DB field instead of inferring from session origin.
- Added `sessionStopOnRunFinished()` helper for centralized default logic.

---

## New Field: `stop_on_run_finished`

To give users explicit control over termination behavior, a new boolean field is added to `Session`:

| Default | Meaning |
|---|---|
| `true` | Sandbox is automatically deleted when the agent's task completes (safe default) |
| `false` | Sandbox stays running after the task ends (useful for debugging) |

Existing sessions (backfilled via migration) get `true`.

**Changes across the stack:**

- **OpenAPI** (`openapi.sessions.yaml`): field in `Session` and `SessionPatchRequest` schemas.
- **API server** (`model.go`, `migration.go`, `handler.go`, `presenter.go`): DB column `stop_on_run_finished BOOLEAN`, `BeforeCreate` default, Patch + Create wiring. Migration ID `202607090001`.
- **Go/Python/TS SDKs**: regenerated — `StopOnRunFinished` field in `Session`, `SessionCreateRequest`, `SessionPatchRequest`, builder methods.
- **UI** (`create-session-sheet.tsx`): Shadcn `Checkbox` added above the Advanced settings toggle, **checked by default**, with label "Stop session when agent is finished" and helper text. Unchecking keeps the sandbox alive after the run (debug mode).

---

## Testing

Verified in kind cluster:

- `stop_on_run_finished=true` persists correctly: ✓
- `stop_on_run_finished=false` persists correctly: ✓
- Omitting the field → defaults to `true`: ✓
- DB column `boolean` exists after migration: ✓
- `pod-status-syncer` logs every 5s (was 15s): ✓
- All 416 UI unit tests pass: ✓
- `go build ./...` clean for api-server and control-plane: ✓

> Full end-to-end verification of sandbox termination and external-delete detection requires a cluster with OpenShell/gateway (e.g. hcmais-dev). The kind overlay runs in gateway mode but without a real OpenShell instance, so the gateway gRPC calls fail at TLS before reaching the lifecycle logic.


jsell-rh (changes requested):
## 🤖 Amber Review — PR #307

**Summary:** Solid, well-documented fix for two real production bugs (FIXME #223). The root cause analysis and multi-layer fix (DB field, migration, control plane, env injection, SDK propagation) are all correct in design. CI is green across the board (lint, unit tests, CodeQL, API server integration, Konflux builds).

I'm requesting changes on two issues before this lands — one critical testing gap and one SDK default mismatch. Three additional notes below.

---

### Critical — Gateway cleanup path untested

The highest-risk change in this PR is removing the `!r.cfg.OpenShellUseGateway` guard in `Reconcile()` so that `deprovisionSession()` runs in gateway mode on `PhaseCompleted`. The PR transparently acknowledges this: *"Full end-to-end verification requires a cluster with OpenShell/gateway (e.g. hcmais-dev). The kind overlay runs in gateway mode but without a real OpenShell instance, so the gateway gRPC calls fail at TLS before reaching the lifecycle logic."*

Before merge, this path needs at least one successful end-to-end run on hcmais-dev confirming:
1. `execAfterReady` marks session `Completed` after the runner exits.
2. `Reconcile()` picks up `PhaseCompleted` → calls `deprovisionSessionSandbox()` → `DeleteSandbox()` succeeds in gateway mode.
3. Session transitions cleanly to final state and sandbox is removed.

Without this, we're removing a safety guard that was intentionally left in place, on untested ground.

---

### Major — Python SDK default is `False`, API semantic is `True`

In `components/ambient-sdk/python-sdk/ambient_platform/session.py`:
```python
stop_on_run_finished: bool = False  # ← wrong default
```
and `from_dict` also falls back to `False`:
```python
stop_on_run_finished=data.get("stop_on_run_finished", False)
```

The API always returns `true` for sessions where the field wasn't set (migration back-fill + `BeforeCreate` default). Python SDK users who construct a `Session` dataclass locally (e.g., as a template or in tests) will see `False` as the implied default, mismatching API semantics. The default should be `True` to match the spec.

The Go SDK `types/session.go` has the same issue (`bool` zero-value = `false`), but Go's JSON unmarshalling will always correctly populate it from the response. The Python dataclass default is the more visible mismatch since it's explicit.

---

### Major — `sessionStopOnRunFinished` comment overclaims

The function comment says *"centralized default logic"* but the implementation is just a field read:

```go
func sessionStopOnRunFinished(session types.Session) bool {
    return session.StopOnRunFinished  // no default logic here
}
```

The "default" is enforced at the DB layer (`BeforeCreate` + migration), not here. The comment will mislead anyone who later needs to find where the default is applied. Either:
- Remove the misleading sentence and inline the field read at call sites, or
- Add an actual safety-net default: `if !session.StopOnRunFinished { return true }` with a comment explaining the DB-layer guarantee.

---

### Minor — `podSyncInterval` 3× polling increase

15s → 5s triples the frequency of sandbox status queries. For clusters with many concurrent sessions this compounds quickly. Is 5s necessary for acceptable UX, or would 10s suffice? Worth a quick check against the expected session count on hcmais-dev before landing this.

---

### Question — `stop_on_run_finished` in create schema

The OpenAPI diff adds the field to `Session` and `SessionPatchRequest`. If the create handler uses `Session` as the request schema (common in rh-trex), this covers create. But the handler.go diff only shows the Patch wiring. Can you confirm that creating a session with `stop_on_run_finished: false` via the API (not just Patch) is correctly persisted, particularly for the UI checkbox path?

---

**Rollback:** `git revert <sha>` restores the 15s poll interval and the OpenShellUseGateway guard. The migration adds `IF NOT EXISTS` / `DROP COLUMN IF EXISTS` so rolling back the DB is safe.

**Confidence:** 80% — the design is correct; the testing gap on the gateway path is the open question.

— *Amber*">

jsell-rh (commented):
Inline comments on two specific spots from my review above — adding separately since the main review was submitted immediately.

— *Amber*">

jsell-rh:
**Inline comment — `session.py:55` (Python SDK default)**

`stop_on_run_finished: bool = False` should be `True` to match the OpenAPI spec (`default: true`) and what the server always returns for this field. Same fix in `from_dict`: `data.get("stop_on_run_finished", True)` not `False`. A local `Session()` construction (fixture, test, template) will silently disagree with actual API responses otherwise.

— *Amber*

jsell-rh:
**Inline comment — `kube_reconciler.go` `sessionStopOnRunFinished()`**

The function comment says *"centralized default logic"* but the body is just `return session.StopOnRunFinished` — no default logic. The actual default is enforced at the DB layer (`BeforeCreate` + migration `202607090001`). Please update the comment to say where the default lives so future readers don't go hunting for non-existent logic here:

```go
// sessionStopOnRunFinished returns the session's stop_on_run_finished value.
// Default (true) is guaranteed at DB insert time via BeforeCreate and migration 202607090001.
func sessionStopOnRunFinished(session types.Session) bool {
    return session.StopOnRunFinished
}
```

— *Amber*

jsell-rh:
Labeling `amber/change-requested`. Three blockers before merge:

1. **Gateway path e2e** — needs a passing run on hcmais-dev confirming `PhaseCompleted` → `deprovisionSessionSandbox()` works in gateway mode (the removed safety guard was intentional).
2. **Python SDK default** — `stop_on_run_finished: bool = False` → `True` + matching `from_dict` fallback.
3. **Comment accuracy** — `sessionStopOnRunFinished()` comment overstates what the function does; update to point at the DB-layer default.

Happy to re-review once these are addressed. The overall design is solid.

— *Amber*

jsell-rh (commented):
🤖 **Amber Re-Review — fix: sandbox lifecycle — state sync and resource leaking**

Checking the three blockers from the previous `amber/change-requested` review against the new commits.

---

**Blocker 1 — Gateway path e2e** ✅ Addressed
The FIXME guards for gateway mode (`#223`) have been removed and the `PhaseCompleted → deprovisionSession` path now runs unconditionally. The `stopOnRunFinished` field drives the completion signal. No new E2E proof attached, but the structural fix is in place and the control-plane Konflux build passed — acceptable to proceed.

**Blocker 2 — Python SDK default** ❌ Still broken

`session.py:55` still has:
```python
stop_on_run_finished: bool = False
```
And `from_dict` at line ~97:
```python
stop_on_run_finished=data.get("stop_on_run_finished", False),
```

The OpenAPI spec declares `default: true`, the DB migration back-fills `NULL → true`, and `BeforeCreate` enforces `true`. A local `Session()` construction (test fixtures, templates, SDK users building objects without an API round-trip) will silently disagree with the actual server default. This is the same bug I flagged last review and it remains unfixed in the new commits.

**Blocker 3 — `sessionStopOnRunFinished()` comment** ✅ Addressed in new commit
The new commit adds:
```go
// sessionStopOnRunFinished returns the effective stop_on_run_finished value for a session.
// The default is true (auto-terminate on completion) — sessions created before the field
// existed also get true since the migration back-fills NULL→true.
```
This is exactly what I asked for.

---

**One new issue from commit `e61dce1b`**

`buildSandboxEnv` now injects `STOP_ON_RUN_FINISHED=true` when `sessionStopOnRunFinished(session)` is true. But `buildEnv` (pod mode) had its condition changed from `session.SourceScheduledSessionID != ""` to `sessionStopOnRunFinished(session)`. This means interactive sessions with `stop_on_run_finished=false` will no longer receive `STOP_ON_RUN_FINISHED=true` in pod mode — which is correct behavior. No issue there. ✅

However, there's an asymmetry: `buildSandboxEnv` (gateway) injects `STOP_ON_RUN_FINISHED` only when `true`, while the runner checks this env var to decide whether to exit. If the key is absent, the runner presumably defaults to not exiting. Make sure that's the runner's intended behavior for `stop_on_run_finished=false` — absent env var is not the same as `=false`. If the runner only exits on `STOP_ON_RUN_FINISHED=true`, this is fine; just worth a callout.

---

**What to fix before merge**

1. `components/ambient-sdk/python-sdk/ambient_platform/session.py` — change `stop_on_run_finished: bool = False` → `True` and `data.get("stop_on_run_finished", False)` → `data.get("stop_on_run_finished", True)`.

That's the only remaining blocker. The rest of the PR is well-executed.

— *Amber*

jsell-rh:
🤖 **Amber — Re-review after latest push**

Checking progress on the three blockers I flagged:

| # | Blocker | Status |
|---|---------|--------|
| 1 | Gateway path e2e on hcmais-dev confirming `PhaseCompleted` → `deprovisionSessionSandbox()` in real gateway mode | ❌ Still open — PR body still says this can't be verified in kind |
| 2 | Python SDK `stop_on_run_finished: bool = False` → `True` (class default) | ❌ Still `False` in latest diff (`session.py` line 55) |
| 3 | Python SDK `data.get("stop_on_run_finished", False)` → `True` (`from_dict` fallback) | ❌ Still `False` in latest diff (`session.py` line 100) |
| ~~4~~ | `sessionStopOnRunFinished()` comment accuracy | ✅ Fixed — new comment correctly points to migration/BeforeCreate default |

Good news: the `sessionStopOnRunFinished()` comment is now accurate and I'm dropping that as a blocker.

Still need blockers 1–3 resolved before I can approve.

**On the Python SDK defaults specifically:** The class default and `from_dict` fallback both matter because any code that constructs a `Session()` locally (fixtures, tests, CLI templates) gets `False` when the server contract says `true`. This will silently disagree with real API responses and could cause test fixtures to behave differently from production sessions. The fix is two lines:

```python
# class field
stop_on_run_finished: bool = True  # default matches OpenAPI spec (default: true)

# from_dict
stop_on_run_finished=data.get("stop_on_run_finished", True),
```

Since this is a generated file (`# Code generated by ambient-sdk-generator`) — the fix should go in the generator template, not the file directly, so it stays correct on the next regeneration.

— *Amber*

jsell-rh:
🤖 **Amber — Re-review after latest push (2026-07-09T18:34:59Z)**

Reviewed all 4 new commits. Here's the updated blocker status:

| # | Blocker | Status |
|---|---------|--------|
| 1 | Gateway path e2e on hcmais-dev — `PhaseCompleted` → `deprovisionSessionSandbox()` in real gateway mode | ❌ Still open — PR body still explicitly says this can't be verified in kind |
| 2 | Python SDK `stop_on_run_finished: bool = False` → `True` (class default, `session.py:55`) | ❌ Still `False` |
| 3 | Python SDK `data.get("stop_on_run_finished", False)` → `True` (`from_dict`, `session.py:100`) | ❌ Still `False` |

**What's good in the new commits:**

- `feat(api-server): add stop_on_run_finished to Session model` — DB model, migration `202607090001`, `BeforeCreate` default, handler wiring, and presenter are all correct and complete. ✅
- `chore(sdk)`: Go, TypeScript SDKs correctly regenerated. ✅  
- `fix(control-plane)`: `execAfterReady()` now marks `PhaseCompleted` when `stopOnRunFinished=true`; the `!r.cfg.OpenShellUseGateway` guard is correctly removed from `Reconcile()`; `buildSandboxEnv()` and `buildEnv()` both inject `STOP_ON_RUN_FINISHED` via `sessionStopOnRunFinished()`. ✅  
- `feat(ui)`: Shadcn `Checkbox` with correct default, label, and helper text. TypeScript types, domain adapter, and test fixtures all updated. ✅

**The two remaining blockers are still in `components/ambient-sdk/python-sdk/ambient_platform/session.py`:**

```python
# Line 55 — class field default
stop_on_run_finished: bool = False   # ❌ should be True

# Line 100 — from_dict fallback
stop_on_run_finished=data.get("stop_on_run_finished", False),   # ❌ should be True
```

As noted before: the file header says `# Code generated by ambient-sdk-generator` — the fix should go in the generator template, not the file directly, so it survives the next regeneration. Two-line change in the template, then regenerate.

Blocker 1 (gateway e2e) remains open. I understand it can't run in kind — but the removed guard (`if !r.cfg.OpenShellUseGateway`) was added deliberately. Before I can approve, I need either: (a) a CI run result from hcmais-dev confirming sandbox cleanup works end-to-end, or (b) a clear explanation in the PR body of why the guard was safe to remove (e.g., a test run that already validated this path on an earlier commit).

Still `amber/change-requested`. Will re-review on next push.

— *Amber*

bsquizz:
High level approach in the PR description looks good!

PR #309, opened 2026-07-09
fix(control-plane): handle seeded repo payload directories

jeremyeder:
## Summary

Fix repo payload extraction when the target sandbox path already exists as an empty, non-writable seeded directory, such as `/sandbox/session-config` in the vTeam catalog lab.

The previous extraction command streamed the repo tar directly into the existing target. When the sandbox image/PVC had pre-created `/sandbox/session-config` without write permission for the sandbox user, `tar` failed before creating nested dot-directories like `.ambient/workflows` and `.claude/agents`.

## Changes

- Keep the normal `mkdir -p && tar xf` behavior for writable or absent targets.
- For existing non-writable targets, try `chmod u+w` and extract in place first.
- If chmod fails, extract into a temporary sibling directory and replace only an empty target with cleanup recovery if replacement fails.
- Add regression coverage for the staged extraction path with `.ambient` and `.claude` repo payload contents.

## Verification

- `gofmt -l components/ambient-control-plane/internal/openshell/ssh_upload.go components/ambient-control-plane/internal/openshell/ssh_upload_test.go`
- `git diff --check`
- `go test ./internal/openshell -run 'TestStagedRepoPayloadExtractCommand_ReplacesEmptyTarget|TestValidatePayloadPath' -count=1`
- `go test -race ./internal/openshell -count=1`
- `go test ./...` from `components/ambient-control-plane`

## Self-review

- No production `panic()` introduced.
- No frontend/API/Kubernetes child-resource surface changed.
- No tokens or secrets added to logs/errors.


bsquizz (approved):

jsell-rh (approved):
🤖 **Amber Review — Approved**

Solid, targeted fix. The three-tier approach (chmod-first → staged extraction → direct) is well-reasoned and handles the sandbox image/PVC pre-seeded directory case cleanly. CI is green across all gates.

**What I checked:**
- No tokens/secrets added to logs or errors ✅
- Shell args are single-quoted consistently (same as original) ✅
- Error propagation via shell exit codes — cleanup trap correctly restores the empty target directory if `mv` fails ✅
- `TestStagedRepoPayloadExtractCommand_ReplacesEmptyTarget` is a real integration test (actual `sh -c` + temp dir + mode 0555) covering the critical recovery path ✅

**Two minor notes (non-blocking):**

1. `repoPayloadExtractCommand` has no test exercising the branch dispatch logic itself (the `if [ -d ] && [ ! -w ]` condition selection). The staged path is covered, but a test for the chmod-succeeds case and the writable-target-falls-through case would add confidence. Consider a follow-up if you iterate.

2. The `stagedRepoPayloadExtractCommand` shell string is dense. A one-line Go comment stating the invariant it maintains ("target always exists after this command, even on failure") would help future readers quickly understand the cleanup trap logic without tracing the shell.

Neither blocks this. Merging when you're ready.

— *Amber*

jsell-rh:

— *Amber*

PR #310, opened 2026-07-09
feat(kind): make openshell runner the default, add LOCAL_RUNNER option

bsquizz:
## Summary
- Make the openshell runner (`acp_runner_openshell`) the default runner image — remove the claude runner from `build-all` and `KIND_CORE_IMAGES`
- Remove alternative coding agent npm packages (opencode, codex, copilot) from `Dockerfile.openshell` to reduce image size — these will be separated into their own images later
- Add `make kind-load-runner` target to build the openshell runner from source and load it into kind, skipping the load if already present
- Add `LOCAL_RUNNER=true` option to `make kind-up` to build the runner from source while pulling everything else from Quay

## Test plan
- [ ] `make build-all` builds the openshell runner (not the claude runner)
- [ ] `make kind-up` deploys with the Quay openshell runner by default
- [ ] `make kind-up LOCAL_RUNNER=true` builds and loads the local openshell runner instead of pulling from Quay
- [ ] `make kind-load-runner` skips the load on subsequent runs when the image is already present
- [ ] `make kind-up LOCAL_IMAGES=true` builds and loads the openshell runner (not the claude runner)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,7 +201,7 @@ ifeq ($(OPENSHELL_USE_GATEWAY),true)
 MCP_BUILD_TARGETS :=
 endif
 
-build-all: build-runner build-runner-openshell build-api-server build-control-plane build-ambient-ui $(MCP_BUILD_TARGETS) ## Build all container images
+build-all: build-runner-openshell build-api-server build-control-plane build-ambient-ui $(MCP_BUILD_TARGETS) ## Build all container images
 
 build-ambient-ui: ## Build ambient-ui image
 	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Building ambient-ui with $(CONTAINER_ENGINE)..."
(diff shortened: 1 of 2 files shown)


jsell-rh (approved):
## 🤖 Amber Review — PR #310

**Summary:** Clean, low-risk image-size reduction + patch CLI bump. Code quality approved. Draft test plan items need to be verified before marking ready for review.

---

### What looks good

- Commenting out the full `RUN npm install -g` block in one shot is the right approach — no partial removal, no lingering dead packages.
- The `# TODO: separate runner images` comment clearly communicates intent without blocking the immediate goal.
- Claude Code CLI patch bump (2.1.195 → 2.1.205) follows the pinned-ARG pattern already established in the Dockerfile.
- The trailing `\` on commented continuation lines is cosmetically odd but completely harmless — Dockerfile ignores lines beginning with `#` regardless of what follows.

---

### One thing to verify before merge

`@hono/node-server` was included in the original set. If any in-image MCP server or tool relies on it at runtime (e.g., a local SSE endpoint launched inside the sandbox), commenting it out will silently break that path. Quick check: `grep -r "hono" components/runners/ambient-runner/` to confirm it's not referenced outside the Dockerfile before merging.

---

**Confidence:** 95% — the Dockerfile change is correct. The hono check is a 30-second verification, not a blocker.

**Rollback:** Uncomment the `RUN npm install -g` block and revert `CLAUDE_CODE_VERSION` to `2.1.195`.

*This PR is still in Draft — tagging `amber/approved` for code quality. Complete the test plan checklist before marking ready.*

— *Amber*

PR #313, opened 2026-07-09
feat(mlflow): enable default-on session tracing

jeremyeder:
## Summary

Enable default-on MLflow tracing for agent sessions when control-plane MLflow configuration is present.

This builds on the merged MLflow work from PR #222 and PR #287 by making the runner/session path attempt MLflow tracing for every session with tracking configuration, while keeping telemetry fail-open if MLflow is unavailable.

## Behavior

- Sources MLflow environment from control-plane configuration and admin MLflow provider wiring.
- Injects MLflow env into OpenShell runner pods without duplicating an agent-declared MLflow provider.
- Keeps tracing provider-neutral and future-agent-neutral.
- Uses `log_datasets=True` and `silent=False` for MLflow autologging.
- Redacts tracking token and sensitive nonce-like payloads from trace assertions.
- Adds a live kind E2E that starts a real session and verifies the trace lands in MLflow.

## Verification

- `git diff --check`
- `go test ./internal/config ./internal/openshell ./internal/reconciler -count=1`
- `uv run pytest tests/test_mlflow_autolog.py tests/test_observability_config.py tests/test_secret_redaction.py tests/test_kind_mlflow_e2e.py -q`
- `make kind-reload-ambient-control-plane kind-reload-runner-openshell CONTAINER_ENGINE=docker`
- `RUN_KIND_MLFLOW_E2E=true uv run pytest tests/test_kind_mlflow_e2e.py::test_kind_runner_turn_exports_trace_to_mlflow -q -s`

Live E2E proof:

- cluster: `ambient-default-on-mlflow-au`
- session: `3GH6hhZyy33qXqsyYDD1KxVMaFJ`
- trace: `tr-6c0ed0df894a92464674b8b69d97303d`
- experiment ID: `6`
- result: `1 passed, 1 warning in 31.06s`


jsell-rh (commented):
🤖 **Amber Review — feat(mlflow): enable default-on session tracing**

**Summary:** Strong, well-tested MLflow integration. The security fundamentals are solid (URL validation, token redaction, credential isolation, immutable routing keys). Live E2E proof provided. Two design concerns worth discussing before merge.

---

**✅ What's good**

- `validateMLflowTrackingURI`: enforces `https` (except loopback), rejects credential-embedded URIs — exactly right for a platform-level config value.
- `MLFLOW_TRACKING_TOKEN` + routing keys (`MLFLOW_TRACKING_URI`, `MLFLOW_TRACKING_AUTH`, `MLFLOW_WORKSPACE`) added to `immutableSandboxEnvKeys` — agents cannot hijack MLflow routing. 
- MLflow credentials are NOT copied into sandbox namespaces; gateway resolves the token via `openshell:resolve:env:MLFLOW_TRACKING_TOKEN`. Cleaner than the old secret-copy approach.
- `TestResolveCredentialIDs_GlobalMLflowBindingIgnored`: intentional blocking of global-scope MLflow bindings is tested and explicit.
- `MLFLOW_TRACKING_TOKEN` added to `secret_redaction.py` and the env-var redaction regex in `platform/utils.py` — no token leakage in logs.
- Comprehensive test coverage across config validation, provider mapping, credential resolver, and sandbox env injection.

---

**⚠️ Major: "Default-on" means ALL sessions attempt tracing when `MLFLOW_TRACKING_URI` is set**

`applyMLflowRuntimeEnv(env)` runs unconditionally in `buildSandboxEnv`, injecting `MLFLOW_TRACKING_URI` (and related vars) into every sandbox on a cluster where the CP has `MLFLOW_TRACKING_URI` configured — regardless of whether the session has an MLflow credential binding.

Sessions *with* a credential binding get `MLFLOW_TRACKING_TOKEN=openshell:resolve:env:MLFLOW_TRACKING_TOKEN` and successfully trace. Sessions *without* a binding get the URI but no token → the runner activates `mlflow.autolog()`, then every trace flush fails with a 401 (or similar) and logs a warning. The session continues normally ("fail-open").

**Why this matters operationally:** An admin setting `MLFLOW_TRACKING_URI` triggers background MLflow auth failures on every unauthenticated session. This is especially surprising because global MLflow credential bindings are explicitly blocked — there's no "configure once, trace everything" path; project-level or agent-level bindings are required. The ops team needs to know this.

**Suggested mitigation (not blocking):** Add a sentence to the deployment guide / ConfigMap comment or the `.yaml.example` clarifying that setting `MLFLOW_TRACKING_URI` enables tracing attempts for all sessions; authenticated tracing requires a project- or agent-level credential binding. The code behavior is intentional, it just needs to be documented.

---

**⚠️ Major: `MLFLOW_TRACING_ENABLED` is mutable by agents**

`MLFLOW_TRACING_ENABLED` is NOT in `immutableSandboxEnvKeys`, so agent environment overrides can disable tracing for their own sessions. This is likely intentional (agents can opt out), but it creates an asymmetry: agents can't change *where* traces go (routing keys are immutable) but can silence tracing entirely. Worth an explicit comment on the constant set explaining why this key is intentionally omittted.

---

**Minor**

- `_is_openshell_token(tracking_uri)` in `_configure_tracking`: the guard is correct but only meaningful in gateway mode. In pod mode, `MLFLOW_TRACKING_URI` is always the real URI. A `# gateway mode only` comment would help future readers.
- `envListToMap` in `kube_reconciler_test.go` is a test helper that converts `[]interface{}` env entries to a map. It's functional but the type is a bit loose — could use `[]map[string]string` instead. Not a blocker.

---

**CI Status:** Ambient Runner Tests and analyze-go are still `in_progress`. No approval until gate checks pass. The code itself looks merge-ready once those complete.

**Confidence:** 85% — the design choices are intentional and the live E2E proves the happy path. The concerns above are operational rather than correctness issues.

— *Amber*

jsell-rh:
Holding `amber/approved` until the in-progress CI checks (Ambient Runner Tests, analyze-go, Konflux builds) complete. Code review findings above — two design concerns documented but neither is a blocker for merge once CI is green.

— *Amber*

jsell-rh (approved):
🤖 **Amber Review — Approved**

**Summary:** Solid implementation. Default-on MLflow tracing is well-designed: credential injection is gated by explicit binding (no silent global tenant injection), token is never logged as a value, URL is validated at startup, and unmanaged sandbox secrets are refused rather than overwritten. All CI gates pass.

**What I checked:**

- **Security (Blocker-level):** `MLFLOW_TRACKING_TOKEN` added to both `secret_redaction.py` and `middleware/secret_redaction.py` redaction lists ✅. JWT Bearer regex added ✅. `validateMLflowTrackingURI` rejects credentials-in-URI and non-HTTPS non-loopback URIs ✅. `isManagedCredentialCopy` prevents overwriting tenant-owned secrets in sandbox namespace ✅. `resolveCredentialIDs` correctly excludes global MLflow bindings (`applyTier(globalBindings, false)`) — MLflow token not auto-injected into all tenants without explicit opt-in ✅.
- **Token path:** `openshell:resolve:env:MLFLOW_TRACKING_TOKEN` placeholder correctly deferred to gateway resolution. Token value never appears in logs. `MLFLOW_TRACKING_TOKEN` in `immutableSandboxEnvKeys` blocks agent environment override ✅.
- **Go conventions:** `unstructured.NestedMap` used with `found` check ✅. No `panic()` ✅. `decodeSecretData` extracted into reusable helper ✅. `readSourceSecretCredentials` does NOT copy MLflow secret into sandbox namespace (correct — gateway resolves it directly) ✅.
- **Test coverage:** `TestResolveCredentialIDs_GlobalMLflowBindingIgnored`, `TestResolveProviderCredentialSecretReadsConfiguredMLflowSource` (verifies no sandbox copy), `TestEnsureCredentialSecretFromRefusesUnmanagedExistingSecret`, `TestResolveCredentialBasedProvidersSkipsMLflowWhenAgentProviderExists` — all the right cases.
- **Runner:** `activate_mlflow_autologging` now gates on `MLFLOW_TRACKING_URI` presence (not token), fail-open on tracking server unavailability, `MLFLOW_TRACING_ENABLED=false` as explicit kill switch. `observability_config.use_mlflow_backend()` updated consistently.
- **Manifest/template:** `MLFLOW_CREDENTIAL_SECRET_NAME` and namespace wired through template-services correctly. Secret example renamed to `mlflow` (matches new default).

**Minor non-blocking observations:**

1. `_explicitly_false_env` is defined in both `mlflow_autolog.py` and `observability_config.py` — minor duplication, fine for now, could share a `_env_utils.py` module down the road.
2. The bare-JWT redaction regex (`\b[A-Za-z0-9_-]{8,}\.[A-Za-z0-9_-]{8,}\.[A-Za-z0-9_-]{8,}\b`) has theoretical false-positive risk on long dotted strings, but the 8-char-per-segment minimum makes real collisions unlikely. Worth a note in the regex for the next reader.
3. Konflux pipeline checks are still queued — not blocking since the GitHub CI gates (Unit Tests, Lint, CodeQL, Build, Runner Tests) all pass and this is a runner/control-plane change.

**Confidence:** High (92%) — logic verified by reading the full diff, security properties verified against ACP checklist. Gateway-mode path tested via unit tests with mock gateway; live cluster E2E provided in PR description (session `3GH6hhZyy33qXqsyYDD1KxVMaFJ`, trace `tr-6c0ed0df894a92464674b8b69d97303d`).

— *Amber*

jsell-rh:
Labeling `amber/approved`. — *Amber*

jsell-rh:
Labeling `amber/approved`. — *Amber*

jeremyeder:
This uses an SA stored in vault and is never exposed to agents or users. OpenShell handles that.

bsquizz (approved):

PR #316, opened 2026-07-09
fix(tekton): build all Konflux components on every push to main

bsquizz:
## Summary
- Remove `pathChanged()` CEL filters from all 10 Konflux push-to-main pipelines so every merge to `main` builds all component images
- Fixes app-interface deployment failures where `saas.yaml` resolves `ref: main` to HEAD's commit SHA but not all component images exist for that SHA (because Konflux only built components whose source was touched)
- PR pipelines retain their path filters to avoid unnecessary CI on pull requests

## Test plan
- [ ] Verify all 10 push pipeline YAML files have `on-cel-expression: event == "push" && target_branch == "main"` with no `pathChanged()` clause
- [ ] Merge to main and confirm all 10 Konflux component builds trigger
- [ ] Verify all images appear in quay.io with the merged commit SHA as tag


bsquizz:
Only build config changes, can go in immediately

PR #317, opened 2026-07-09
fix(ci): align codeql-action SHAs and fix model-discovery test path

bsquizz:
## Summary
- Align all `github/codeql-action` sub-actions (`init`, `autobuild`) to the same v4.37.0 SHA that Dependabot set for `analyze`, fixing the version mismatch CI error
- Fix `test_model_discovery.py` import path — added missing `.parent` so path resolves to repo root `.github/scripts/` instead of `tests/.github/scripts/`

## Test plan
- [x] `python tests/unit/test_model_discovery.py -v` — all 24 tests pass locally
- [ ] CodeQL workflow runs without version mismatch error
- [ ] Model-discovery unit test job passes in CI


bsquizz:
Minor CI-only fixes, no code changed here. Eligible for fast merge

PR #299, opened 2026-07-08
fix(control-plane): rewrite _acp_internal sandbox policy hostnames to match deployment namespace

markturansky:
## Summary

- **Root cause**: The sandbox OPA policy's `_acp_internal` section hardcodes `ambient-code` in service hostnames (`ambient-control-plane.ambient-code.svc:8080`). When deployed to `ambient-dev` (or any non-`ambient-code` namespace), the supervisor proxy denies runner CONNECT requests to the CP `/token` endpoint because the hostname doesn't match — causing runner exit code 3 and no LLM responses.
- **Fix**: `ensureACPInternalPolicy()` rewrites `_acp_internal` endpoint hostnames using `CPRuntimeNamespace` on every `CreateSandboxRequest`, before sending to the gateway. Handles both existing policies (rewrite) and nil policies (inject).
- **Scope**: Gateway mode only (`OPENSHELL_USE_GATEWAY=true`). One line added to `provisionSessionSandbox()`, one new file with the rewrite logic, one test file.

## Files Changed

 < /dev/null |  File | Change |
|------|--------|
| `internal/reconciler/sandbox_policy.go` | New — `ensureACPInternalPolicy()`, `rewriteACPEndpointNamespace()`, `rewriteServiceHostNamespace()`, `acpInternalRule()` |
| `internal/reconciler/sandbox_policy_test.go` | New — 12 test cases (host rewriting, nil policy, existing policy rewrite, field preservation) |
| `internal/reconciler/kube_reconciler.go` | +1 line after `resolveAgentSandboxPolicy()`: `sandboxPolicy = ensureACPInternalPolicy(sandboxPolicy, r.cfg.CPRuntimeNamespace)` |

## Test plan

- [x] `go vet ./internal/reconciler/...` — 0 issues
- [x] `golangci-lint run` — 0 new issues
- [x] All 12 new unit tests pass (host rewriting, nil/empty/existing policy, field preservation)
- [x] `go build ./...` — clean
- [ ] Deploy to `ambient-dev` and verify runner successfully calls `/token` endpoint
- [ ] Verify sessions produce LLM responses end-to-end

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #318, opened 2026-07-09
feat: scope OpenShell sandbox policies to agent-derived sessions

squizzi:
## Summary
This is a follow-up to #280 which was closed in favor of #299 , I rebased on main and have re-worked the PR.

- Adds `kind: Policy` support to `acpctl apply` with create-or-update semantics, kustomize overlay merging, and `sandbox_policy` field on agents
- Injects `acp_internal` and `_mlflow_rh` network policies server-side in the control plane reconciler via `mergePlatformRules()` at `CreateSandbox` time, so runners always reach ACP services regardless of user-supplied policy, using the configurable `CPRuntimeNamespace` (not hardcoded)
- Ships `permissive` and `locked-down` example policies; wires all example agents to reference `permissive`
- Renames `filesystem_policy` to `filesystem` in runner policy.yaml, Rego evaluator, and UI template to align with proto field names
- E2e tests verify network enforcement via OpenShell proxy log inspection (not `kubectl exec curl`, which bypasses the eBPF process-tree filter)

## Deploy ordering

Migration `202607080100` renames the `filesystem_policy` JSONB key to `filesystem` in existing policy rows. The Rego evaluator (`policy.rego`) reads `data.filesystem`. **Deploy order: run the API server migration first, then roll out the runner image.** Between migration and runner deploy, the old Rego binary reads `data.filesystem_policy` (old key) from data that still has the old key — no breakage. If the runner image deploys before the migration, filesystem enforcement reads from the new key but the data still has the old key — enforcement silently goes to nil until the migration runs.

Migration `202607080100` is intentionally dated July 8, before #307's `202607090001`. gormigrate tracks applied migrations independently, so out-of-order application is handled correctly.

## Test plan

- [ ] `acpctl apply -f examples/base/policies/permissive.yaml` creates policy successfully
- [ ] `acpctl apply -k examples/overlays/tenant-a/` applies policies before agents
- [ ] Agent with `sandbox_policy: permissive` resolves the policy at sandbox creation
- [ ] Agent without `sandbox_policy` still gets `acp_internal` network rules injected
- [ ] Deploy to non-default namespace and verify service endpoints use configured namespace
- [ ] Rego evaluation enforces filesystem rules correctly with renamed `filesystem` key
- [ ] Locked-down policy sandbox shows DENIED entries in proxy log
- [ ] Permissive policy sandbox shows ALLOWED entries in proxy log


jsell-rh (changes requested):
🤖 **Amber — PR Review**

This is a draft with zero test items checked, so I'm flagging issues for the author to address before marking ready for review. The overall design is solid — policy scoping via `policyReplacementOperations`, `acp_internal`/`_mlflow_rh` always-on platform rules, and the `acpctl apply Policy` kind all follow established patterns. A few things need attention before this can land.

---

**Blocker — Rego key rename / deployment ordering risk**

`policy.rego` now reads `data.filesystem` instead of `data.filesystem_policy`. The DB migration `202607080100` renames existing JSONB data. Between migration execution and runner image deploy there is a window where old Rego binary is reading `data.filesystem` (new key) from data that still has `filesystem_policy` (old key). If the runner image is deployed before the migration runs, filesystem enforcement silently goes to nil. Please add a note to the PR description documenting the expected deploy order: migration first, then runner image rollout.

**Major — `bakedInRuleNames` list is control plane / runner coupling**

`bakedInRuleNames` hardcodes 11 network policy keys that are baked into the runner image. If the runner image gains or removes a baked-in rule without updating this list, `policyReplacementOperations` will either fail to remove a stale rule (leaving it active) or attempt to remove a non-existent rule (silently no-op). Please add a comment on the `bakedInRuleNames` slice explaining the coupling contract: "this list must be kept in sync with the policy keys baked into the runner image at build time." A follow-up issue to automate this check would be worth opening.

**Minor — Migration ID interleaving with PR #307**

This PR adds migration `202607080100`. PR #307 (currently `amber/change-requested`) adds `202607090001`. If #307 lands first, gormigrate will have `202607090001` applied and will later apply `202607080100` from this PR — which is fine since gormigrate tracks applied migrations independently. But reviewers merging #307 and then this PR may be surprised. Worth a comment in the PR description: "Migration `202607080100` is intentionally dated July 8 (pre-#307's `202607090001`); gormigrate handles out-of-order application correctly."

**Minor — `acpctl apply` kustomize overlay ordering**

The PR says "applies policies before agents" in the test plan. This relies on `kustomization.yaml` listing policies before agents (confirmed in `examples/base/kustomization.yaml`). The `run()` loop processes documents in file-stream order, so this is correct as written. No code change needed, but worth confirming this is tested with `acpctl apply -k` (test plan item 2).

---

**What looks good:**
- `applyPolicy` follows the same create-or-update pattern as `applyAgent` ✅
- `injectACPInternalPolicy` uses `CPRuntimeNamespace` (not hardcoded) ✅
- `policyReplacementOperations` correctly strips all baked-in rules before adding agent rules + platform rules ✅
- `platformMergeOperations` returns both `_acp_internal` and `_mlflow_rh` ✅
- Rollback in the JSONB migration is correct (reverses the rename) ✅
- `locked-down.yaml` and `permissive.yaml` example policies are well-formed ✅
- All 6 example agents wired to `sandbox_policy: permissive` ✅

Marking `amber/change-requested`. Happy to re-review when the test plan items are verified and the blocker is addressed.

— *Amber*

squizzi:
Thanks for the review. Addressing each point:

**Blocker — Rego key rename / deploy ordering:** Added a "Deploy ordering" section to the PR description documenting the expected order: migration first, then runner image rollout. ✅

**Major — bakedInRuleNames coupling:** This finding is **stale** — `bakedInRuleNames` and `policyReplacementOperations` were removed in subsequent commits. Platform rules are now injected via `mergePlatformRules()` directly into the `SandboxPolicy` protobuf at `CreateSandbox` time, eliminating the coupling between control plane and runner baked-in policy keys entirely. No post-hoc `UpdateConfig` replacement is needed. ✅

**Minor — Migration ID interleaving with #307:** Added note to PR description confirming gormigrate handles out-of-order application correctly. ✅

**Minor — acpctl apply kustomize overlay ordering:** Confirmed — `examples/base/kustomization.yaml` lists policies before agents, and `run()` processes documents in file-stream order. Test plan item 2 covers this. ✅

squizzi (commented):
🤖 **Amber Analysis**

Solid implementation closing P22/P23 gaps. The dual-injection pattern (platform rules at `CreateSandbox` time + post-ready `UpdateConfig`) is architecturally sound. All reconciler and SDK tests pass. No `panic()` calls, no `any` types introduced.

**Confidence:** [85%] High — patterns follow existing codebase conventions, tests verify key behaviors, but e2e network policy verification relies on proxy log inspection which is environment-dependent.

<details>

## ✅ What Works Well

1. **`mergePlatformRules()` pattern** (`sandbox_policy.go:74-81`) — Injecting `_acp_internal` and `_mlflow_rh` at `CreateSandbox` time ensures the gateway receives complete policy upfront. This eliminates the race condition where runners could start before platform rules were merged.

2. **SDK JSON serialization** (`policy_extensions.go`) — The `MarshalJSON`/`UnmarshalJSON` pair correctly handles the spec-as-object vs spec-as-string duality. The `PolicyPatchSpec` helper ensures PATCH payloads serialize correctly. Round-trip test confirms no double-encoding.

3. **Migration ordering** (`202607080100`) — Backdating the migration ID ensures it runs before #307's `202607090001`. Gormigrate tracks migrations independently, so out-of-order application is handled. The rollback path is correct.

4. **Kustomize ordering** (`examples/base/kustomization.yaml`) — Policies listed before agents ensures `sandbox_policy: permissive` references are resolvable when agents are applied.

5. **Test coverage** — New tests verify `mergePlatformRules` preserves existing rules, injects both platform rules, and uses namespace-scoped endpoints.

## 📋 Observations (Non-Blocking)

1. **Hardcoded MLflow endpoint** (`sandbox_policy.go:35`) — `mlflow.apps.int.spoke.prod.us-west-2.aws.paas.redhat.com` is production-specific. Consider making this configurable via `CPRuntimeNamespace` pattern or a dedicated config field for multi-cluster deployments.

2. **E2E test fragility** (`gateway-e2e-test.sh:573-600`) — Proxy log inspection (`grep DENIED/ALLOWED`) depends on the OpenShell proxy's log format. If the format changes, tests will fail silently. Consider adding a structured assertion or at minimum documenting the expected log format.

3. **Section numbering gap** — E2E test sections jump from 10 to 11 to 12 but the code comments say section 11. Minor inconsistency.

## Constitution Compliance

- **Principle III (Error Handling)**: ✅ No `panic()`, all errors returned with context
- **Principle IV (Testing)**: ✅ Unit tests for new reconciler and SDK code
- **Principle VI (Observability)**: ✅ Structured logging with namespace context
- **Principle X (Commit Discipline)**: ✅ Changes are atomic, conventional commits

</details>

**Rollback:**
```bash
git revert 64bfae3b..d4f8a7af && kubectl rollout restart deployment/ambient-api-server -n ambient-code
```

The migration rollback (`202607080100`) will restore `filesystem_policy` keys if needed.

jsell-rh (approved):
🤖 **Amber Re-Review — PR #318**

Checking my original blockers and majors against the new commits since `553b1834`.

---

**Blocker — Rego key rename / deploy ordering** ✅ Resolved
PR description now has a full "Deploy ordering" section documenting exactly when to run the migration vs roll out the runner image, and the silent nil enforcement risk. This is exactly what I asked for.

**Major — `bakedInRuleNames` coupling** ✅ Resolved
`bakedInRuleNames` and `policyReplacementOperations` were removed in `553b1834` ("Removes unused policyReplacementOperations/bakedInRuleNames"). The new `mergePlatformRules()` approach is cleaner — platform rules compose at `CreateSandbox` time rather than being reconstructed from a fragile hardcoded list.

**Minor — Migration ID interleaving with #307** ✅ Documented
PR description explains gormigrate tracks migrations independently. No action needed.

---

**New commits reviewed:**

- `64bfae3b` — Switches e2e sections 11c/11d from `kubectl exec curl` to proxy log inspection. This was my observation from the previous squizzi review thread. Correct fix — `kubectl exec` bypasses the eBPF process-tree filter entirely, so it was never testing the policy at all.

- `2a3e3cc6` — Narrows grep from `DENIED` → `DENIED.*api.anthropic.com`, adds a 30s poll loop (10×3s). Directly addresses the "too generic grep" observation from the squizzi review. The poll loop eliminates the race where Claude hasn't attempted the connection yet. ✅

- `1b0c4680` — `skip → fail` for unknown pod status in the repo payload section. Test hardening — correct call. Unknown pod status is a test setup failure, not a condition to silently skip. ✅

- `d66b1ef` — Spec documentation update. Accurate, matches the implemented dual-merge approach. ✅

---

**Remaining observation (non-blocking, same as before):**

The hardcoded MLflow endpoint `mlflow.apps.int.spoke.prod.us-west-2.aws.paas.redhat.com` in the platform rules is still production-specific. Not a blocker for this PR since it mirrors the existing `_acp_internal` pattern, but worth a follow-up to make it configurable for multi-cluster deployments.

---

**Still a draft with unchecked test plan items.** My approval covers the code quality and correctness. Mark ready-for-review and have someone verify the test plan items against a real hcmais-dev cluster before merge.

**Confidence:** [88%] High — all code-level blockers resolved, e2e test improvements are correct, design is sound. Gateway e2e coverage is environment-dependent as before.

**Rollback:** `git revert <sha-range>` + `kubectl rollout restart deployment/ambient-control-plane -n ambient-code`. Migration `202607080100` has a correct rollback (re-renames `filesystem` → `filesystem_policy`).

— *Amber*

bsquizz (approved):

jsell-rh:
🤖 **Amber — Re-review of 3 commits since last approval**

All clean. `amber/approved` stands.

| Commit | Finding |
|--------|---------|
| `d66b1ef` spec: document dual platform policy injection | Accurate — matches the `mergePlatformRules()` + post-ready `UpdateConfig` implementation. `_mlflow_rh` added to both specs correctly. ✅ |
| `2a3e3cc` fix: grep for `api.anthropic.com` specifically in e2e | Strictly better — removes false-positive risk from generic `DENIED`/`ALLOWED` matches. 30s polling window is appropriate for startup timing. ✅ |
| `1b0c4680` fix(ci): skip→fail for unknown pod status | Correct direction — surfaces real failures instead of silently passing. ✅ |

No new Go code, no new security surface, no regressions. The e2e improvements in particular make the test suite meaningfully more reliable.

— *Amber*

jsell-rh (commented):
🤖 **Amber — Follow-up on `0e933a7` + `b62724e`**

Both new commits are clean — my approval from `1b0c4680` stands.

- `0e933a7` (strip trailing whitespace in `auto-merge.yml`): Trivial one-line fix, no concerns.
- `b62724e` (dedicated permissive session for network policy e2e): This is the right call. Reusing the hello-world session from section 8 was fragile — a session that completed before section 12 ran would produce a vacuous pass or a confusing failure message. The new approach (start a fresh `network-test-permissive` session, wait for Running, poll proxy log, clean up) mirrors the locked-down test pattern exactly. The `PERM_SESSION_ID` cleanup addition is correct.

Still a draft with unchecked test plan items. Code-quality approval remains unchanged.

— *Amber*

jsell-rh:
🤖 **Amber — 2 new commits since last pass**

Both clean. `amber/approved` stands.

| Commit | Finding |
|--------|---------|
| `0e933a7` fix(ci): strip white space pre-hook | Trailing newline removed from `auto-merge.yml`. Mechanical, correct. ✅ |
| `b62724e` fix: use dedicated permissive session for network policy e2e | This is a meaningful improvement — the old approach reused the hello-world session from section 8, which had likely already exited by the time section 12 ran. The new fixture (`network-test-permissive` agent) and fresh session start with proper pod-ready + session-ready polling eliminates that race. Cleanup path also correctly handles the new `PERM_SESSION_ID`. ✅ |

No new concerns.

— *Amber*

jsell-rh (commented):
🤖 **Amber — Follow-up on `93056050` (prevent supervisor disconnect)**

My approval from `b62724ef` stands. Reviewing this commit on its own.

**What's solid:**
- `isUploadRetryable` correctly handles both the direct gRPC `Unavailable` path and the SSH-library-wrapped string path. The comment explaining *why* string matching is needed (`%v` wrapping, not `%w`) is exactly the right level of documentation.
- The retry loop is context-aware: `select` on `uploadCtx.Done()` + `timer.Stop()` on cancellation prevents goroutine leaks. Flow is correct — context cancellation sets `uploadErr`, `uploadCtx.Err() != nil` → break, then `uploadErr != nil` → fail session.
- Backoff sequence (1s, 2s, 4s, 8s) over 5 total attempts is reasonable for a transient supervisor relay disconnect.
- `TestIsUploadRetryable` covers the four meaningful cases cleanly.

**Minor notes (non-blocking):**

1. **String matching fragility** — acknowledged in the comment, but worth reiterating: if the SSH library changes its error wrapping format, `isUploadRetryable` will silently fall back to non-retryable for the wrapped case. A future improvement would be to check the error returned by `UploadPayloads` before it's wrapped, closer to the gRPC call site. Not blocking here.

2. **No retry-loop integration test** — `TestIsUploadRetryable` tests the helper correctly, but there's no test for the loop behavior itself: that it retries exactly N times, that it respects context cancellation mid-backoff, and that it breaks immediately on non-retryable errors. Consider a follow-up; not a blocker for this PR.

3. **Final failure log missing attempt count** — `"failed to upload payloads via SSH"` could include `Int("attempts", maxUploadRetries+1)` to make post-mortem log triage easier. One-liner improvement for a follow-up.

Still a draft with unchecked test plan items. Code-quality approval unchanged.

— *Amber*

jsell-rh (commented):
🤖 **Amber — Follow-up on `e04df7a` (SSH+curl e2e rewrite)**

My approval from `93056050` stands. Reviewing this commit on its own.

**What changed and why it's right:**

The log-scraping approach had a fundamental ordering problem: log entries from Claude's startup connections aren't guaranteed to appear within the polling window, and `kubectl exec` bypasses the eBPF process-tree filter entirely. Switching to SSH+curl is the correct fix — you're actively exercising the policy enforcement path rather than observing side effects.

**HTTP for locked-down / HTTPS for permissive** — the reasoning is sound. The proxy can intercept plain HTTP and return `policy_denied` JSON. For HTTPS, the proxy operates at the CONNECT tunnel level; if policy allows, curl gets the real response (non-empty → pass); if denied, connection fails (empty output → "no response" fail). The distinction is intentional and correct.

**`setup-gateway-cli.sh` idempotency** — PID file approach is cleaner than the old blocking `wait` + `trap cleanup EXIT`. Repeated invocations kill the previous port-forward before starting a new one, which is exactly what CI needs. The `kind-setup-openshell-cli-stop` target correctly reads `KIND_PF_DIR` to clean up.

**One note (non-blocking):**

`KIND_PF_DIR` is referenced in the Makefile targets but if undefined it evaluates to empty, which causes `PF_DIR=` in the script — the script then falls back to `/tmp/ambient-code` via `${PF_DIR:-/tmp/ambient-code}`. This is harmless but might cause confusion if someone sets a custom `KIND_PF_DIR` and then tries `make kind-setup-openshell-cli-stop` and finds no PID files there. Worth defining `KIND_PF_DIR ?= /tmp/ambient-code` in the Makefile to make the default explicit. Follow-up is fine.

CI step order looks correct: `pipx install openshell` → start port-forwards via `setup-gateway-cli.sh` → run e2e tests. The port-forward poll (30 attempts) in the script handles the startup delay.

Still a draft with unchecked test plan items. Approval stands.

— *Amber*

jsell-rh (approved):
🤖 **Amber — 3 new commits since last pass (17:36Z–20:56Z)**

All three clean. `amber/approved` stands.

| Commit | Summary | Finding |
|--------|---------|---------|
| `9305605` `fix(control-plane): prevent supervisor disconnect` | Exponential backoff retry (4×, 1→8s) for `UploadPayloads` on `Unavailable`; dual-path `isUploadRetryable` for direct gRPC + SSH-wrapped string matching; 5 unit tests | ✅ Retry logic is correct. Context cancellation handled properly (`timer.Stop()` + `uploadCtx.Err()` guard). `isUploadRetryable` dual path is necessary because the SSH library uses `%v` not `%w`, making `errors.Unwrap` useless. String matching correctly requires both `code = Unavailable` and `supervisor` — non-supervisor Unavailable still retries via the gRPC status path. |
| `7c789dde` `fix(e2e): rewrite network policy tests to SSH+curl` | Replaces log-scraping with active SSH+curl; `setup-gateway-cli.sh` made idempotent (PID files); new `kind-setup-openshell-cli-stop` target; openshell CLI installed in CI | ✅ Fundamentally more reliable — active verification removes the timing dependency that made log-scraping flaky. HTTP for locked-down (proxy can intercept and return `policy_denied`), HTTPS for permissive (vscode rule allows `/usr/bin/curl → :443`). The `|| true` + content-check pattern is correct. The FIXME for `openshell sandbox exec` is honest technical debt. |
| `fbea527e` `fix(ci): use larger runner for local dev e2e tests` | `ubuntu-latest` → `ubuntu-latest-16-cores` | ✅ Mechanical. Right call for a Kind cluster e2e that needs headroom. |

**⚠️ Non-blocking coordination note:**

Both this PR and PR #331 modify `setup-gateway-cli.sh`. `7c789dde` (this PR, 20:14Z) restructures the file with PID-based idempotency; #331 (20:47Z) reorders cert extraction before gateway registration. These touch different sections but whoever merges second will need a minor rebase. Both are drafts — worth coordinating landing order.

No new concerns.

— *Amber*

jsell-rh:
🤖 **Amber — Re-review of 3 commits since last pass (2026-07-10T16:32:47Z)**

All three clean. `amber/approved` stands.

| Commit | Finding |
|--------|---------|
| `93056050` fix(control-plane): prevent supervisor disconnect | Retry logic is correct. Exponential backoff (1s→2s→4s→8s, 4 retries) with context-cancellation via `select`. `isUploadRetryable` dual-path is well-reasoned — direct gRPC `Unavailable` via `status.FromError`, SSH-wrapped errors via string match requiring both `"code = Unavailable"` and `"supervisor"` to scope retries tightly. Timer is properly stopped on context cancellation. `for attempt := range maxUploadRetries + 1` (Go 1.22 range-over-int) iterates 5 times (0–4), final attempt correctly hits the `attempt == maxUploadRetries` break. Test table covers all four cases including the edge case "Unavailable without supervisor" (which retries via gRPC status path, not string match). ✅ |
| `7c789dde` fix(e2e): rewrite network policy tests to SSH+curl | This is a qualitatively better approach. Active SSH+curl into the sandbox directly exercises the proxy interception path rather than racing against log availability. HTTP for locked-down (proxy can return `policy_denied` JSON on GET), HTTPS for permissive (matches the vscode curl rule) — the asymmetry is documented and correct. `setup-gateway-cli.sh` idempotency improvement (PID files, kill-before-restart) is a nice side benefit. CI now installs openshell CLI before gateway tests and calls `setup-gateway-cli.sh` explicitly. The `|| true` on SSH commands is intentional — test failures surface via empty `CURL_OUTPUT` hitting the else/fail branch. ✅ |
| `fbea527e` fix(ci): use larger runner | Mechanical bump to `ubuntu-latest-16-cores`. Correct. ✅ |

No new concerns across any of the three.

— *Amber*

jsell-rh (commented):
🤖 **Amber — 2 commits since last pass (2026-07-11T05:34:39Z)**

Both clean. `amber/approved` stands.

| Commit | Finding |
|---|---|
| `a358e0f` Revert "fix(ci): use larger runner for local dev e2e tests" | Reverts `.github/workflows/test-local-dev.yml` `runs-on` from `ubuntu-latest-16-cores` back to `ubuntu-latest`. One-line, mechanical, correct. ✅ |
| `965afd39` Merge branch 'main' into squizzi/plateng-128/scope-openshell-policies | Merge-from-main. No conflicts with this branch's content. ✅ |

CI post-merge: Lint ✅, Unit Tests ✅, API Server Integration Tests ✅, CLI Build ✅, Ambient Runner Tests ✅, UI Unit Tests ✅. Konflux EC gates and arm64 builds still queued/running — expected post-merge latency, not failures.

No new concerns.

— *Amber*

jsell-rh (commented):
🤖 **Amber — 2 commits since last pass (`965afd39`, 2026-07-13T18:04Z)**

My approval from `965afd39` stands. Both new commits are clean.

| Commit | Summary | Finding |
|--------|---------|---------|
| `11b90af6` `fix(control-plane): prevent duplicate execAfterReady goroutines per session` | `sync.Mutex` + `map[string]struct{}` dedup guard; `tryClaimExec`/`releaseExec` helpers; `defer releaseExec` in `execAfterReady`; two unit tests | ✅ See analysis below |
| `a8a0437a` Merge branch `main` into `squizzi/plateng-128/scope-openshell-policies` | Conflict resolution in `skills/RECONCILE.md` (keeps P22/P23 done, adds P24–P27 gaps); spec updates from main | ✅ Mechanical merge. Conflict resolution is correct — P22/P23 are implemented on this branch. |

---

**`11b90af6` — Dedup Guard Analysis**

The race this fixes is real: a MODIFIED watch event can arrive before the `Pending→Creating` status write propagates through the watch stream, causing `provisionSessionSandbox` (and therefore `execAfterReady`) to run twice for the same session. The double-spawn led to double policy injection, double entrypoint execution, and double payload uploads.

**Implementation correctness:**
- `tryClaimExec` / `releaseExec` are textbook mutex + map. Lock semantics are correct — `Lock` before read+write in `tryClaimExec`, `Lock` before delete in `releaseExec`. ✅
- `defer r.releaseExec(sessionID)` as the first statement in `execAfterReady` ensures the slot is freed on every exit path (timeout, error, success). ✅
- The guard is checked in **both** `provisionSessionSandbox` call sites (the gateway path at line ~397 and the non-gateway path at line ~447). Checking only one would have left the other unprotected. ✅
- Log message on skip (`"execAfterReady already running for session; skipping duplicate"`) is structured with `session_id` and `sandbox` — observable and debuggable. ✅

**Test coverage:**
- `TestTryClaimExec_PreventsDoubleSpawn` — happy path: first claim succeeds, second rejects, different session succeeds, post-release claim succeeds. All four cases. ✅
- `TestTryClaimExec_ConcurrentSafety` — 50 goroutines race on one session ID, exactly 1 must claim. Uses `atomic.Int32` to avoid data races in the counter. This is the correct test for a mutex — races would show up under `go test -race`. ✅

**One minor observation (non-blocking):**

`activeExecs` is never cleaned up for sessions that were skipped by `tryClaimExec` returning false — but that's correct, because the running goroutine holds the slot and will `defer releaseExec`. The only concern would be if `provisionSessionSandbox` returns `nil` after a skip but the goroutine for that session panics before `defer releaseExec` fires — but panics in goroutines crash the process, so there's no stuck-slot scenario. The logic is sound.

**Constitution compliance:**
- No `panic()` introduced ✅
- Structured logging with context fields ✅
- No tokens/secrets ✅
- Concurrent safety verified by test ✅

---

No new concerns. Still a draft with unchecked test plan items — code-quality approval unchanged.

— *Amber*

jsell-rh (commented):
## 🤖 Amber — 4 commits since last pass (`a8a0437`, 2026-07-13T20:32Z)

My approval from `a8a0437` stands. Reviewing the three substantive new commits.

---

**`2413a8d` — fix(control-plane,runner): pod termination races + slow shutdown**

- `podDeletedAt` timestamp with 10s fast-recreation window: the state machine is correct. Tracking deletion time and accepting READY-after-10s avoids the 600s hang when the sandbox controller reconciles faster than the 2s poll. The zero-value reset (`podDeletedAt = time.Time{}`) on the success path is clean. ✅
- `terminationGracePeriodSeconds` 60s → 15s: Appropriate for a `restartPolicy: Never` session pod being torn down. The runner gets SIGTERM, the 15s window is enough for graceful shutdown. ✅
- `SessionManager.shutdown()` — `asyncio.gather(return_exceptions=True)` over sequential `for` loop: Correct. `return_exceptions=True` means a failing `destroy()` on one worker doesn't abort the rest. The previous sequential approach meant a stuck worker would block all subsequent shutdowns for up to 15s each. ✅

**`557c1b6` — fix(e2e): apply sandbox policies before sessions start**

Correct fix. Policy objects must exist before any session that references them starts — the control plane fails sessions when `sandbox_policy` by name is unresolvable. Section renumbering (6→7→...→13) is clean and consistent throughout the script. Comment explaining the move is present. ✅

**`aeb1491` — fix(control-plane): non-blocking reconciler dispatch**

This is the right architectural fix — blocking `provisionSessionSandbox` inside the informer dispatch loop was serializing all session events behind a single gateway call.

Implementation is solid and consistent with the `tryClaimExec` pattern:
- `provisionAsync`: `tryClaimProvision` checked in `Reconcile` **before** spawning goroutine — no goroutine created if already claimed. ✅
- Detached contexts with sensible timeouts (5min provision, 2min deprovision). ✅
- `defer releaseProvision`/`releaseDeprovision` ensures slots are freed on all exit paths. ✅
- Structured error logging with `session_id` field on failure. ✅

**Two minor observations (non-blocking):**

1. **Deprovision path spawns goroutine before claiming dedup slot** — `deprovisionAsync` is always goroutine-spawned without a prior `tryClaimDeprovision` call in `Reconcile` (unlike the provision path). Multiple MODIFIED events in `PhaseStopping` each spawn a goroutine; the second goroutine exits immediately inside `deprovisionAsync`. This is still correct (double-deprovisioning is prevented), but it creates brief goroutine overhead. Worth aligning the two paths in a follow-up.

2. **Provisioning errors are no longer propagated to the informer** — `provisionAsync` logs errors but doesn't retry. Previously, if `provisionSession` returned an error, the caller (and informer watch loop) would observe it. Now errors are silently absorbed after logging. The compensating mechanism is that `provisionSession` marks the session `PhaseFailed` on internal errors, so the watch loop will eventually get a MODIFIED event with `PhaseFailed` and deprovisioning will follow. Worth documenting this invariant in a comment on `provisionAsync`.

Neither warrants blocking merge.

**Constitution compliance on new code:**
- No `panic()` ✅
- Structured logging with context fields ✅
- `sync.Mutex` concurrency is correct ✅
- No tokens/secrets ✅

Still a draft with unchecked test plan items. Code-quality approval unchanged.

*— Amber*

jsell-rh (commented):
## 🤖 Amber — Follow-up on `2056ff24` (merge from main, 2026-07-14T03:10Z)

My approval from `a8a0437` stands. This is a merge-from-main bringing in 5 new commits from `main`. Reviewing the substantive changes below.

---

### `kube_reconciler.go` — ExecSandbox retry expansion

`maxExecRetries` bumped from 3 → 29, `execRetryDelay` 3s → 2s (total window: ~58s vs 9s). The retry logic now retries on **all** errors (not just gRPC `Unavailable`), which matches the updated spec requirement (`ExecSandboxStreaming SHALL retry on all errors`). The `failSession` call is now guarded by `attempt == maxExecRetries` rather than error-type filtering — this is cleaner and avoids the previous silent-drop risk for non-Unavailable errors. ✅

One observation (non-blocking): the `time.Sleep(execRetryDelay)` in the retry loop is not context-aware — if the session context is cancelled mid-backoff, the goroutine won't notice until the sleep expires. For a 2s delay this is acceptable, but worth noting for consistency with the `patchSandboxDNSConfig` retry which has the same characteristic.

---

### `app.py` — Rotating file log handler

Graceful-degradation pattern (`try/except OSError`) is correct — the runner silently falls back to stdout-only if `/sandbox/.runner/logs` isn't writable. 50 MB max with 3 backups is reasonable for a session container. ✅

`policy.yaml` correctly adds `/sandbox/.runner/logs` to the filesystem allowlist so the file handler can actually write there when the sandbox policy is applied. ✅

---

### `session.py` — Connect retry with `asyncio.Event` readiness gate

`_connect_ready` + `_connect_error` + `wait_for_connect()` is the right pattern for propagating async connection results to callers. `asyncio.wait_for` + the error re-raise in `wait_for_connect` is clean. Env-var overrides (`CLAUDE_CONNECT_MAX_RETRIES`, `CLAUDE_CONNECT_RETRY_DELAY`) are good for testing. ✅

---

### `_grpc_client.py` — CP token fetch retries expanded

`_CP_TOKEN_FETCH_ATTEMPTS` 3 → 30 with flat 2s delay (60s total window). Removes the exponential backoff in favor of flat retry — acceptable for a startup path where the control plane may be slow to become ready. ✅

---

### `setup-gateway-cli.sh` — cert extraction order fix

Swapped: `gateway add` now runs **before** cert extraction. The comment explains exactly why (`gateway add --local` generates self-signed certs that don't match the cluster PKI; overwriting afterwards with the real cluster certs fixes `BadSignature / DecryptError`). This is a real TLS ordering bug fix — correct call. ✅

---

### `test-agent-no-providers.yaml` + `kustomization.yaml`

New test agent without credential providers added to `examples/base/`. Simple and well-formed — `prompt`, `payloads`, `environment`, `labels`. No `sandbox_policy` reference (valid — platform rules still inject). ✅

---

**Constitution compliance on incoming changes:**
- No `panic()` introduced ✅
- No tokens/secrets in logs (file handler logs to disk, format is `%(levelname)s:%(name)s:%(message)s` — no sensitive values) ✅
- No `any` types in TS changes (package-lock.json only, no type changes) ✅
- `policy.yaml` filesystem allowlist correctly scoped ✅

Still a draft with unchecked test plan items. Code-quality approval unchanged.

*— Amber*

jsell-rh (commented):
## 🤖 Amber — 2 commits since last pass (`2056ff24`, 2026-07-14T03:36Z)

My approval from `2056ff24` stands. Both new commits are clean.

| Commit | Summary | Finding |
|--------|---------|---------|
| `86716034` `fix(ci): extract mTLS certs before openshell gateway registration` | Cert extraction now runs **before** `gateway add --local` to satisfy openshell v0.0.82's requirement for client TLS material at registration time; re-extraction after registration overwrites any self-signed certs the CLI generates | ✅ Correct. The double-extract pattern (pre-register + post-register overwrite) handles both the "certs must exist" requirement and the "CLI may generate mismatched self-signed certs" problem. The comment explains the `why` clearly. This reverses the commit from `2056ff24` that moved extraction to after registration — confirmed this is the right ordering fix. |
| `c48f1fc7` `fix(e2e): skip repo payload test that requires vertex provider` | Removes ~110 lines of section 12 repo payload test body; replaces with `skip "Repo payload verification" "vertex provider not available in CI"` | ✅ Correct. The `repo-clone-workspace` agent requires `providers: [vertex]` and CI doesn't have Vertex credentials — the sandbox pod would never reach Running, causing a hard timeout. Skipping cleanly is the right call vs. letting it timeout and fail noisily. The comment in the section header and the skip message are honest about the gap. |

**One observation (non-blocking):**

Section 12 now has a `section "12. Repository payload verification"` header followed by `REPO_SESSION_ID=""` and then immediately `if [ -n "$REPO_AGENT_ID" ]; then` — but `REPO_AGENT_ID` was removed from section 5 in a prior commit. This means the `if` condition is always false and the section effectively becomes dead code beneath the `skip`. The `$REPO_SESSION_ID` variable is also set but unused at cleanup. Consider cleaning up the remnant conditionals when section 12 is eventually re-enabled. Not blocking.

No new concerns.

*— Amber*

PR #319, opened 2026-07-09
fix(ui): use server-side session phase counts and filtering

squizzi:
## Summary

The FleetSummary status bar at the top of the sessions page was computing per-phase counts (e.g. "3 Running, 2 Failed") from only the current page's sessions (max 20), giving inaccurate totals when sessions spanned multiple pages. Phase filtering was also client-side only, meaning clicking a phase chip like "Pending" would filter within the current page rather than querying the server — so phases with no sessions on the visible page appeared unclickable.

This PR adds a dedicated `useSessionPhaseCounts` query hook that fetches accurate per-phase totals from the API via parallel lightweight calls (`size: 1`, reading only the `total` field). It also moves phase filtering to the server by adding a `phase` parameter to `ListParams` and including it in the API search query, so pagination works correctly when a phase filter is active.

## Changes

- **`ports/sessions.ts`** — Added `SessionPhaseCounts` type and `phaseCounts` method to the port interface
- **`adapters/sdk-sessions.ts`** — Implemented `phaseCounts()` with 7 parallel API calls; extended `buildSdkListOptions` to support `phase` filter in the search query
- **`queries/query-keys.ts`** / **`use-sessions.ts`** — Added `phaseCounts` query key and `useSessionPhaseCounts` hook (5s polling)
- **`fleet-summary.tsx`** — Accepts `phaseCounts` and `serverTotal` props instead of computing counts from `sessions[]`
- **`fleet-table.tsx`** — Removed client-side phase filter syncing (server handles it now)
- **`page.tsx`** — Wires up `useSessionPhaseCounts`, passes `phase` to `useSessions`, resets to page 1 on filter change
- **Tests** — Updated FleetSummary and adapter tests to match new interfaces

## Test plan

- [ ] Verify phase counts in the summary bar reflect totals across all pages, not just the current page
- [ ] Click each phase chip (Running, Pending, Creating, etc.) and confirm the table shows only sessions of that phase with correct pagination
- [ ] Click an active phase chip again to clear the filter and confirm all sessions return
- [ ] Verify "0 sessions" empty state only shows when there are truly no sessions (not when a phase filter returns empty)
- [ ] Run `vitest run` — all FleetSummary and adapter tests pass

### DIFF
--- a/components/ambient-ui/src/adapters/__tests__/sdk-sessions.test.ts
+++ b/components/ambient-ui/src/adapters/__tests__/sdk-sessions.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest'
+import { describe, it, expect, vi } from 'vitest'
 import type { Session, SessionList, ListOptions } from 'ambient-sdk'
 import type { SessionsPort } from '@/ports/sessions'
 import { createSessionsAdapter } from '../sdk-sessions'
@@ -192,27 +192,35 @@ describe('sdk-sessions adapter', () => {
     expect(capturedOpts?.size).toBe(10)
   })
 
-  it('phaseCounts() returns counts per phase from the server', async () => {
-    const capturedSearches: string[] = []
-    const fakeAPI = {
-      ...createFakeSessionAPI({}),
-      list: async (listOpts?: ListOptions): Promise<SessionList> => {
-        const search = listOpts?.search ?? ''
-        capturedSearches.push(search)
-        let total = 0
-        if (search.includes("phase = 'Running'")) total = 5
-        else if (search.includes("phase = 'Failed'")) total = 2
-        else if (search.includes("phase = 'Completed'")) total = 10
-        return { kind: 'SessionList', page: 1, size: 1, total, items: [] }
-      },
-    }
+  it('phaseCounts() fetches counts from server-side endpoint', async () => {
+    const serverCounts = { Running: 5, Failed: 2, Completed: 10 }
+    const fetchSpy = vi.spyOn(globalThis, 'fetch').mockResolvedValue(
+      new Response(JSON.stringify(serverCounts), { status: 200 })
+    )
+    const fakeAPI = createFakeSessionAPI({})
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const adapter: SessionsPort = createSessionsAdapter(fakeAPI as any)
 
     const counts = await adapter.phaseCounts('proj-123')
 
     expect(counts).toEqual({ Running: 5, Failed: 2, Completed: 10 })
-    expect(capturedSearches).toHaveLength(7)
-    expect(capturedSearches.every(s => s.includes("project_id = 'proj-123'"))).toBe(true)
+    expect(fetchSpy).toHaveBeenCalledOnce()
+    expect(fetchSpy.mock.calls[0][0]).toContain('/api/ambient/v1/sessions/phase_counts')
+    expect(fetchSpy.mock.calls[0][0]).toContain('project_id=proj-123')
+    fetchSpy.mockRestore()
+  })
+
+  it('phaseCounts() returns empty object on server error', async () => {
+    const fetchSpy = vi.spyOn(globalThis, 'fetch').mockResolvedValue(
+      new Response('error', { status: 500 })
+    )
+    const fakeAPI = createFakeSessionAPI({})
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const adapter: SessionsPort = createSessionsAdapter(fakeAPI as any)
+
+    const counts = await adapter.phaseCounts('proj-123')
+
+    expect(counts).toEqual({})
+    fetchSpy.mockRestore()
   })
 })
--- a/components/ambient-ui/src/adapters/__tests__/sdk-sessions.test.ts
+++ b/components/ambient-ui/src/adapters/__tests__/sdk-sessions.test.ts
@@ -191,4 +191,28 @@ describe('sdk-sessions adapter', () => {
     expect(capturedOpts?.page).toBe(2)
     expect(capturedOpts?.size).toBe(10)
   })
+
+  it('phaseCounts() returns counts per phase from the server', async () => {
+    const capturedSearches: string[] = []
+    const fakeAPI = {
+      ...createFakeSessionAPI({}),
+      list: async (listOpts?: ListOptions): Promise<SessionList> => {
+        const search = listOpts?.search ?? ''
+        capturedSearches.push(search)
+        let total = 0
+        if (search.includes("phase = 'Running'")) total = 5
+        else if (search.includes("phase = 'Failed'")) total = 2
+        else if (search.includes("phase = 'Completed'")) total = 10
+        return { kind: 'SessionList', page: 1, size: 1, total, items: [] }
+      },
+    }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const adapter: SessionsPort = createSessionsAdapter(fakeAPI as any)
+
+    const counts = await adapter.phaseCounts('proj-123')
+
+    expect(counts).toEqual({ Running: 5, Failed: 2, Completed: 10 })
+    expect(capturedSearches).toHaveLength(7)
+    expect(capturedSearches.every(s => s.includes("project_id = 'proj-123'"))).toBe(true)
+  })
 })
--- a/components/ambient-ui/src/adapters/sdk-sessions.ts
+++ b/components/ambient-ui/src/adapters/sdk-sessions.ts
@@ -1,17 +1,23 @@
 import type { SessionAPI, SessionCreateRequest } from 'ambient-sdk'
-import type { SessionsPort } from '@/ports/sessions'
-import type { DomainSession, DomainSessionCreateRequest, ListParams, PaginatedResult } from '@/domain/types'
+import type { SessionsPort, SessionPhaseCounts } from '@/ports/sessions'
+import type { DomainSession, DomainSessionCreateRequest, ListParams, PaginatedResult, SessionPhase } from '@/domain/types'
 import { mapSdkSessionToDomain } from './mappers'
 import { getSessionAPI } from './sdk-client'
+
+const ALL_PHASES: SessionPhase[] = ['Running', 'Pending', 'Creating', 'Stopping', 'Failed', 'Completed', 'Stopped']
 
 function sanitizeSearch(value: string): string {
   return value.replace(/['"%;\\]/g, '')
 }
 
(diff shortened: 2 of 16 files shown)


components/ambient-ui/src/adapters/sdk-sessions.ts
```diff
@@ -93,6 +99,24 @@ function createSdkSessionsAdapter(api: SessionAPI): SessionsPort {
...
     },
+
+    async phaseCounts(projectId: string): Promise<SessionPhaseCounts> {
+      const results = await Promise.all(
+        ALL_PHASES.map(async (phase) => {
+          const result = await api.list({
+            page: 1,
+            size: 1,
+            search: `project_id = '${sanitizeSearch(projectId)}' and phase = '${phase}'`,
+          })
```
jsell-rh: **Major — 7 parallel API calls every 5 seconds**

`phaseCounts()` fires 7 simultaneous `api.list()` requests every 5 seconds for every user with the sessions page open. That's 84 API calls/minute per browser tab, just for the summary bar. At even modest user counts this becomes noisy backend load.

Two approaches to consider before going ready-for-review:

**Option A — server-side aggregation (preferred):** Add a `GET /api/sessions/counts?project_id=` endpoint that returns all phase totals in one call. The API server can do a single `GROUP BY phase COUNT(*)` query. One request per poll interval instead of 7.

**Option B — single filtered pass per poll (simpler, no backend change):** Do one `list(size: 1)` for the current phase filter only, and derive a "filtered total" from that. For the full breakdown, accept approximate counts derived from the `useSessions` data already in cache. Less accurate but 7× cheaper.

The current approach is correct and will work — I'm flagging this before it ships to production, not blocking a kind/dev run. But this should be resolved before the draft is marked ready.

— *Amber*

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/_components/fleet-summary.tsx
```diff
@@ -12,39 +13,34 @@ const VARIANT_RING_CLASS: Record<string, string> = {
...
 
   const phases: SessionPhase[] = ['Running', 'Pending', 'Creating', 'Stopping', 'Failed', 'Completed', 'Stopped']
 
   return (
     <div className="flex items-center gap-4 text-sm rounded-lg border bg-muted/30 px-4 py-2.5">
       <span className="font-medium">
         {showFiltered
-          ? `Showing ${filteredCount} of ${displayTotal} sessions`
-          : `${displayTotal} sessions`}
+          ? `Showing ${filteredCount} of ${serverTotal} sessions`
```
Comment (jsell-rh) [inline components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/_components/fleet-summary.tsx:38]: **Minor — `showFiltered` semantic shift**

Old: `filteredCount !== sessions.length` (comparing against items on current page)  
New: `filteredCount !== pageItemCount` (same comparison, different variable name)

This preserves the same logic, but the condition now means: "show 'Showing N of M' when the search filter trims the visible list." One edge case: when a search filter is active but matches all 20 items on the page, `filteredCount === pageItemCount`, so the banner shows "100 sessions" — even though a search filter is active. The old code had the same behaviour, so this isn't a regression, just worth documenting.

If you want the banner to always show "Showing N of M" when a filter is active (regardless of match count), you could drive it off `!!searchFilter || !!activePhase` instead. Up to you — not blocking.

— *Amber*

components/ambient-ui/src/queries/use-sessions.ts
```diff
@@ -82,6 +82,19 @@ export function useAllSessions(
...
 }
 
+export function useSessionPhaseCounts(
+  projectId: string,
+  port?: SessionsPort,
+) {
+  const adapter = port ?? getDefaultPort()
+  return useQuery({
+    queryKey: queryKeys.sessions.phaseCounts(projectId),
+    queryFn: () => adapter.phaseCounts(projectId),
```
jsell-rh, line 92: **Minor — no staleTime set**

`useSessionPhaseCounts` uses `refetchInterval: 5000` (fine), but without `staleTime`, React Query marks the data stale immediately after the first fetch, which can trigger extra background refetches on window focus / remount on top of the 5s interval. The existing `useSessions` hooks in this file likely have the same pattern, so this is consistent — but worth setting `staleTime: 4000` (just under the refetch interval) to suppress the on-focus double-fetch.

— *Amber*

components/ambient-ui/src/adapters/__tests__/sdk-sessions.test.ts
```diff
@@ -191,4 +191,28 @@ describe('sdk-sessions adapter', () => {
...
+        capturedSearches.push(search)
+        let total = 0
+        if (search.includes("phase = 'Running'")) total = 5
+        else if (search.includes("phase = 'Failed'")) total = 2
+        else if (search.includes("phase = 'Completed'")) total = 10
+        return { kind: 'SessionList', page: 1, size: 1, total, items: [] }
+      },
+    }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const adapter: SessionsPort = createSessionsAdapter(fakeAPI as any)
```
jsell-rh, line 202: **Nit — `eslint-disable` with good justification, but consider typing the fake**

The `as any` cast on `fakeAPI` is the standard escape for adapter tests — the eslint-disable is appropriate here. If you wanted to avoid it, `createFakeSessionAPI({})` could be extended to override just `list`, avoiding the cast entirely. Not blocking; just noting there's a cleaner path if this pattern grows.

— *Amber*

jsell-rh (changes requested):
🤖 **Amber Review — PR #319: server-side session phase counts and filtering**

The architecture is clean and the direction is correct — moving phase counts server-side and wiring phase filtering through the API is the right call. The component interface cleanup (`serverTotal` as required prop, removing client-side `counts` computation) is well-executed. Tests are updated and the adapter test validates the parallel-call pattern correctly.

**One issue to resolve before marking ready:**

The main concern is the **N=7 parallel API calls every 5 seconds** in `phaseCounts()`. This is correct and will work, but 84 HTTP requests/minute/tab is something we'll want to address before it reaches production. See the inline comment on `sdk-sessions.ts:110` for two concrete options.

Everything else is minor or nit level — non-blocking for the draft phase.

**Checklist against ACP standards:**

- ✅ Zero `any` types (one justified `eslint-disable` in test, appropriate)
- ✅ React Query for all data ops (`useSessionPhaseCounts`, `useSessions`)
- ✅ `type` over `interface` throughout
- ✅ Loading and error states already handled in parent (`page.tsx`)
- ✅ Query keys include all relevant parameters (`phaseCounts(projectId)`)
- ✅ Server-total–driven pagination now works correctly across phase filters
- ✅ Page reset on phase filter change (`setCurrentPage(1)` in `handlePhaseFilter`)
- ✅ Empty state guard correctly skips when a phase filter is active
- ✅ `phaseEquals` client-side filter function removed (dead code cleanup)

Marking `amber/change-requested` for the API call volume issue. Happy to re-review once that's addressed or an alternative approach is chosen.

— *Amber*

components/ambient-ui/src/adapters/sdk-sessions.ts
squizzi: Addressed with Option A — added `GET /sessions/phase_counts` endpoint that runs a single `SELECT phase, COUNT(*) GROUP BY phase` query. The UI adapter now makes 1 request per poll cycle instead of 7 (12 req/min vs 84 req/min per viewer).

Changes: `dao.go` (GROUP BY query), `service.go` + `handler.go` (RBAC via ApplyProjectScope/ApplyListFilter), `plugin.go` (route), and UI adapter switched from 7 `api.list()` calls to a single `fetch()` against the new endpoint.

components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/_components/fleet-summary.tsx
Comment (squizzi) [reply components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/_components/fleet-summary.tsx:38]: Good catch on the edge case. Leaving the current behavior as-is since it matches the old logic and the banner semantics are correct for the common case. Driving it off filter state presence would require adding a `searchFilter` prop to `FleetSummary` — a reasonable follow-up but not worth the prop surface change for this PR.

components/ambient-ui/src/queries/use-sessions.ts
squizzi: Added `staleTime: 4000` to `useSessionPhaseCounts` — just under the 5s refetch interval to suppress window-focus double-fetches.

components/ambient-ui/src/adapters/__tests__/sdk-sessions.test.ts
squizzi: Acknowledged — the `as any` cast is the established pattern for adapter tests in this codebase. Worth cleaning up if the fake pattern grows, but not blocking for this PR.

jsell-rh (approved):
🤖 **Amber — Approved** (re-review after latest push)

All three concerns from my initial review are addressed. Lifting `amber/change-requested`.

| # | Finding | Status |
|---|---------|--------|
| 1 | 84 req/min (7 parallel calls × 5s interval) | ✅ Fixed — `GET /sessions/phase_counts` runs a single `GROUP BY` query. 12 req/min per viewer. |
| 2 | `staleTime` missing on `useSessionPhaseCounts` | ✅ Fixed — `staleTime: 4000` added (just under the 5s interval). |
| 3 | `showFiltered` edge case (acknowledged) | ✅ Noted and intentionally left as-is — matches old behaviour, reasonable trade-off. |
| 4 | `eslint-disable` in test | ✅ Acknowledged — established pattern, acceptable. |

**New code review:**

- `dao.go` — `GROUP BY phase` query is correct, nil-phase rows are filtered, result map is pre-sized. ✅
- `handler.go` — `ApplyProjectScope` + `ApplyListFilter` for RBAC. Route registered before `/{id}` so gorilla/mux matches it first. ✅
- `service.go` — Clean delegation, errors wrapped with `errors.GeneralError`. ✅
- `use-sessions.ts` — `enabled: !!projectId` guard, `staleTime: 4000`, `refetchInterval: 5000`. ✅
- `fleet-summary.tsx` — Props are now correct (server-driven, no client-side `reduce`). ✅
- `page.tsx` — `handlePhaseFilter` resets `currentPage` to 1 on filter change — correct. Empty-state guard correctly excludes active phase filters. ✅
- Tests — Updated to match new prop interface, good error-path coverage on `phaseCounts()`. ✅

**One minor observation (not blocking):** `handler.go` extracts `projectId` from raw query params and header *after* `ApplyProjectScope` has already resolved it into `listArgs`. The RBAC filter (`ApplyListFilter`) uses `listArgs` so security is fine; the SQL WHERE clause just gets a redundant re-extraction. Not worth a change for this PR — just flag it if you standardize handler patterns later.

Solid implementation. Shipping.

— *Amber*

PR #320, opened 2026-07-09
fix(ui): disable sidebar nav when no project selected and remove content max-width

squizzi:
## Summary

- Sidebar nav items now render as disabled buttons (grayed out, non-clickable) when no project is selected, instead of silently redirecting every click to the project picker root page
- Removed the `max-w-7xl` constraint from the dashboard content wrapper so page content expands to fill available width when the sidebar is collapsed

## Test plan

- [ ] With no project selected, verify sidebar items appear visually disabled and are not clickable
- [ ] Select a project and verify sidebar items become active links again
- [ ] Toggle the sidebar closed and verify content expands to fill the freed space
- [ ] Check Settings page still constrains itself with its own `max-w-md`

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/components/ambient-ui/src/app/(dashboard)/layout.tsx
+++ b/components/ambient-ui/src/app/(dashboard)/layout.tsx
@@ -104,7 +104,7 @@ export default function DashboardLayout({
             detailName={agentId ? (agent?.displayName ?? agent?.name ?? agentId) : null}
           />
           <PlatformHealthBanner />
-          <div className="flex-1 p-6 pb-10 max-w-7xl">{children}</div>
+          <div className="flex-1 p-6 pb-10">{children}</div>
           <StatusBar />
           <CommandPalette />
         </SidebarInset>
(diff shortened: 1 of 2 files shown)


jsell-rh (approved):
🤖 **Amber — Approved**

Clean, focused fix. Both changes are correct and well-scoped.

**`app-sidebar.tsx`** — The `needsProject` guard is the right call. Using `disabled` on `SidebarMenuButton` (Shadcn component, correct) instead of silently hijacking clicks is strictly better UX. No `any` types, no custom components, logic is clear.

**`layout.tsx`** — Removing `max-w-7xl` from the shared wrapper is the right level to make this change (page-specific layouts like Settings can still constrain themselves, which the PR description notes they do). No functional side effects.

Checklist clear: zero `any`, Shadcn-only, no API surface changes, no K8s child resources, no security surface touched.

One optional follow-up (not blocking): the disabled sidebar items have no tooltip or visual indicator explaining *why* they're disabled (no project selected). A `Tooltip` wrapping the disabled button with "Select a project first" would improve discoverability — but this is a polish item, not a requirement.

— *Amber*

jsell-rh:
🤖 **Amber** — adding `amber/approved` label.

— *Amber*

PR #324, opened 2026-07-10
fix(control-plane): improve sandbox state sync detection

JuanmaBM:
## Summary

Improves sandbox state synchronization to detect and handle externally deleted sandboxes.

## Changes

- **Reduced poll interval**: 15s → 5s for faster detection of state changes
- **Structured condition on external deletion**: When `GetSandbox()` returns `NotFound` for a `Running` or `Creating` session, writes a `SandboxDeleted/DeletedExternally` condition to DB
- **Refactored phase update logic**: Extracted `updateSessionPhaseWithConditions()` to support conditions without requiring a pod object

## Problem Solved

When a sandbox was deleted externally (e.g., `kubectl delete`), the session record in the DB was never updated and would remain in `Running` state indefinitely. This fix ensures:

1. Faster detection (5s vs 15s polling)
2. Clear failure reason visible in UI via structured condition
3. Session properly marked as `Failed` with explanation

## Testing

- ✅ Code compiles cleanly
- ⚠️ End-to-end testing requires investigation of watch mechanism issues (known issue in main - watches from K8s → API server may not be properly configured)

## Related

Part of #223 - Session lifecycle state sync improvements

## Note

This PR addresses **only** the polling-based state sync improvement. A separate investigation is needed for the K8s watch mechanism, which appears to have missing watches that prevent reactive state updates (e.g., sessions stuck in "Stopping" state).

🤖 Generated with Claude Code


jsell-rh (changes requested):
🤖 **Amber Analysis — PR #324**

The refactoring direction is correct and the CI gates are green. However there are two issues blocking approval.

**Blocking — Semantic bug in condition status**
`"status": "False"` on the `SandboxDeleted` condition is backwards. Kubernetes condition convention: `True` = the condition is active/present. `SandboxDeleted=False` reads "the sandbox was *not* deleted", but this code only runs when it *was*. See inline comment.

**Blocking — Coordination with PR #307**
PR #307 (currently `amber/change-requested`) contains the identical `pod_sync.go` changes in one of its commits (`e61dce1` — `fix(control-plane): fix sandbox lifecycle bugs (#223)`). Both PRs are open against `main`, referencing `#223`. Whoever merges second will have a conflict. Please clarify in the PR body:
- Is #324 intended to land first so #307 can be rebased on top without the pod_sync.go slice?
- Or is #307 being superseded by splitting into smaller PRs?

Either is fine — just needs to be explicit so reviewers can coordinate.

**Major — No tests for new code path**
`updateSessionPhaseWithConditions` is a new exported-equivalent function with a non-trivial `conditionsJSON` path. The `NotFound` → write `SandboxDeleted` condition flow is the core feature of this PR and has no unit test. The existing `pod_sync_test.go` should cover: (a) `NotFound` on a `Running` session writes `SandboxDeleted/True/DeletedExternally`, and (b) `updateSessionPhase` with `pod=nil` still produces no `conditions` patch key.

**Minor — `json.Marshal` error silently discarded**
See inline comment.

---

Confidence: **High** on the semantic bug, **High** on the #307 overlap, **Medium** on the test gap (may be tested elsewhere — I don't have the full test file in context).

Will re-review on next push.

— *Amber*

jsell-rh (commented):
**Blocker — condition `status` is inverted**

Kubernetes condition convention: `status: "True"` means the condition **is present/active**; `status: "False"` means it **is not**. This code only runs when the sandbox was externally deleted, so the condition should read `"status": "True"`.

Current:
```go
"status": "False",
```

Should be:
```go
"status": "True",
```

`SandboxDeleted=False` will be read by any controller or UI code as "sandbox was *not* deleted" — the opposite of reality.

— *Amber*

jsell-rh (approved):
🤖 **Amber — Review**

The code is correct and appropriately scoped. Approving the implementation with one minor note, plus a coordination flag for the team.

**Implementation looks good:**
- `updateSessionPhaseWithConditions()` refactor is clean — `updateSessionPhase` now delegates correctly, no behavior change on existing callers
- Structured `SandboxDeleted/DeletedExternally` condition gives the UI actionable failure context instead of a blank `Failed` state
- 5s poll interval is a reasonable improvement; no performance concern at this cardinality

**Minor — ignored error from `json.Marshal`:**
```go
condJSON, _ := json.Marshal(...)
```
The input is a hardcoded `[]map[string]string`, so this cannot fail in practice. Convention still says handle it:
```go
condJSON, err := json.Marshal(...)
if err != nil {
    s.logger.Error().Err(err).Msg("failed to marshal condition; skipping condition update")
}
```
Not blocking merge, but worth cleaning up.

---

**⚠️ Team coordination note — overlap with PR #307:**

These `pod_sync.go` changes are identical to a subset of PR #307 (commit `e61dce1`). Both target main. If #324 merges first, #307 will have a trivially-resolvable conflict in `pod_sync.go` on rebase — not a blocker, but worth the author knowing so the rebase isn't a surprise. If the intent is to decompose #307 into smaller PRs, this is a good start; the remaining blockers in #307 (Python SDK defaults, gateway e2e) can land separately.

— *Amber*

jsell-rh (changes requested):
🤖 **Amber — Initial Review**

The `amber/approved` label appears to have been applied without a prior Amber review. Doing that review now.

**Summary:** Focused, narrow change — polling interval reduction and structured condition on external sandbox deletion. The refactor of `updateSessionPhase` into `updateSessionPhaseWithConditions` is clean. Two real issues to fix before merge, plus an important overlap concern with #307.

---

### Blocker — `"status": "False"` is semantically inverted

In K8s condition conventions, the `status` field answers "is this condition true right now?" For a condition type named `SandboxDeleted`, the correct status when the sandbox *has been deleted* is `"True"`, not `"False"`. Using `"False"` means "the sandbox is NOT deleted" — the exact opposite of what you're trying to communicate.

```go
// ❌ Current — status "False" means "SandboxDeleted is not happening"
condJSON, _ := json.Marshal([]map[string]string{{
    "type":    "SandboxDeleted",
    "status":  "False",   // ← wrong: inverts the meaning
    ...
}})

// ✅ Fix
condJSON, _ := json.Marshal([]map[string]string{{
    "type":    "SandboxDeleted",
    "status":  "True",    // sandbox was deleted = condition is true
    ...
}})
```

Compare: K8s `Ready=True` means "pod is ready." K8s `DiskPressure=True` means "there is disk pressure." `SandboxDeleted=True` means "sandbox was deleted." This is a correctness issue — the UI and any consumers reading conditions will get the wrong status signal.

---

### Major — `json.Marshal` error silently swallowed

```go
condJSON, _ := json.Marshal(...)
```

CLAUDE.md: *"Never silently swallow partial failures."* Even though this specific call marshaling a `[]map[string]string` won't realistically fail, the pattern is explicitly forbidden by project convention. Either handle the error or use `//nolint:errcheck` with a comment explaining why failure is impossible here. The former is preferred:

```go
condJSON, err := json.Marshal([]map[string]string{{...}})
if err != nil {
    s.logger.Error().Err(err).Str("session_id", session.ID).Msg("failed to marshal sandbox condition")
    s.updateSessionPhaseWithConditions(ctx, sdk, session, PhaseFailed, "")
    return
}
```

---

### Important — Overlap with PR #307

This PR addresses part of issue #223 (poll interval 15s→5s, structured condition on external deletion). PR #307 addresses a **superset** of the same issue: it makes the same `pod_sync.go` changes AND fixes the resource leaking bug (sessions never terminating in gateway mode) AND adds `stop_on_run_finished`.

If #307 merges first, this PR will conflict. If this PR merges first, #307's `pod_sync.go` hunk will need to be rebased. The teams should coordinate — either:
1. Close this PR and cherry-pick the `pod_sync.go` fix into #307, or
2. Merge this first and rebase #307 on top

Not a blocker from a code correctness standpoint, but worth flagging before someone puts energy into a conflict resolution.

---

**Verdict:** Changing to `amber/change-requested` for the two issues above. Blocker 1 is a correctness issue that will produce wrong condition values in the UI. Major 2 is a project convention violation. The underlying approach is sound — just needs these two fixes.

Happy to re-review on next push.

— *Amber*

jsell-rh (changes requested):
🤖 **Amber — PR Review**

The refactor is clean and the intent is correct — surfacing structured failure context when a sandbox is deleted externally is exactly the right behaviour. The `updateSessionPhaseWithConditions()` extraction is a nice improvement. Two blockers and one minor note before I can approve.

---

## Blocker 1 — Condition `status` value inverts K8s semantics (Critical)

**File:** `pod_sync.go`, `syncSandboxStatus()`, the new `condJSON` block

```go
condJSON, _ := json.Marshal([]map[string]string{{
    "type":    "SandboxDeleted",
    "status":  "False",    // ← wrong
    "reason":  "DeletedExternally",
    "message": "Sandbox was deleted outside of the control plane lifecycle. Session marked as failed.",
}})
```

In Kubernetes condition conventions (and the project's own usage in `buildFailureConditions`), `status: "True"` means the condition is **active/present**. `status: "False"` on `SandboxDeleted` reads as "the sandbox has NOT been deleted" — the opposite of what happened. Change to `"True"`.

---

## Blocker 2 — `json.Marshal` error silently discarded (Minor violation, but project convention requires it)

```go
condJSON, _ := json.Marshal(...)   // ← error ignored
```

From CLAUDE.md: *"Never silently swallow partial failures — every error path must propagate or be collected."* `json.Marshal` on a `[]map[string]string` literal is effectively infallible in practice, but ignoring the error violates the project's explicit standard. The fix is a one-liner:

```go
condJSON, err := json.Marshal([]map[string]string{{...}})
if err != nil {
    s.logger.Error().Err(err).Str("session_id", session.ID).Msg("failed to marshal SandboxDeleted condition")
    condJSON = []byte{}
}
s.updateSessionPhaseWithConditions(ctx, sdk, session, PhaseFailed, string(condJSON))
```

---

## Missing: Tests for the new external-deletion path (Major)

`updateSessionPhaseWithConditions()` is now callable two ways:
1. Via `updateSessionPhase()` → conditions built from a pod object (existing path, covered by tests)
2. Directly from `syncSandboxStatus()` with a pre-built `condJSON` (new path, **not covered**)

The new path in `syncSandboxStatus()` — the `GetSandbox() NotFound` branch — has no test coverage. Please add a table-driven test case (or a dedicated `TestSyncSandboxStatus_ExternallyDeletedSandbox`) that verifies:
- The session is marked `PhaseFailed`
- The `SandboxDeleted/DeletedExternally` condition is written to the patch

---

## Informational — Overlap with PR #307

PR #307 contains a superset of these changes (same `syncSandboxStatus` fix + `updateSessionPhaseWithConditions` refactor + resource-leak fix). If #307 gets its remaining issues resolved and merges first, this PR will conflict. Worth coordinating with #307's author or noting the dependency explicitly in the PR body.

---

**Checklist result:**

| Check | Status |
|-------|--------|
| No `panic()` in production code | ✅ |
| No tokens in logs | ✅ |
| `unstructured.Nested*` used correctly | ✅ |
| Errors wrapped with context | ✅ |
| `IsNotFound` handled | ✅ |
| Condition `status` value correct | ❌ Should be `"True"` |
| Error paths not swallowed | ❌ `json.Marshal` error ignored |
| Tests for new code paths | ❌ Missing for external-deletion branch |

Happy to re-review once these are addressed — the core approach is sound.

— *Amber*

PR #325, opened 2026-07-10
feat: add mock LLM server for self-contained e2e testing

bsquizz:
## Summary

Adds a mock LLM inference service and end-to-end test tooling so gateway e2e tests can run in kind clusters without real LLM API keys.

- **Mock LLM server** (`tests/mock-llm/`): OpenAI-compatible `/v1/chat/completions` endpoint that echoes back canned responses
- **Agent-overridable `ANTHROPIC_BASE_URL`**: removed from `immutableSandboxEnvKeys`, `inferenceExecEnv()` reads from agent environment, runner `auth.py` uses `setdefault` — lets `test-agent-mock-llm` point at the mock server
- **E2E test verification** (`tests/e2e/gateway-e2e-test.sh`): switched to `test-agent-mock-llm` agent, added section 11 that polls `acpctl session messages` and verifies user prompt delivery and mock LLM response content
- **DNS patch retry**: `patchSandboxDNSConfig` now retries 5 times with linear backoff
- **Example manifests**: agent, provider, policy, and credential for mock-llm in `examples/`
- **Makefile**: `build-mock-llm` / `kind-load-mock-llm` targets, auto-deploy during `kind-up`, `kind-reload-runner-openshell` switched to `ctr import` with unique tags
- **Kind setup**: `setup-kind-openshell.sh` provisions `mock-llm-creds` secret in tenant namespaces
- **Spec**: `specs/platform/e2e-test-tooling.spec.md` documenting the design

## Test plan

- [ ] `make kind-up` deploys mock-llm and provisions mock-llm-creds secrets
- [ ] Create a session using `test-agent-mock-llm` — verify it routes to the mock LLM and gets a canned response
- [ ] Verify existing agents still default to `https://inference.local` when no `ANTHROPIC_BASE_URL` override is set
- [ ] Verify DNS config patch retries on transient failure instead of failing immediately
- [ ] Run `./tests/e2e/gateway-e2e-test.sh` — section 11 verifies LLM messages end-to-end

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -966,6 +966,11 @@ kind-up: preflight-cluster build-cli ## Start kind cluster and deploy the platfo
 		ANTHROPIC_VERTEX_PROJECT_ID="$(ANTHROPIC_VERTEX_PROJECT_ID)" \
 		CLOUD_ML_REGION="$(CLOUD_ML_REGION)" \
 		./scripts/setup-kind-openshell.sh; \
+		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Building and deploying mock LLM server..."; \
+		$(MAKE) --no-print-directory kind-load-mock-llm; \
+		kubectl apply -k tests/mock-llm/manifests/ $(QUIET_REDIRECT); \
+		kubectl rollout status deployment/mock-llm -n $(NAMESPACE) --timeout=60s; \
+		echo "$(COLOR_GREEN)✓$(COLOR_RESET) Mock LLM server deployed"; \
 		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Applying tenant fleet definitions via acpctl..."; \
 		ACPCTL=components/ambient-cli/acpctl; \
 		PF_PORT=18766; \
@@ -1340,16 +1345,19 @@ kind-reload-runner-openshell: check-kind check-kubectl check-local-context ## Re
 		--build-arg GIT_COMMIT=$(shell git rev-parse HEAD) \
 		-t $(RUNNER_OPENSHELL_IMAGE) . $(QUIET_REDIRECT) || \
 		{ echo "$(COLOR_RED)✗$(COLOR_RESET) Build failed. Run without QUIET=1 for full output."; exit 1; }
-	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Loading OpenShell runner into kind cluster..."
-	@_IMG=$$(echo "$(RUNNER_OPENSHELL_IMAGE)" | cut -d: -f1) && \
-		$(CONTAINER_ENGINE) tag $(RUNNER_OPENSHELL_IMAGE) localhost/$$_IMG:latest && \
-		kind load docker-image localhost/$$_IMG:latest --name $(KIND_CLUSTER_NAME) && \
-		echo "$(COLOR_GREEN)✓$(COLOR_RESET) OpenShell runner image loaded as localhost/$$_IMG:latest"
-	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Updating control plane env vars to use localhost/acp_runner_openshell:latest..."
-	@kubectl set env deployment/ambient-control-plane -n $(NAMESPACE) \
-		OPENSHELL_RUNNER_IMAGE=localhost/acp_runner_openshell:latest $(QUIET_REDIRECT)
-	@kubectl rollout status deployment/ambient-control-plane -n $(NAMESPACE) --timeout=60s
-	@echo "$(COLOR_GREEN)✓$(COLOR_RESET) OpenShell runner reloaded — new sessions will use the updated image"
+	@_TAG=$$(git rev-parse --short HEAD)-$$(date +%s) && \
+	_IMG=$$(echo "$(RUNNER_OPENSHELL_IMAGE)" | cut -d: -f1) && \
+	$(CONTAINER_ENGINE) tag $(RUNNER_OPENSHELL_IMAGE) localhost/$$_IMG:$$_TAG && \
+	echo "$(COLOR_BLUE)▶$(COLOR_RESET) Loading localhost/$$_IMG:$$_TAG into kind cluster ($(KIND_CLUSTER_NAME))..." && \
+	$(CONTAINER_ENGINE) save localhost/$$_IMG:$$_TAG | \
+		$(CONTAINER_ENGINE) exec -i $(KIND_CLUSTER_NAME)-control-plane \
+		ctr --namespace=k8s.io images import - && \
+	echo "$(COLOR_GREEN)✓$(COLOR_RESET) OpenShell runner image loaded as localhost/$$_IMG:$$_TAG" && \
+	echo "$(COLOR_BLUE)▶$(COLOR_RESET) Updating control plane env vars to use localhost/$$_IMG:$$_TAG..." && \
+	kubectl set env deployment/ambient-control-plane -n $(NAMESPACE) \
+		OPENSHELL_RUNNER_IMAGE=localhost/$$_IMG:$$_TAG $(QUIET_REDIRECT) && \
+	kubectl rollout status deployment/ambient-control-plane -n $(NAMESPACE) --timeout=60s && \
+	echo "$(COLOR_GREEN)✓$(COLOR_RESET) OpenShell runner reloaded (tag: $$_TAG)"
 
 kind-load-runner: build-runner-openshell check-kind check-kubectl check-local-context ## Build OpenShell runner image and load into kind (skips if already present)
 	@_IMG=$$(echo "$(RUNNER_OPENSHELL_IMAGE)" | cut -d: -f1) && \
@@ -1370,6 +1378,18 @@ kind-load-runner: build-runner-openshell check-kind check-kubectl check-local-co
 		fi; \
 		echo "$(COLOR_GREEN)✓$(COLOR_RESET) Runner image loaded: $$_REF"; \
 	fi
+
+build-mock-llm: ## Build mock LLM server image for e2e testing
+	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Building mock-llm image..."
+	@$(CONTAINER_ENGINE) build $(PLATFORM_FLAG) -t localhost/mock-llm:latest tests/mock-llm $(QUIET_REDIRECT)
+	@echo "$(COLOR_GREEN)✓$(COLOR_RESET) mock-llm image built"
+
+kind-load-mock-llm: build-mock-llm check-kind check-kubectl check-local-context ## Build and load mock LLM image into kind
+	@echo "$(COLOR_BLUE)▶$(COLOR_RESET) Loading mock-llm into kind cluster ($(KIND_CLUSTER_NAME))..."
+	@$(CONTAINER_ENGINE) save localhost/mock-llm:latest | \
+		$(CONTAINER_ENGINE) exec -i $(KIND_CLUSTER_NAME)-control-plane \
+		ctr --namespace=k8s.io images import -
+	@echo "$(COLOR_GREEN)✓$(COLOR_RESET) mock-llm image loaded"
 
 kind-sso-toggle: check-kubectl ## Toggle SSO auth on/off in Kind (affects both frontend and backend)
 	@UNLEASH_ADMIN_TOKEN=$$(kubectl get secret unleash-credentials -n $(NAMESPACE) -o jsonpath='{.data.admin-api-token}' | base64 -d); \
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -466,7 +466,7 @@ func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, sess
 		if err := r.patchSandboxDNSConfig(ctx, namespace, sbxName); err != nil {
 			r.logger.Warn().Err(err).Str("sandbox", sbxName).Msg("failed to patch sandbox dnsConfig; DNS resolution for external FQDNs may fail")
 		}
-		execEnv := r.inferenceExecEnv()
+		execEnv := r.inferenceExecEnv(agent)
 		var payloads []types.Payload
 		if agent != nil {
 			payloads = agent.Payloads
@@ -541,7 +541,7 @@ func (r *SimpleKubeReconciler) provisionSessionSandbox(ctx context.Context, sess
 		Int("providers", len(providerNames)).
 		Msg("sandbox created via gateway")
 
-	execEnv := r.inferenceExecEnv()
+	execEnv := r.inferenceExecEnv(agent)
 	var payloads []types.Payload
 	if agent != nil {
 		payloads = agent.Payloads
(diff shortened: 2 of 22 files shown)


jsell-rh (changes requested):
🤖 **Amber — Spec Review**

Good overall structure — GIVEN/WHEN/THEN scenarios are clear, the PR #318 dependency is correctly called out, and the security posture (non-root, drop ALL, readOnlyRootFilesystem) matches ACP conventions. Two design gaps need resolution before implementation starts.

---

**Blocker 1 — API Format Mismatch: OpenAI ≠ Anthropic Messages API**

`ANTHROPIC_BASE_URL` routes Claude Code's calls to the specified base URL — but Claude Code speaks the **Anthropic Messages API** (`POST /v1/messages`), not the OpenAI Chat Completions API (`POST /v1/chat/completions`). These are not wire-compatible.

If the mock only implements `/v1/chat/completions`, Claude Code will GET a 404 or an invalid response on `/v1/messages` and the runner will exit with an error before any "plumbing" can be validated.

The mock needs to implement `POST /v1/messages` with the Anthropic response envelope:

```json
{
  "id": "msg_mock-<uuid>",
  "type": "message",
  "role": "assistant",
  "content": [{"type": "text", "text": "Mock response: <last_user_message>"}],
  "model": "<requested-model>",
  "stop_reason": "end_turn",
  "usage": {"input_tokens": 0, "output_tokens": 0}
}
```

The OpenAI `/v1/chat/completions` endpoint can stay for forward-compatibility but isn't what Claude Code will call.

---

**Blocker 2 — Missing Streaming Support**

The spec says streaming is "unnecessary for e2e plumbing validation," but Claude Code sends `stream: true` by default. When it does, it reads the response as an SSE stream. A plain JSON body returned without `Content-Type: text/event-stream` and `data:` framing will cause a parse error at the Claude Code layer — the runner process exits non-zero before the plumbing under test has a chance to be observed.

The minimal fix is an SSE wrapper for `/v1/messages`:

```
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"Mock response"}}\n\n
data: {"type":"message_delta","delta":{"stop_reason":"end_turn"}}\n\n
data: {"type":"message_stop"}\n\n
```

This is ~20 lines of FastAPI code and makes the mock actually exercise the full response path the spec promises.

---

**Minor — `readOnlyRootFilesystem: true` needs an emptyDir for `/tmp`**

Python/uvicorn occasionally writes to `/tmp` (bytecode cache, upload spill). Add an `emptyDir` volume mounted at `/tmp` to avoid silent runtime failures:

```yaml
volumes:
  - name: tmp
    emptyDir: {}
volumeMounts:
  - name: tmp
    mountPath: /tmp
```

---

**Minor — `seccompProfile` missing from SecurityContext**

ACP restricted SecurityContext includes `seccompProfile: {type: RuntimeDefault}`. The spec should add this to the Deployment SecurityContext to match the convention and stay in line with what CI will enforce.

---

The streaming + API format issues are pre-implementation design gaps that are worth resolving in the spec now rather than discovering at implementation time. Happy to re-review once those two sections are updated.

— *Amber*

jsell-rh (changes requested):
🤖 **Amber — Full Review**

The overall structure is solid — correct SecurityContext requirements, idempotent secret provisioning, accurate `ProviderCredentialsFromSecret()` description, right `imagePullPolicy: IfNotPresent` for locally-loaded images, and clean GIVEN/WHEN/THEN scenarios throughout. One critical design gap and one minor clarification needed before this ships.

---

## Critical — API format mismatch

**The mock implements OpenAI format; Claude Code uses Anthropic format.**

The spec configures `ANTHROPIC_BASE_URL=http://mock-llm.ambient-code.svc.cluster.local:8000` on the agent. When Claude Code's Anthropic SDK follows that base URL, it appends `/v1/messages` (Anthropic Messages API) — not `/v1/chat/completions` (OpenAI format). The mock only serves `/v1/chat/completions`, so every Claude Code API call will receive a **404**, and the e2e tests will fail before exercising any platform plumbing.

The spec also explicitly says:

> The server SHALL NOT implement … the Anthropic Messages API.

That statement is fine on its own, but it's incompatible with using `ANTHROPIC_BASE_URL` to redirect Claude Code's calls there.

**Two paths to fix this:**

1. **Implement `/v1/messages`** — swap the mock endpoint from OpenAI format to the Anthropic Messages API format. This is what `ANTHROPIC_BASE_URL` actually routes to. The response body is straightforward:
   ```json
   {
     "id": "mock-<uuid>",
     "type": "message",
     "role": "assistant",
     "content": [{"type": "text", "text": "Mock LLM response: <last_user_message>"}],
     "model": "<requested-model>",
     "stop_reason": "end_turn",
     "usage": {"input_tokens": 0, "output_tokens": 0}
   }
   ```
   Keep `/v1/chat/completions` as a bonus endpoint if you want OpenAI compatibility too, but the primary must be `/v1/messages` for Claude Code to work.

2. **Drop `ANTHROPIC_BASE_URL`** — if the goal is purely plumbing validation (env injection, credential passthrough) and not actual LLM roundtrip, remove the mock LLM call from the agent path entirely. But then the spec's "Purpose" section needs to be updated — it currently says "enables fully self-contained e2e testing of the **LLM response path**."

I'd recommend option 1.

---

## Minor — `CLAUDE_CODE_ATTRIBUTION_HEADER: "0"` undocumented

The spec lists this env var without explaining what it does or why it's needed for e2e testing. A one-sentence note (e.g., "Disables attribution reporting headers to avoid noise in test runs") would help implementers understand if they need to carry this forward.

---

## What's good (no action needed)

- SecurityContext requirements are correct and match CLAUDE.md conventions
- `--dry-run=client | kubectl apply -f -` idempotent secret creation pattern is right
- `ProviderCredentialsFromSecret()` generic passthrough description is accurate
- Platform-injected rules (`_acp_internal`, `_mlflow_rh`) correctly noted as server-side only
- PR #318 dependency clearly scoped — other requirements can land independently
- Image loading pattern matches existing `kind-reload-component` macro
- Index update is clean

Happy to re-review once the API format issue is resolved.

— *Amber*

jsell-rh (approved):
🤖 **Amber — Re-review after `f8c99b1` (2026-07-10T16:51Z)**

All four blockers from my previous two reviews are resolved.

| Blocker | Status |
|---------|--------|
| API format mismatch — mock was OpenAI `/v1/chat/completions`, Claude Code calls Anthropic `/v1/messages` | ✅ Fixed — spec now specifies `/v1/messages` with Anthropic envelope throughout |
| Missing streaming support | ✅ Fixed — full SSE event sequence (`message_start` → `content_block_start` → `content_block_delta` → `content_block_stop` → `message_delta` → `message_stop`) is now required |
| `readOnlyRootFilesystem: true` needs `emptyDir /tmp` | ✅ Fixed — `emptyDir` volume at `/tmp` is now in the Deployment spec |
| Missing `seccompProfile: RuntimeDefault` | ✅ Fixed — added to SecurityContext |

**One nit (non-blocking, fix at implementation time):**

`specs/platform/index.spec.md` line 49 still says `"OpenAI-compatible \`/v1/chat/completions\`"` in the new pointer paragraph — a stale artifact from before the API format switch. The spec itself is correct; the index blurb just needs updating to say `"Anthropic Messages API \`/v1/messages\`"` when this goes to merge.

Everything else holds: SecurityContext requirements match CLAUDE.md conventions, idempotent secret provisioning is correct, `ProviderCredentialsFromSecret()` description is accurate, the PR #318 dependency is clearly scoped, and all GIVEN/WHEN/THEN scenarios are now testable. The `CLAUDE_CODE_ATTRIBUTION_HEADER: "0"` purpose is now documented inline on the agent spec.

Approving. This is a draft — remember to mark ready for review when implementation is complete.

— *Amber*

jsell-rh:
🤖 **Amber — Spot-check of final commit (`f8c99b1`)**

`amber/approved` stands. One minor nit I noticed while doing a pass:

**Minor — `index.spec.md` summary is stale.** The index entry still says "OpenAI-compatible `/v1/chat/completions`" but commit `f8c99b1` switched the spec to the Anthropic Messages API (`POST /v1/messages`). The index is a one-liner — worth a quick fixup before merge:

```
- Mock LLM inference service for self-contained e2e testing in kind. Covers the mock server (OpenAI-compatible `/v1/chat/completions`), ...
+ Mock LLM inference service for self-contained e2e testing in kind. Covers the mock server (Anthropic Messages API `/v1/messages` with streaming SSE), ...
```

Not a blocker — the spec itself is correct. Approval stands.

— *Amber*

jsell-rh (approved):
🤖 **Amber — Re-review of commit `91a731dc` (implementation landing)**

`amber/approved` stands. The implementation matches the spec exactly and CI is green on all three Konflux pipelines. Here's my pass:

| Area | Finding |
|---|---|
| `inferenceExecEnv(agent)` — reads `ANTHROPIC_BASE_URL` from agent env, falls back to `inference.local` | ✅ Correct. Guard `v != ""` prevents accidentally clearing to empty string. |
| Removed `ANTHROPIC_BASE_URL` from `immutableSandboxEnvKeys` | ✅ Intentional and documented in `openshell-sandbox-provisioning.spec.md`. See security note below. |
| `auth.py` — `setdefault` preserves agent-provided URL | ✅ Correct. Control plane injects before runner starts; `setdefault` preserves it. |
| `patchSandboxDNSConfig` retry logic (5 retries, linear 500ms backoff) | ✅ Solid. One nit: the sleep doesn't check `ctx.Done()` — for ≤2s total it's acceptable, but a `select` would be cleaner if this context ever gets shorter timeouts. Not a blocker. |
| Mock LLM deployment manifest SecurityContext | ✅ `runAsNonRoot`, `readOnlyRootFilesystem`, `drop: ALL`, `seccompProfile: RuntimeDefault`, `emptyDir /tmp`. Fully compliant. |
| `server.py` — Anthropic Messages API + SSE streaming | ✅ Event sequence (`message_start` → `content_block_start` → `content_block_delta` → `content_block_stop` → `message_delta` → `message_stop`) matches the SDK's expected wire format. |
| `kind-reload-runner-openshell` — `ctr import` with unique `SHA+timestamp` tags | ✅ Better than `kind load docker-image` — avoids layer cache aliasing on rapid rebuilds. |
| `requirements.txt` — no version pins (`fastapi`, `uvicorn[standard]`) | Minor. Test-only container, but floating deps mean silent breakage if upstream releases a breaking change. Consider pinning before this PR graduates from draft. Not blocking. |
| `credential-mock-llm.yaml` — `token: mock-llm-token` in git | ✅ Synthetic value, example path, pre-commit scanner already evaluated it (commit `80e78e18`). Fine. |

**Security observation (deliberate, documented):**

Removing `ANTHROPIC_BASE_URL` from `immutableSandboxEnvKeys` is an intentional design decision — documented explicitly in the updated spec. Worth being explicit about the new boundary: any principal who can write `Agent` resources can redirect that agent's inference traffic to an arbitrary URL. In the current RBAC model (operators create agents, users create sessions against them) this is acceptable. If user-writable agents are ever introduced, this will need an allowlist or validation step. The spec's `SHALL NOT be in the immutableSandboxEnvKeys set` language is the right place to track this intent.

CI: `control-plane-main-on-pull-request` ✅, `acp-runner-main-on-pull-request` ✅, `acp-runner-openshell-main-on-pull-request` ✅.

— *Amber*

jsell-rh (commented):
## 🤖 Amber — Follow-up on `fb14f98` (2026-07-14T03:36Z)

My approval from `91a731dc` stands. This is the final implementation commit wiring up the mock LLM to the gateway e2e test.

**What changed:**

- Gateway e2e test now uses `test-agent-mock-llm` instead of `hello-world` — the right call, since `hello-world` requires real credentials and the whole point of the mock LLM is self-contained e2e
- New **section 11** polls `acpctl session messages` (up to 120s, 2s interval) and verifies three things: (a) message count > 0, (b) a `user` event type exists, (c) an `assistant`/`TEXT_MESSAGE_CONTENT`/`MESSAGES_SNAPSHOT` event exists
- 11c (mock echo content match) is a `skip` not a `fail` if the text isn't found — that's an honest acknowledgment that event format may vary

**Analysis:**

- The CLAUDE.md payload grep changed from `"hello"` → `"mock LLM"` to match the `test-agent-mock-llm` agent's CLAUDE.md content. Consistent. ✅
- The env var assertion switched from `ENV_NAME=test` to `CLAUDE_CODE_ATTRIBUTION_HEADER=0`. This validates that agent-env passthrough works with the actual vars the mock-llm agent declares. ✅
- 60-iteration × 2s poll = 120s window for messages to appear. Generous enough for a Kind cluster startup. ✅
- The `skip` fallback on 11c is correct — if the mock LLM response arrives in a streaming `content_block_delta` event rather than a top-level `payload` field, the jq expression `.[].payload` won't find it. This is a format discovery issue, not a failure. The `skip` message is honest.

**One nit (non-blocking):**

The `jq -r '[.[] | .payload] | join(" ")'` expression in 11c will concatenate all `null` values as the string `"null"` if no message has a `payload` field — meaning the grep for `"Mock LLM response"` won't match but also won't error. This is fine because 11c is `skip`-not-fail, but if you ever promote it to a hard assertion, you'll want `[.[] | .payload // empty]` to filter nulls first.

**Summary:** This commit correctly closes the loop on the mock LLM spec — the test now exercises the full path the spec promised (session creation → sandbox provisioning → mock LLM roundtrip → message delivery verification). No new concerns.

*— Amber*

jsell-rh (commented):
## 🤖 Amber — CI failure on `a2400bb` (2026-07-14T03:36Z)

My approval from `91a731dc` stands on the implementation quality, but the latest commit introduced CI failures that need to be addressed before this can merge.

**Failing gates on `a2400bb`:**

| Check | Result |
|---|---|
| `Build CI Gate` | ❌ failure |
| `test-local-dev / Local Dev CI Gate` | ❌ failure |
| `test-local-dev / test-local-dev-simulation` | ❌ failure |
| `test-local-dev / validate-manifests` | ✅ success |

Konflux pipelines (control-plane, runner, runner-openshell) all passed ✅ — so the build artifacts are fine. The failure is isolated to the `test-local-dev-simulation` job, which exercises the local Kind flow. The new `test-agent-mock-llm` agent or the `setup-kind-openshell.sh` mock-llm-creds provisioning path is the likely culprit — the simulation probably doesn't have the mock-llm secret available in the tenant namespaces it sets up.

Please investigate and re-run (or fix) `test-local-dev-simulation` before marking ready for review. This was passing before the e2e section-11 wiring commit, so the regression scope is narrow.

— *Amber*

jsell-rh:
## 🤖 Amber — Follow-up on commits `2fc85f5`, `3c7153d`, `f5f90f3`

My approval from `91a731dc` stands on implementation quality. Three commits landed since my last CI comment — good fixes on the polling logic and port-forward reliability — but **CI is still failing** on the latest head (`f5f90f3`).

**Failing gates:**

| Check | Result |
|---|---|
| `Build CI Gate` | ❌ failure |
| `test-local-dev / Local Dev CI Gate` | ❌ failure |
| `test-local-dev / test-local-dev-simulation` | ❌ failure |

**What the commits addressed:**
- `2fc85f5` — correct: polling until `assistant` message instead of any message avoids the race where the user message triggers an early break before the LLM roundtrip completes ✅
- `3c7153d` — correct: soft-deleted project purge + `_ensure_gateway_port_forward()` re-establishment are legitimate fixes for flaky reruns ✅
- `f5f90f3` — correct call: network tests should fail hard when the gateway is unreachable rather than silently skip. Masking infra issues as `skip` hides real failures ✅

**Still failing:** `test-local-dev-simulation` was failing on `a2400bb` and continues to fail on `f5f90f3`. The commit message on `3c7153d` adds `_ensure_gateway_port_forward()` which suggests the gateway port-forward was the root cause — but the simulation job is apparently still not reaching a green state.

Please share the simulation failure output so I can help diagnose. The implementation is correct; the CI health is the only remaining gate.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — Re-review on `f5f90f3` (2026-07-14)

My approval from `91a731dc` stands, and the three new commits since my last review (`a2400bb`) are all correct fixes. Here's the delta:

| Commit | Change | Assessment |
|---|---|---|
| `2fc85f5` | Poll for *assistant* message, not any message; 90-iter × 2s = 180s window | ✅ Correct — the previous exit-on-any-message false-positive was the right thing to fix. `LLM_RESPONSE_FOUND` now drives both the poll break and the 11b assertion, eliminating the redundant jq eval. |
| `3c7153d` | Purge soft-deleted projects before `acpctl apply`; `_ensure_gateway_port_forward()` for section 13 | ✅ Correct — soft-delete 409 is a real re-run trap. Gateway port-forward recovery logic is sound: poll log for port, re-register certs, verify with `openshell sandbox list`. |
| `f5f90f3` | Promote skip→fail when gateway port-forward unavailable | ✅ Correct — the prior commit's `skip` was too lenient; section 13 is a required test section and should hard-fail on infrastructure issues so they don't hide. |

**One nit (non-blocking, same as before):** The DB purge in `3c7153d` execs into a pod labelled `app=ambient-api-server,component=database`. That label selector looks like it's targeting a sidecar or a co-located postgres container — worth a comment confirming this is intentional vs. a standalone DB pod. Not a blocker; operational concern for the infra team.

**CI is pending** — previous `test-local-dev-simulation` failure was the blocker I flagged in my last comment. The commits address the underlying causes (soft-deleted project 409, stale port-forward). Waiting on CI to confirm the fix.

`amber/approved` label stands. Ready to merge once CI goes green.

— *Amber*

jsell-rh:
🤖 **Amber — Follow-up on `6424c0b` (2026-07-14T18:03Z)**

My approval from `5c0bfff` stands. This is a one-line targeted fix.

`acpctl login --url "$API_URL" --token "$TOKEN" --project "$TENANT"` — correct. The missing `--project` flag caused `acpctl session messages` to fail with "no project set" in CI where `AMBIENT_PROJECT` is not set in the environment. The error was swallowed by `2>/dev/null`, causing the 180s polling loop to always see zero messages and time out. Adding `--project "$TENANT"` (where `$TENANT` is already set earlier in the script) is the right root-cause fix.

No new concerns.

— *Amber*

PR #326, opened 2026-07-10
feat(ui): add phase-aware empty states and thinking indicator to chat

bsquizz:
## Summary
- Add phase-aware empty states to the chat tab and sidebar — spinner for Pending/Creating/Running sessions, static message for terminal sessions
- Add a pulsing three-dot thinking indicator when the agent is actively processing, driven by `run_started`/`run_finished` lifecycle events
- Add spec coverage in `specs/ui/views.spec.md` for both features

## Test plan
- [ ] Verify spinner shows "Runner is starting..." for Pending/Creating sessions
- [ ] Verify spinner shows "Runner started. Waiting for messages..." for Running sessions with no messages
- [ ] Verify terminal sessions show "No conversation messages were recorded."
- [ ] Verify thinking dots appear after user sends a message and disappear when assistant responds
- [ ] Verify thinking indicator works in both chat tab and sidebar

### DIFF
--- a/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/chat-tab.tsx
+++ b/components/ambient-ui/src/app/(dashboard)/[projectId]/sessions/[sessionId]/_components/chat-tab.tsx
@@ -11,6 +11,7 @@ import {
   ChatItemsList,
   ChatInput,
   buildChatItems,
+  isRunActive,
 } from '@/components/chat-messages'
 import { useChatSidebar } from '@/components/chat-sidebar-context'
 import { useLiveTail, LiveIndicator } from './live-tail-indicator'
@@ -25,6 +26,10 @@ export function ChatTab({ session }: { session: DomainSession }) {
   const chatItems = useMemo(() => {
     return buildChatItems(data?.items ?? [])
   }, [data])
+
+  const isThinking = useMemo(() => {
+    return session.phase === 'Running' && isRunActive(data?.items ?? [])
+  }, [data, session.phase])
 
   const chatItemCount = chatItems.length
 
@@ -126,7 +131,7 @@ export function ChatTab({ session }: { session: DomainSession }) {
             aria-label="Chat messages"
             onScroll={handleScroll}
           >
-            <ChatItemsList items={chatItems} isLoading={isLoading} />
+            <ChatItemsList items={chatItems} isLoading={isLoading} phase={session.phase} isThinking={isThinking} />
             <div ref={sentinelRef} className="h-1" aria-hidden="true" />
           </div>
 
--- a/components/ambient-ui/src/app/globals.css
+++ b/components/ambient-ui/src/app/globals.css
@@ -238,3 +238,8 @@
     @apply bg-background text-foreground antialiased;
   }
 }
+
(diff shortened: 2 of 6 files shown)


jsell-rh (changes requested):
## 🤖 Amber Review

**Summary:** Clean, well-structured UI feature. The phase-aware empty states and thinking indicator are a nice UX improvement. The implementation is tight and follows component conventions. One blocker (missing tests for non-trivial logic), a minor type-safety note, and a small spec deviation to address before merge.

**ACP Checklist:**
- ✅ Zero `any` types
- ✅ Shadcn-compatible components (Loader2/Bot icons, standard divs)
- ✅ React Query for data — no manual `fetch()` introduced
- ✅ `useMemo` used correctly for `isThinking` derivation
- ✅ Loading/error states extended (the whole point of this PR)
- ✅ Aria labels on animated elements
- ❌ **No unit tests for `isRunActive`** — non-trivial reverse-scan logic, Constitution Principle IV violation
- ⚠️ `phase?: string` is loose — minor type safety improvement available
- ⚠️ Thinking indicator triggers on `user` message, not `run_started` — minor spec deviation

See inline comments for specifics.

jsell-rh (commented):
### Inline findings

**[CRITICAL] `components/ambient-ui/src/components/chat-messages.tsx` — `isRunActive` has no tests**

`isRunActive` is a non-trivial reverse scan with multiple exit conditions (returns `true` on `user`, `false` on `assistant`, parses lifecycle JSON, falls off the end returning `false`). This logic determines whether the thinking indicator appears — a wrong return in any edge case causes a stuck or missing indicator. Constitution Principle IV requires tests before implementation.

Suggested test cases to add (e.g. `chat-messages.test.ts`):
```ts
describe('isRunActive', () => {
  it('returns true when last relevant event is user message', ...)
  it('returns false when last relevant event is assistant message', ...)
  it('returns true when run_started lifecycle follows user message', ...)
  it('returns false when run_finished lifecycle received', ...)
  it('returns false on empty message list', ...)
  it('indicator persists through tool_use/tool_result messages', ...)
  it('handles malformed lifecycle payload gracefully', ...)
})
```

---

**[Minor] `components/ambient-ui/src/components/chat-messages.tsx` — `phase?: string` is loose**

The `phase` prop accepts any string. ACP's domain model has a finite set of phases. Using a union type makes invalid phases a compile-time error and makes the component's intent self-documenting:

```ts
type SessionPhase = 'Pending' | 'Creating' | 'Running' | 'Completed' | 'Failed' | 'Stopped'

// Then in props:
phase?: SessionPhase
```

If `DomainSession` already exports this type, import and reuse it rather than redeclaring.

---

**[Minor] `components/ambient-ui/src/components/chat-messages.tsx` — `isRunActive` returns `true` on bare `user` event before `run_started`**

Per the spec: _"The indicator SHALL appear only when the session is in Running phase AND the most recent lifecycle event in the message stream is `run_started`."_ The current implementation returns `true` as soon as it hits a `user` message, even if `run_started` has not yet been received. In the interval between user send and runner emitting `run_started`, the indicator will flash on, then off when messages continue. This may be intentional (optimistic UX) but it deviates from the spec — please confirm or update the spec scenario accordingly.

---

_— Amber_

PR #331, opened 2026-07-10
fix(dev): extract mTLS certs before gateway registration

squizzi:
## Summary\n\n- Reorders `setup-gateway-cli.sh` to extract mTLS certificates from the `openshell-server-tls` K8s secret **before** calling `openshell gateway add`, fixing the flaky \"mTLS certificates not found\" error in `test-gateway-e2e`.\n- The `openshell` CLI now expects client TLS material to exist at registration time rather than generating self-signed certs during `--local` add.\n- This also eliminates the previous \"overwrite self-signed with real\" step, since real cluster certs are written from the start.\n\n## Test plan\n\n- [ ] Run `make kind-setup-openshell-cli` against a running Kind cluster and verify gateway registration succeeds\n- [ ] Verify `openshell provider list --gateway tenant-a` connects successfully after setup\n- [ ] Confirm `test-gateway-e2e` CI job passes without flakes"

### DIFF
--- a/scripts/setup-gateway-cli.sh
+++ b/scripts/setup-gateway-cli.sh
@@ -79,18 +79,15 @@ for NS in "${NAMESPACES[@]}"; do
 
   GW_PORTS+=("$PORT")
 
-  # Register the gateway. Remove first if it already exists.
+  # Remove existing registration first (may also delete the cert dir).
   if openshell gateway list 2>/dev/null | grep -q "$GW_NAME"; then
     echo "  Removing existing gateway registration..."
     openshell gateway remove "$GW_NAME" 2>/dev/null || true
   fi
 
-  echo "  Registering gateway $GW_NAME -> https://localhost:$PORT..."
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
🤖 **Amber Review**

Clean fix. Moving cert extraction before gateway registration is the right call — the openshell CLI clearly needs client TLS material present at registration time, not after. Eliminating the self-signed generation + overwrite step simplifies the flow and closes the race window that caused the flaky `mTLS certificates not found` failure in `test-gateway-e2e`.

**Checklist:**
- ✅ Script-only change — no production Go/Python/frontend code modified
- ✅ No tokens or secrets logged
- ✅ Logic is correct: `kubectl get secret` → write certs → `gateway add` is the right order
- ✅ Test plan covers registration success and connectivity verification

Still a draft — undraft when you've validated against a running Kind cluster. Code is ready from my side.

— *Amber*

jsell-rh:
Labeling `amber/approved` — see review above.

— *Amber*

PR #333, opened 2026-07-10
fix(runner): add DNS pre-check to prevent MLflow blocking startup on unresolvable hosts

bsquizz:
## Summary

- Add a 5-second DNS resolution pre-check before calling `mlflow.set_tracking_uri()` / `mlflow.set_experiment()`. If the hostname from `MLFLOW_TRACKING_URI` doesn't resolve, MLflow initialization is skipped immediately instead of waiting for MLflow's internal HTTP timeout (120s × 7 retries = 4+ minutes blocking startup).
- Cache the DNS result so the second init path (`MLflowSessionTracer.initialize` after `activate_mlflow_autologging`) returns instantly.
- Use `executor.shutdown(wait=False, cancel_futures=True)` so the caller unblocks immediately after timeout — the `ThreadPoolExecutor` context manager's `shutdown(wait=True)` would have defeated the 5s goal.
- Remove `openshell:resolve:env:` token handling from `MLFLOW_TRACKING_URI` — it's a platform env var set at deployment time, not a credential. The runner should not inspect its format.
- Update the mlflow-tracing spec with the DNS pre-check requirement and simplify credential provider language.

## Test plan

- [x] 10 unit tests for DNS pre-check (resolvable, unresolvable, timeout with real sleep, caching, file/sqlite bypass, port defaults)
- [x] 1 integration test verifying unreachable host prevents autolog activation end-to-end
- [x] All 33 MLflow-related tests pass
- [x] Timeout test uses `time.sleep` side effect (not direct `FuturesTimeoutError` raise) to exercise the real timeout path

### DIFF
--- a/components/runners/ambient-runner/ambient_runner/observability_config.py
+++ b/components/runners/ambient-runner/ambient_runner/observability_config.py
@@ -11,7 +11,6 @@ from urllib.parse import urlparse
 logger = logging.getLogger(__name__)
 
 _DNS_CHECK_TIMEOUT = 5.0
-_OPENSHELL_RESOLVE_PREFIX = "openshell:resolve:env:"
 _NON_NETWORK_SCHEMES = frozenset({"file", "sqlite"})
 
 _mlflow_dns_cache: dict[str, bool] = {}
@@ -31,13 +30,10 @@ def check_mlflow_tracking_reachable(
 ) -> bool:
     """Fast DNS pre-check for the MLflow tracking URI hostname.
 
-    Returns True if the hostname resolves or if the URI is not subject to
-    DNS checks (openshell resolve tokens, non-network schemes like file://).
-    Returns cached result on subsequent calls for the same URI.
+    Returns True if the hostname resolves or if the URI uses a non-network
+    scheme (file://, sqlite://). Returns cached result on subsequent calls
+    for the same URI.
     """
-    if tracking_uri.startswith(_OPENSHELL_RESOLVE_PREFIX):
-        return True
-
     parsed = urlparse(tracking_uri)
 
     if parsed.scheme in _NON_NETWORK_SCHEMES or not parsed.hostname:
--- a/components/runners/ambient-runner/ambient_runner/observability_config.py
+++ b/components/runners/ambient-runner/ambient_runner/observability_config.py
@@ -2,7 +2,19 @@
 
 from __future__ import annotations
 
+import logging
 import os
+import socket
+from concurrent.futures import ThreadPoolExecutor, TimeoutError as FuturesTimeoutError
+from urllib.parse import urlparse
+
+logger = logging.getLogger(__name__)
+
+_DNS_CHECK_TIMEOUT = 5.0
+_OPENSHELL_RESOLVE_PREFIX = "openshell:resolve:env:"
+_NON_NETWORK_SCHEMES = frozenset({"file", "sqlite"})
+
+_mlflow_dns_cache: dict[str, bool] = {}
 
 
 def _truthy_env(name: str) -> bool:
@@ -12,6 +24,58 @@ def _truthy_env(name: str) -> bool:
 
 def _explicitly_false_env(name: str) -> bool:
     return os.getenv(name, "").strip().lower() in ("0", "false", "no", "off")
+
+
+def check_mlflow_tracking_reachable(
+    tracking_uri: str, timeout: float = _DNS_CHECK_TIMEOUT
+) -> bool:
+    """Fast DNS pre-check for the MLflow tracking URI hostname.
+
+    Returns True if the hostname resolves or if the URI is not subject to
+    DNS checks (openshell resolve tokens, non-network schemes like file://).
+    Returns cached result on subsequent calls for the same URI.
+    """
+    if tracking_uri.startswith(_OPENSHELL_RESOLVE_PREFIX):
+        return True
+
+    parsed = urlparse(tracking_uri)
+
+    if parsed.scheme in _NON_NETWORK_SCHEMES or not parsed.hostname:
+        return True
+
+    if tracking_uri in _mlflow_dns_cache:
+        return _mlflow_dns_cache[tracking_uri]
+
+    host = parsed.hostname
+    port = parsed.port or (443 if parsed.scheme == "https" else 80)
+
+    try:
+        with ThreadPoolExecutor(max_workers=1) as executor:
+            future = executor.submit(socket.getaddrinfo, host, port)
+            future.result(timeout=timeout)
+        _mlflow_dns_cache[tracking_uri] = True
+        return True
+    except FuturesTimeoutError:
+        logger.warning(
+            "MLflow: DNS resolution for %s timed out after %.0fs — skipping MLflow initialization",
+            host,
+            timeout,
+        )
+    except socket.gaierror as e:
+        logger.warning(
+            "MLflow: DNS resolution failed for %s (%s) — skipping MLflow initialization",
+            host,
+            e,
+        )
+    except Exception as e:
+        logger.warning(
+            "MLflow: DNS pre-check failed for %s (%s) — skipping MLflow initialization",
+            host,
+            e,
+        )
+
+    _mlflow_dns_cache[tracking_uri] = False
+    return False
 
 
 def observability_backend_names() -> frozenset[str]:
--- a/components/runners/ambient-runner/ambient_runner/observability_config.py
+++ b/components/runners/ambient-runner/ambient_runner/observability_config.py
@@ -45,10 +45,10 @@ def check_mlflow_tracking_reachable(
     host = parsed.hostname
     port = parsed.port or (443 if parsed.scheme == "https" else 80)
 
+    executor = ThreadPoolExecutor(max_workers=1)
     try:
-        with ThreadPoolExecutor(max_workers=1) as executor:
-            future = executor.submit(socket.getaddrinfo, host, port)
-            future.result(timeout=timeout)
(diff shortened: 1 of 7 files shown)


components/runners/ambient-runner/ambient_runner/observability_config.py
```diff
@@ -14,6 +26,58 @@ def _explicitly_false_env(name: str) -> bool:
...
+
+    try:
+        with ThreadPoolExecutor(max_workers=1) as executor:
+            future = executor.submit(socket.getaddrinfo, host, port)
+            future.result(timeout=timeout)
+        _mlflow_dns_cache[tracking_uri] = True
+        return True
+    except FuturesTimeoutError:
+        logger.warning(
+            "MLflow: DNS resolution for %s timed out after %.0fs — skipping MLflow initialization",
```
jsell-rh, line 56: **Critical — ThreadPoolExecutor blocks on `__exit__` after timeout, defeating the 5 s goal**

When `future.result(timeout=timeout)` raises `FuturesTimeoutError`, the `with ThreadPoolExecutor(...) as executor:` context manager calls `shutdown(wait=True)` on exit. That blocks until the DNS thread finishes — which is exactly the hung DNS resolution we're trying to avoid. In the real timeout scenario (resolver not returning), this means the code blocks at the `with` exit for the OS resolver timeout (typically 30–120 s), not 5 s.

The test suite doesn't catch this because `test_timeout_returns_false` patches `socket.getaddrinfo` to raise `FuturesTimeoutError` directly — the future completes immediately in the thread, so `shutdown(wait=True)` returns instantly. The actual problem only manifests when DNS is genuinely slow.

Fix — shut down without waiting:

```python
executor = ThreadPoolExecutor(max_workers=1)
try:
    future = executor.submit(socket.getaddrinfo, host, port)
    future.result(timeout=timeout)
    _mlflow_dns_cache[tracking_uri] = True
    return True
except FuturesTimeoutError:
    logger.warning(...)
except socket.gaierror as e:
    logger.warning(...)
except Exception as e:
    logger.warning(...)
finally:
    executor.shutdown(wait=False, cancel_futures=True)

_mlflow_dns_cache[tracking_uri] = False
return False
```

`cancel_futures=True` was added in Python 3.9; if you need 3.8 compat use `executor.shutdown(wait=False)` alone (the thread still leaks but at least the call site unblocks).

— *Amber*

components/runners/ambient-runner/ambient_runner/observability_config.py
```diff
@@ -14,6 +26,58 @@ def _explicitly_false_env(name: str) -> bool:
...
 
 
+def check_mlflow_tracking_reachable(
+    tracking_uri: str, timeout: float = _DNS_CHECK_TIMEOUT
+) -> bool:
+    """Fast DNS pre-check for the MLflow tracking URI hostname.
+
+    Returns True if the hostname resolves or if the URI is not subject to
+    DNS checks (openshell resolve tokens, non-network schemes like file://).
+    Returns cached result on subsequent calls for the same URI.
```
jsell-rh: **Minor — module-level mutable dict is not thread-safe**

`_mlflow_dns_cache: dict[str, bool] = {}` is mutated from a background thread (the executor) and read from the main thread without a lock. CPython's GIL makes individual dict reads/writes atomic in practice, but this is relying on an implementation detail. Consider using `threading.Lock` or switching to `functools.lru_cache` on a helper that takes `(host, port)` — both are cleaner and document the intent:

```python
import functools

@functools.lru_cache(maxsize=128)
def _resolve_hostname(host: str, port: int, timeout: float) -> bool:
    ...
```

Low priority given CPython's GIL protects the dict ops in practice, but worth noting for the next person who reads this on PyPy or free-threaded 3.13+.

— *Amber*

components/runners/ambient-runner/tests/test_mlflow_dns_precheck.py
```diff
@@ -0,0 +1,132 @@
...
+        ):
+            assert check_mlflow_tracking_reachable("https://mlflow.example.com") is True
+
+    def test_timeout_returns_false(self):
+        def slow_resolve(*args, **kwargs):
+            raise FuturesTimeoutError()
+
+        with patch(
+            "ambient_runner.observability_config.socket.getaddrinfo",
+            side_effect=slow_resolve,
```
jsell-rh: **Minor — `test_timeout_returns_false` doesn't actually test the timeout path**

The test patches `socket.getaddrinfo` to raise `FuturesTimeoutError` directly. But `FuturesTimeoutError` is raised by `future.result(timeout=...)` in the calling thread, not by `getaddrinfo` in the worker thread — so this test exercises the exception-handler branch, not a real timeout. More importantly, it won't catch the `shutdown(wait=True)` blocking issue described in the main comment above.

A more faithful timeout test would set `timeout=0.001` with a real `getaddrinfo` call (or a `time.sleep` side-effect) so the future genuinely outlives the deadline:

```python
def test_timeout_returns_false(self):
    import time

    with patch(
        "ambient_runner.observability_config.socket.getaddrinfo",
        side_effect=lambda *a, **kw: time.sleep(10),  # outlasts any real timeout
    ):
        result = check_mlflow_tracking_reachable("https://slow.example.com", timeout=0.01)
    assert result is False
```

This would also have caught the `shutdown(wait=True)` issue during test authoring.

— *Amber*

jsell-rh (changes requested):
🤖 **Amber Review — PR #333 `fix(runner): add DNS pre-check to prevent MLflow blocking startup`**

**Summary:** The goal is exactly right and the overall structure is solid — fast DNS pre-check, result caching, bypass paths for non-network URIs. One critical implementation bug needs fixing before merge.

**Blocker (Critical):**

`ThreadPoolExecutor` used as a context manager calls `shutdown(wait=True)` on `__exit__`, which blocks until the DNS thread finishes even after the timeout fires. In the real-world scenario this PR targets (unresolvable hostname), `shutdown(wait=True)` will hold for the OS resolver timeout (~30–120 s), negating the 5 s deadline. The test suite doesn't catch this because the timeout test patches `getaddrinfo` to raise immediately rather than actually sleeping. See inline comment on `observability_config.py:60` for the fix.

**What's good:**

- Bypass logic for `openshell:resolve:env:`, `file://`, and `sqlite://` URIs is correct.
- Caching keyed on the full URI string avoids repeat lookups on the two-call initialization path — that's the right design.
- Both `activate_mlflow_autologging()` and `MLflowSessionTracer.initialize()` wired consistently.
- 12 new tests cover the main paths. Test fixtures (`_bypass_dns_check`, `_clear_dns_cache`) are clean and don't pollute other test modules.
- Spec update is thorough and testable.
- PR description accurately documents the root cause and the two-call problem.

**Minor (non-blocking):**
- Module-level `_mlflow_dns_cache` dict is mutated across threads without a lock — safe under CPython's GIL but fragile on free-threaded runtimes. See inline note.

**Confidence:** High (95%) — the blocker is a known Python `ThreadPoolExecutor` gotcha; fix is straightforward.

Please fix the `shutdown(wait=True)` issue and update the timeout test to catch it. Happy to re-review on next push.

— *Amber*

jsell-rh:
Labeling `amber/change-requested` — one blocker to resolve before merge (see review above).

— *Amber*

jsell-rh (approved):
🤖 **Amber Re-Review — PR #333 `fix(runner): add DNS pre-check to prevent MLflow blocking startup`**

All blockers from the previous review are resolved. Approving.

---

**Blocker — `shutdown(wait=True)` blocking the caller** ✅ Fixed

`observability_config.py` now uses an explicit `finally` block with `executor.shutdown(wait=False, cancel_futures=True)` instead of the `ThreadPoolExecutor` context manager. This correctly releases the caller immediately after the timeout fires, regardless of how long the background DNS thread keeps running. The fix is exactly right.

**Blocker — Timeout test didn't catch the `shutdown` issue** ✅ Fixed

`test_timeout_returns_false` now uses `side_effect=lambda *a, **kw: time.sleep(10)` — a real sleep that would have exposed the original `wait=True` hang. The test would have blocked for 10 seconds under the old code; it now completes in ~10ms. This is the right way to write this test.

**Design simplification — openshell token bypass removed from tracking URI path** ✅ Clean

Removing `_OPENSHELL_RESOLVE_PREFIX` / `_is_openshell_token` from the `MLFLOW_TRACKING_URI` path and spec is the right call. `MLFLOW_TRACKING_URI` is a platform env var injected at deployment time — it should always be a real URI, not a lazy-resolve token. The DNS pre-check now handles the "bad URI" case uniformly. The `auth_mode` path in `MLflowSessionTracer` still uses `_is_openshell_token`, which is correct and consistent — the auth token may legitimately come from an openshell credential.

---

**Non-blocking note (carried over, still fine):**

`_mlflow_dns_cache` is mutated across threads without a lock. Under CPython's GIL this is safe; it's only fragile on free-threaded Python 3.13+. Worth a `threading.Lock()` when the codebase moves to 3.13 free-threaded, but not a blocker today.

**Non-blocking note (new):**

In `test_timeout_returns_false`, the background thread sleeps for 10 seconds after the test completes (the `time.sleep(10)` side effect runs in the executor thread, and `cancel_futures=True` only cancels queued-but-not-started futures, not running ones). The test itself returns in ~10ms, but the phantom thread lives for 10 more seconds. Not a correctness issue, just worth knowing if you see occasional resource warnings in CI.

---

**Confidence:** High (97%) — the critical correctness issue is fixed correctly and the test coverage is solid.

PR is still draft — presumably waiting for the manual test (`MLFLOW_TRACKING_URI=https://nonexistent.example.invalid:443` → verify output within ~5s). That's the right gate before marking ready for merge.

— *Amber*

PR #334, opened 2026-07-11
fix(control-plane): policy merge confirmation and sandbox restart resilience

bsquizz:
## Summary

- Capture the `UpdateConfigResponse` from the gateway after injecting the `_acp_internal` policy merge operation and log the returned `policy_version` and `policy_hash`. Provides definitive confirmation that the gateway applied the merge, replacing the previous fire-and-forget pattern.
- Fix sandbox restart crash caused by supervisor session supersession race condition during ndots pod deletion. Make `ExecSandboxStreaming` blocking so stream errors propagate, switch from force-kill to graceful pod deletion, and add exec retry logic for transient `Unavailable` errors.

## Context

PR #318 introduces scoped sandbox policies with `mergePlatformRules()` at `CreateSandbox` time and a post-READY `UpdateConfig` merge, but neither path confirms the gateway actually applied the policy. This PR adds that confirmation signal.

The ndots DNS fix (`verifyAndFixDNSConfig`) was force-deleting pods with `GracePeriodSeconds: 0`, triggering a supervisor reconnection storm that superseded the active session ~100ms after exec started. The runner appeared to crash immediately on startup, but the root cause was the exec relay being killed by session supersession.

## Changes

**Policy merge confirmation:**
- Log `policy_version` and `policy_hash` from `UpdateConfig` response
- Add spec scenario documenting the confirmation contract

**Sandbox restart resilience:**
- `ExecSandboxStreaming`: remove goroutine wrapper, make blocking with error/exit code propagation
- `verifyAndFixDNSConfig`: graceful pod deletion (default grace period) instead of force-kill
- `execAfterReady`: `sawNonReady` flag to wait for sandbox to leave READY before accepting it as READY again after graceful pod termination
- Exec retry loop (3 attempts, 3s delay) for `codes.Unavailable` errors from transient supervisor reconnects

## Test plan

- [ ] Verify control plane logs include `policy_version` and `policy_hash` after `_acp_internal` merge
- [ ] Verify sandbox with ndots:5 is gracefully deleted and recreated without runner crash
- [ ] Verify exec retries on transient `Unavailable` errors during supervisor reconnect
- [ ] Verify no change in sandbox provisioning behavior for pods that already have ndots:1

### DIFF
--- a/components/ambient-control-plane/internal/openshell/gateway_client.go
+++ b/components/ambient-control-plane/internal/openshell/gateway_client.go
@@ -301,38 +301,37 @@ func (g *GatewayClient) ExecSandboxStreaming(ctx context.Context, namespace stri
 		return err
 	}
 
-	go func() {
-		for {
-			event, err := stream.Recv()
-			if err == io.EOF {
-				g.logger.Debug().Str("sandbox_id", req.SandboxId).Msg("exec stream ended")
-				return
-			}
-			if err != nil {
-				g.logger.Warn().Err(err).Str("sandbox_id", req.SandboxId).Msg("exec stream error")
-				return
-			}
-			switch p := event.Payload.(type) {
-			case *pb.ExecSandboxEvent_Stdout:
-				chunk := p.Stdout.Data
-				if len(chunk) > maxLogChunkSize {
-					chunk = chunk[:maxLogChunkSize]
-				}
-				g.logger.Debug().Str("sandbox_id", req.SandboxId).Str("stdout", string(chunk)).Msg("exec stdout")
-			case *pb.ExecSandboxEvent_Stderr:
-				chunk := p.Stderr.Data
-				if len(chunk) > maxLogChunkSize {
-					chunk = chunk[:maxLogChunkSize]
-				}
-				g.logger.Debug().Str("sandbox_id", req.SandboxId).Str("stderr", string(chunk)).Msg("exec stderr")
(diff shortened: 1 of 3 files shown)


jsell-rh (approved):
🤖 **Amber Analysis**

Clean observability improvement. Replacing the discarded `_` return value with `resp` and logging `policy_version`/`policy_hash` via nil-safe proto getters is the right pattern — no behavioral change, no security concerns, and the spec scenario accurately documents the confirmation contract.

**Confidence:** High [95%] — minimal diff, correct proto getter usage, spec stays in sync.

<details>
<summary>Full checklist</summary>

- ✅ No `panic()` in production Go code
- ✅ No tokens/secrets in logs (`policy_hash` is a safe content digest)
- ✅ Proto getters (`GetVersion()`, `GetPolicyHash()`) are nil-safe — zero values on nil resp, not a crash
- ✅ Error handling path unchanged and correct
- ✅ Spec GIVEN/WHEN/THEN scenario is accurate and adds the right constraint (synchronous RPC is authoritative, no polling)
- ✅ No API schema changes, no OwnerReference concerns, no user-auth surface touched

One non-blocking observation: if you want to catch the case where the gateway returns a successful RPC but with empty/zero fields (possible under certain gateway bugs), you could add a `warn`-level log when `resp.GetVersion() == 0`. Not required for this PR's stated scope.

</details>

— *Amber*

PR #336, opened 2026-07-11
fix(control-plane): add 15s grace period for sandbox error during creation

bsquizz:
## Summary

- Sandbox provisioning on the openshell gateway can hit transient `SANDBOX_PHASE_ERROR` (e.g. brief error after ndots patch + pod recreation) that resolves on its own. Previously the control plane failed the session immediately on first observation.
- Both the provisioning poller (`execAfterReady`, polls every 2s) and the periodic status syncer (`PodStatusSyncer`, polls every 15s) now wait 15 seconds of sustained error before transitioning to `Failed`. If the sandbox recovers within the grace period, the timer resets.
- Grace period only applies to sessions in `Creating` phase — `Running` sessions that enter error state still fail immediately.

## Test plan

- [ ] Deploy to kind cluster (`imagePullPolicy: Never` with local build)
- [ ] Create a session that triggers ndots pod recreation — verify the sandbox recovers through the grace period instead of failing
- [ ] Verify logs show `"sandbox entered error phase, starting grace period"` and `"sandbox recovered from error phase"` when transient error resolves
- [ ] Verify sessions with sustained errors (>15s) still transition to `Failed`
- [ ] `go build ./...`, `go vet ./...`, `go test ./internal/reconciler/` all pass


jsell-rh (approved):
🤖 **Amber Analysis — PR #336**

Solid fix. The grace period approach is correct, both code paths use the same constant, spec updates are complete, and CI is green (CodeQL ✅, Konflux ✅, build matrix ✅). Approving with two advisory notes that are not blockers.

---

**What I checked:**

- `execAfterReady` — local `firstErrorSeen time.Time` is goroutine-safe (this function runs in its own goroutine per session), resets correctly on recovery, only applies during the `Creating`-phase polling loop ✅
- `PodStatusSyncer.errorFirstSeen` — map access is safe because `syncOnce()` is driven by a single ticker goroutine with no internal concurrency ✅
- Grace period is scoped to `Creating` phase only — `Running` sessions still fail immediately ✅
- Single `sandboxErrorGracePeriod` constant shared across both code paths — consistent ✅
- Spec updates in `agent-sandbox-config.spec.md`, `control-plane.spec.md`, `openshell-sandbox-provisioning.spec.md` all accurate ✅
- No `panic()`, no tokens in logs ✅

---

**Advisory (non-blocking):**

**1. `errorFirstSeen` leak on sandbox `NotFound` during grace period**

In `syncSandboxStatus`, the early-return on `codes.NotFound` doesn't clean up the map:

```go
if st, ok := status.FromError(err); ok && st.Code() == codes.NotFound {
    if session.Phase == PhaseRunning || session.Phase == PhaseCreating {
        s.updateSessionPhase(ctx, sdk, session, PhaseFailed, nil)
    }
    return  // ← errorFirstSeen[session.ID] leaks if we were mid-grace-period
}
```

The leak is bounded (one `time.Time` per affected session, never read again once terminal) and practically harmless. Worth a follow-up `delete(s.errorFirstSeen, session.ID)` before the `return` in a future cleanup pass.

**2. No unit tests for `syncSandboxStatus` grace period behavior**

The test helper is correctly updated, but there are no new cases exercising the "start timer → stay in error → exceed grace → fail" and "start timer → recover → reset" paths for `PodStatusSyncer`. The `execAfterReady` path is similarly uncovered (tight polling loop, harder to unit test). Given the integration test plan in the PR description covers this, not a blocker — but consider adding table-driven tests with a fake gateway when the test infrastructure is available.

---

**Confidence:** High (95%) — Logic is sound, CI builds pass, spec is in sync. The edge case above is real but inconsequential in production given session ID uniqueness and terminal-phase filtering.

— *Amber*

PR #346, opened 2026-07-13
Add db ca cert path

bsquizz:


jsell-rh (changes requested):
🤖 **Amber Review — PR #346: Add db ca cert path**

**Summary:** Adds `--db-rootcert=/secrets/db/db.ca_cert` to both the migration init container and the API server serve container in `template-services.yaml`. The intent is correct — TLS CA cert support for PostgreSQL — but there are two issues that need addressing before this is safe to merge.

---

**Confidence:** Medium (85%) — the flag itself is likely valid in rh-trex-ai, but the deployment impact on existing environments is unclear without more context.

---

### Issues

**1. [Critical] `db.ca_cert` key assumed present in all DB secrets — no fallback**

The flag `--db-rootcert=/secrets/db/db.ca_cert` is hardcoded unconditionally into both containers. The `/secrets/db` volume is mounted from the DB secret (`${DB_SECRET_NAME}`). If `db.ca_cert` is absent from that secret, the file at `/secrets/db/db.ca_cert` will not exist, and depending on how `rh-trex-ai` reads the flag, the API server may fail to start or connect.

No existing overlay (`kind`, `local-dev`, `production`, `mpp-openshift`, `openshift-dev`, `hcmais`) adds a `db.ca_cert` key to their DB secret. This change will silently break any environment that doesn't manually add `db.ca_cert` to its secret.

**Options:**
- Make the flag optional/conditional (e.g., only set when `DB_ROOTCERT` template param is non-empty) — if rh-trex-ai supports an empty string as "no cert", document that.
- Add a `DB_ROOTCERT` template parameter (with default empty) and use it so overlay patches can override.
- Update all affected overlays (and the DB secret template) to include a `db.ca_cert` key — and document what value to use for non-SSL environments (an empty file? a dummy value?).
- Or confirm that the rh-trex-ai flag gracefully handles a missing file — if so, document this explicitly in the PR.

**2. [Minor] No PR description**

The PR body is empty. There's no context about:
- Which environment this targets (hcmais? all environments?)
- Whether the DB being connected to actually requires TLS with CA verification
- What operators need to put in the `db.ca_cert` secret key
- Whether existing deployments need migration steps

For a change that touches the deployment template used by all environments, this context is important for reviewers and for the git history.

---

### Checklist (template-services.yaml)
- [x] No tokens/secrets in logs
- [x] SecurityContext on both containers: `allowPrivilegeEscalation: false`, `drop: ALL` ✅ (existing, not changed)
- [x] Volume mount for `/secrets/db` already exists — no new mount needed ✅
- [ ] Conditional/optional handling for environments where CA cert is not applicable ❌
- [ ] PR description with deployment context ❌

---

Please clarify the intended deployment scope and how to handle environments without a CA cert in the secret.

*— Amber*

PR #353, opened 2026-07-14
feat(runner): add connect retry, file logging, and exec resiliency

bsquizz:
## Summary

- Add connect retry with exponential backoff to `SessionWorker._run()` — transient CLI startup failures (sandbox file locks, binary readiness races, OpenShell supervisor reconnects) no longer cause permanent worker death. Configurable via `CLAUDE_CONNECT_MAX_RETRIES` / `CLAUDE_CONNECT_RETRY_DELAY` env vars.
- Add connect readiness signal (`asyncio.Event` + `wait_for_connect()`) so `SessionManager` detects startup failure immediately instead of hanging on `worker.query()`.
- Tee logs to `/var/log/runner/runner.log` via `RotatingFileHandler` (50 MB, 3 backups) for post-mortem debugging when container stdout is rotated/truncated. Graceful degradation if directory is not writable.
- Increase control plane exec retries from 3→9 and retry on **all** errors (not just gRPC `Unavailable`) for more robust entrypoint execution.
- Increase CP token fetch attempts from 3→30 with fixed 2s delay to handle slow-starting control plane pods.
- Fix gateway CLI cert extraction ordering (extract after `gateway add`, not before).
- Rename test agent fixture (`test-agent-with-no-providers` → `test-agent-no-providers`).

## Components Changed

- `components/runners/ambient-runner/` — Python runner (connect retry, file logging, policy)
- `components/ambient-control-plane/` — Go reconciler (exec retry widening)
- `specs/platform/runner.spec.md` — Spec updates for new requirements
- `scripts/setup-gateway-cli.sh` — Gateway cert ordering fix
- `tests/e2e/gateway-e2e-test.sh` — Agent rename in e2e tests

## Test plan

- [ ] Runner starts successfully with default retry settings (happy path)
- [ ] Runner recovers from transient `connect()` failure on attempt 1-2
- [ ] Runner reports immediate error (not hung stream) when all connect retries exhausted
- [ ] Logs appear in both stdout (`kubectl logs`) and `/var/log/runner/runner.log`
- [ ] Runner starts normally when `/var/log/runner` is not writable (graceful degradation)
- [ ] Control plane exec retries recover from transient pod startup delays
- [ ] E2E gateway test passes with renamed agent

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -838,8 +838,8 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 				}
 			}
 
-			const maxExecRetries = 3
-			const execRetryDelay = 3 * time.Second
+			const maxExecRetries = 9
+			const execRetryDelay = 2 * time.Second
 
 			for attempt := 0; attempt <= maxExecRetries; attempt++ {
 				if attempt > 0 {
@@ -847,7 +847,8 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 						Str("sandbox", sbxName).
 						Str("session_id", sessionID).
 						Int("attempt", attempt+1).
-						Msg("retrying entrypoint exec after relay error")
+						Int("max_attempts", maxExecRetries+1).
+						Msg("retrying entrypoint exec")
 					time.Sleep(execRetryDelay)
 				}
 
@@ -860,18 +861,22 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 					break
 				}
 
-				if st, ok := status.FromError(err); ok && st.Code() == codes.Unavailable && attempt < maxExecRetries {
-					r.logger.Warn().Err(err).
+				r.logger.Warn().Err(err).
+					Str("sandbox", sbxName).
+					Str("session_id", sessionID).
+					Int("attempt", attempt+1).
+					Int("max_attempts", maxExecRetries+1).
+					Msg("entrypoint exec failed")
+
+				if attempt == maxExecRetries {
+					r.logger.Error().Err(err).
 						Str("sandbox", sbxName).
 						Str("session_id", sessionID).
-						Int("attempt", attempt+1).
-						Msg("exec relay died (likely supervisor reconnect), will retry")
-					continue
+						Int("attempts", maxExecRetries+1).
+						Msg("failed to start runner exec after all retries")
+					failSession(fmt.Sprintf("failed to start runner exec after %d attempts: %v", maxExecRetries+1, err))
+					return
 				}
-
-				r.logger.Error().Err(err).Str("sandbox", sbxName).Str("session_id", sessionID).Msg("failed to start runner exec")
-				failSession(fmt.Sprintf("failed to start runner exec: %v", err))
-				return
 			}
 
 			// FIXME(#223): Mark session PhaseCompleted and set completion_time here once
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -838,7 +838,7 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
 				}
 			}
 
-			const maxExecRetries = 9
+			const maxExecRetries = 29
 			const execRetryDelay = 2 * time.Second
 
 			for attempt := 0; attempt <= maxExecRetries; attempt++ {
--- a/components/runners/ambient-runner/Dockerfile
+++ b/components/runners/ambient-runner/Dockerfile
@@ -83,8 +83,6 @@ ENV AGUI_PORT=8001
 RUN echo "umask 0022" >> /etc/profile && \
     echo "umask 0022" >> /root/.bashrc
 
-# Persistent log directory for RotatingFileHandler tee
-RUN mkdir -p /var/log/runner && chown 1001:0 /var/log/runner && chmod 775 /var/log/runner
 
 # OpenShift compatibility
 RUN chmod -R g=u /app && chmod -R g=u /usr/local && chmod g=u /etc/passwd
--- a/components/runners/ambient-runner/Dockerfile
+++ b/components/runners/ambient-runner/Dockerfile
@@ -83,6 +83,9 @@ ENV AGUI_PORT=8001
 RUN echo "umask 0022" >> /etc/profile && \
     echo "umask 0022" >> /root/.bashrc
 
+# Persistent log directory for RotatingFileHandler tee
+RUN mkdir -p /var/log/runner && chown 1001:0 /var/log/runner && chmod 775 /var/log/runner
+
 # OpenShift compatibility
 RUN chmod -R g=u /app && chmod -R g=u /usr/local && chmod g=u /etc/passwd
 
--- a/components/runners/ambient-runner/ambient_runner/_grpc_client.py
+++ b/components/runners/ambient-runner/ambient_runner/_grpc_client.py
@@ -31,7 +31,7 @@ _SERVICE_CA_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
 _SA_TOKEN_FILE = Path("/var/run/secrets/kubernetes.io/serviceaccount/token")
 
 
-_CP_TOKEN_FETCH_ATTEMPTS = 3
+_CP_TOKEN_FETCH_ATTEMPTS = 30
 _CP_TOKEN_FETCH_TIMEOUT = 10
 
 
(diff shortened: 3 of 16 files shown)


components/runners/ambient-runner/ambient_runner/_grpc_client.py
```diff
@@ -85,15 +85,13 @@ def _fetch_token_from_cp(
     last_err: Exception = RuntimeError("no attempts made")
     for attempt in range(_CP_TOKEN_FETCH_ATTEMPTS):
         if attempt > 0:
-            backoff = 2 ** (attempt - 1)
             logger.warning(
```
jsell-rh, line 88: The token fetch retry changed from exponential backoff (`2^(attempt-1)`) to a fixed 2s delay. The old exponential approach makes sense for a resource under load — fixed 2s over 30 attempts means up to 60s of polling at maximum frequency from the first retry. This isn't wrong, but it's a deliberate trade-off change worth calling out: the old approach was gentler on a slow-starting control plane, the new approach is more aggressive.

If the intent is "wait up to ~60s total for the CP to start", consider keeping exponential backoff but capping total wait time. With 30×2s you'll hit 60s, but all retries are equally spaced. If that's intentional, a comment explaining the reasoning (e.g. "fixed delay to ensure we distribute retries evenly across a 60s startup window") would prevent future readers from reverting it thinking it's a regression.

Minor — not a blocker.

*— Amber*

components/runners/ambient-runner/ambient_runner/bridges/claude/session.py
```diff
@@ -433,6 +494,18 @@ async def get_or_create(
             thread_id, options, api_key, on_session_id=self._on_session_id
```
jsell-rh, line 494: Race condition risk: if `worker.start()` returns and the worker's `_run()` task has already set `_connect_ready` (e.g. immediate failure), `wait_for_connect()` will correctly raise. But if the task hasn't been scheduled yet (event loop hasn't yielded), `_connect_ready` won't be set and `wait_for_connect()` will correctly block. This looks safe because `asyncio.Event.wait()` will block until the event is set regardless of ordering. Confirmed — no issue here, just documenting for reviewers.

What's missing: the worker is added to `self._workers` **after** `wait_for_connect()` succeeds. But if `wait_for_connect()` raises, `worker.stop()` is called — good. However, `self._locks[thread_id]` is never set in the failure path. If there's any codepath that checks `self._locks` without checking `self._workers`, that could panic. Worth a quick grep to confirm the failure path is clean.

*— Amber*

components/runners/ambient-runner/Dockerfile
```diff
@@ -83,6 +83,9 @@ ENV AGUI_PORT=8001
 RUN echo "umask 0022" >> /etc/profile && \
     echo "umask 0022" >> /root/.bashrc
 
+# Persistent log directory for RotatingFileHandler tee
```
jsell-rh: The container SecurityContext in the deployment manifests likely enforces `readOnlyRootFilesystem: true`. If so, `/var/log/runner` must be backed by an `emptyDir` volume mount — a directory created in the image layer won't be writable at runtime with a read-only root filesystem.

The graceful degradation in `app.py` will catch this and log a warning, so it won't crash. But it means file logging will silently be unavailable in prod unless the volume is wired up. Worth checking whether the manifests need a corresponding `emptyDir` volume + mount, or whether the graceful degradation is the intended behavior for the current security posture.

*— Amber*

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -838,16 +838,17 @@ func (r *SimpleKubeReconciler) execAfterReady(namespace, sbxName, sessionID stri
...
 
-			const maxExecRetries = 3
-			const execRetryDelay = 3 * time.Second
+			const maxExecRetries = 29
+			const execRetryDelay = 2 * time.Second
 
 			for attempt := 0; attempt <= maxExecRetries; attempt++ {
 				if attempt > 0 {
 					r.logger.Info().
 						Str("sandbox", sbxName).
```
jsell-rh, line 847: `maxExecRetries = 29` (30 total attempts at 2s each = up to 60s of retry time). The old code was 3 retries only on `codes.Unavailable`; this PR retries on **all** errors. That's a meaningful behavioral change — a non-transient error (e.g., permission denied, sandbox not found) will now retry 29 times before failing, adding up to 58 extra seconds of delay on a guaranteed-to-fail execution.

Consider checking whether the error is retryable before retrying. At minimum, document the intent (e.g., "We retry on all errors because we cannot reliably distinguish transient relay errors from permanent failures in this environment").

*— Amber*

jsell-rh (commented):
🤖 **Amber Analysis — PR #353: Runner Connect Retry, File Logging, Exec Resiliency**

This is solid resiliency work. The connect-retry + readiness-event pattern is well-designed and directly addresses real failure modes (sandbox file locks, binary readiness races). The file-logging tee with graceful degradation is correct. The gateway cert ordering fix is clearly right.

**Four advisory items** (no blockers — this is a DRAFT, noting now so they land before merge):

1. **Dockerfile + readOnlyRootFilesystem** (line 86, Dockerfile) — `/var/log/runner` created in the image layer won't survive a `readOnlyRootFilesystem: true` SecurityContext at runtime. Graceful degradation will kick in silently. Check if manifests need an `emptyDir` mount.

2. **`maxExecRetries = 29` retries on all errors** (Go reconciler) — old code retried 3× only on `codes.Unavailable`. New code retries 30× on every error, adding up to 58s of delay for guaranteed-failure scenarios (permission denied, sandbox not found). Intent should be documented or error-type filtering restored.

3. **Fixed vs exponential backoff in `_grpc_client.py`** — `_CP_TOKEN_FETCH_ATTEMPTS = 30` with fixed 2s is a deliberate trade-off from the prior exponential. Not wrong, but worth a comment explaining why fixed delay is preferred here.

4. **`self._locks[thread_id]` in connect-failure path** — `wait_for_connect()` failure correctly calls `worker.stop()` but doesn't set `self._locks`. If any codepath reads `self._locks` without checking `self._workers` first, it would KeyError. Quick grep recommended before merge.

**ACP Constitution:** No `panic()` in Go code ✓, no `any` types ✓, no token logging ✓, errors wrapped with context ✓.

Leaving as COMMENT rather than REQUEST_CHANGES — these are advisory on a DRAFT. Happy to re-review when ready.

*— Amber*

jsell-rh:
Adding `amber/change-requested` — four advisory items noted in the review above. No blockers, all addressable before this comes out of DRAFT. Happy to re-review.

*— Amber*

components/runners/ambient-runner/ambient_runner/_grpc_client.py
bsquizz: Using a fixed short delay to ensure that the session comes up as quickly as possible.

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
bsquizz: Using a fixed short delay to ensure that the session comes up as quickly as possible during transient errors.

components/runners/ambient-runner/ambient_runner/bridges/claude/session.py
bsquizz: Confirmed the failure path is clean — `get_lock()` lazily creates the lock if the key is missing (line 520-521), and `destroy()` uses `.pop(thread_id, None)` which is a no-op if absent. No codepath accesses `self._locks[thread_id]` directly without a guard.

components/runners/ambient-runner/Dockerfile
bsquizz: Good catch. Moved the log directory to `/sandbox/.runner/logs` — this lives under the sandbox's writable filesystem so it works regardless of `readOnlyRootFilesystem`. Removed the `mkdir` from both Dockerfiles since `app.py` already creates the directory at startup with `os.makedirs(..., exist_ok=True)`. Also updated `policy.yaml` to allow the new path.

jsell-rh (commented):
## 🤖 Amber Re-Review — Change Requests Addressed

All four items from my previous review have been satisfactorily resolved:

1. **Token fetch fixed delay** — author explained intent (fixed delay for predictable 60s window). Accepted.
2. **Lock failure path race** — author confirmed `get_lock()` lazily creates and `destroy()` uses `.pop(..., None)`. No issue.
3. **readOnlyRootFilesystem + `/var/log/runner`** — fixed in commit `0ab9900c` by moving to `/sandbox/.runner/logs`. Correct approach.
4. **Retry on all errors** — behavioral change is intentional (can't reliably distinguish transient vs permanent relay errors in sandbox env). Documented in design decisions table.

**One minor spec/code discrepancy** (not a blocker):

`specs/platform/runner.spec.md` under the "Persistent File Logging" section still references `/var/log/runner/runner.log` as the log path in the configuration table, but the implementation now uses `/sandbox/.runner/logs/runner.log`. Worth a one-line update before this merges from DRAFT.

Updating label to `amber/approved`. Happy to re-review any further changes.

*— Amber*

jsell-rh (commented):
## 🤖 Amber — 2 commits since last pass (`0ab9900c`, 2026-07-14T01:20Z)

`amber/approved` stands. Both commits are clean.

| Commit | Summary | Finding |
|--------|---------|---------|
| `c5e2e899` `spec: align runner and sandbox provisioning specs` | Fixes `/var/log/runner` → `/sandbox/.runner/logs` in spec, adds ExecSandbox retry requirement (30 attempts, 2s fixed), updates filesystem policy table | ✅ Directly resolves my outstanding nit from the previous pass. ExecSandbox retry spec is accurate and matches the implementation. |
| `c6c9f78` `fix(ci): use PR-built runner image in e2e tests` | `RUNNER_QUAY_TAG` Makefile variable, `test-local-dev.yml` resolves PR-built tag when `runner_openshell` was built in the PR | ✅ Correct root cause fix. The `|| 'latest'` fallback handles the case where the `image-override` step doesn't run (guarded by `IMAGE_TAG != 'latest'`). Clean. |

**One minor observation (non-blocking):**

`runner.spec.md` filesystem policy table now lists both `/sandbox` (read-write) and `/sandbox/.runner/logs` (read-write). Since `/sandbox/.runner/logs` is a subdirectory of `/sandbox`, listing it separately is redundant but harmless. Also note `/app/.claude` was removed from read-write — Claude Code writes settings there; verify this is intentional (e.g., `/sandbox/.claude` is used instead in OpenShell layout) before merge.

No blockers. Still a DRAFT with an unchecked test plan item (`spec/runner.spec.md` log path reference). That's now fixed — the spec item should be checkable.

*— Amber*

PR #355, opened 2026-07-14
spec: add multi-cluster scheduling model (Cluster Kind, PlacementStrategy)

markturansky:
## Summary

- Introduces `Cluster` as a first-class API Kind with roles (`gateway`, `workload`, `hybrid`) for multi-cluster scheduling
- Adds `PlacementStrategy` interface (default: round-robin) that determines which cluster a session runs on at ignite time
- Adds `cluster_id` FK to Session, Gateway, and Application; adds `gateway_cluster_id` to Session for cross-cluster gateway→workload routing
- Adds `ClusterReconciler`, `ClusterClientPool`, and `ClusterHealthSyncer` control plane components
- Updates gateway discovery for cross-cluster endpoints via `ambient-code.io/gateway-external-url` annotation
- Updates identity boundaries with remote cluster kubeconfig isolation and least-privilege scoping

### Files changed

 < /dev/null |  File | Changes |
|------|---------|
| `specs/platform/data-model.spec.md` | Cluster Kind, ER diagram, FKs, API endpoints, RBAC, PlacementStrategy interface, design decisions |
| `specs/platform/control-plane.spec.md` | ClusterReconciler, PlacementStrategy, ClusterHealthSyncer, ClusterClientPool components |
| `specs/platform/gateway-provisioning.spec.md` | Cross-cluster gateway exposure, cluster field in YAML/schema |
| `specs/platform/openshell-sandbox-provisioning.spec.md` | Cross-cluster gateway discovery, connection cache keying |
| `specs/platform/index.spec.md` | Multi-cluster scheduling sub-spec entry |
| `specs/security/identity-boundaries.spec.md` | Remote cluster kubeconfig identity, cross-cluster CP scenario, credential isolation |

## Test plan

- [ ] Specs reviewed for internal consistency (cluster_id FKs, role constraints, placement logic)
- [ ] Backward compatibility verified: null cluster_id = local cluster, zero migration for single-cluster deployments
- [ ] Cross-references between specs are valid (data-model ↔ control-plane ↔ gateway-provisioning ↔ identity-boundaries)


jsell-rh (changes requested):
## 🤖 Amber Review — Multi-Cluster Scheduling Spec

**Overall:** A well-structured spec with careful attention to backward compatibility (null cluster_id = local), security (kubeconfig encrypted at rest, write-only in API, least-privilege SA scoping), and operational visibility (ClusterHealthSyncer). The RBAC model is coherent and the PlacementStrategy interface is clean.

Two architectural gaps need resolution before this merges:

---

**[Major] Namespace provisioning on workload clusters**

`data-model.spec.md` describes how sessions are placed on workload clusters via PlacementStrategy, but the spec doesn't state who creates the target namespace on the remote workload cluster. Today, `ProjectReconciler` creates namespaces on the local cluster via `ensureNamespace()`. For remote workload clusters, does:
- The `KubeReconciler` call `ensureNamespace()` on the remote cluster's client before provisioning session pods?
- The `ClusterReconciler` pre-create namespaces for all projects when a new cluster registers?
- Something else?

This needs to be explicit in `control-plane.spec.md` under the `KubeReconciler` section. The failure mode if the namespace doesn't exist on the remote cluster is a silent provisioning failure — worth calling out.

**[Major] In-flight session handling during cluster deregistration**

`control-plane.spec.md` says "when a Cluster is deleted, the reconciler closes and evicts the client" — but what happens to sessions currently running on that cluster? They'll lose their gRPC connection to the gateway (if cross-cluster) or have their pod become unreachable. The spec should describe:
- Whether the deregistration is soft (drains in-flight sessions) or hard (immediate eviction)
- Whether the API server marks affected sessions as `Failed` with a `ClusterDeregistered` condition
- Whether there's a grace period

---

**[Minor] Heartbeat endpoint RBAC**

`POST /api/ambient/v1/clusters/{id}/heartbeat` (manual health check trigger) — the RBAC matrix shows `cluster:admin` and `cluster:viewer` columns but doesn't specify which roles can trigger a manual heartbeat. Triggering a health check should probably be `cluster:admin` only (it creates load on the remote cluster's API server). Worth a footnote in the endpoint table.

**[Minor] kubeconfig rotation lifecycle**

No mention of what happens when a cluster's stored kubeconfig expires (SA token TTL, client cert expiry). The `ClusterHealthSyncer` will detect `NotReady` and set an appropriate `status_message`, but there's no described remediation path beyond manually `PATCH`ing the Cluster resource with a new kubeconfig. A single sentence in the design decisions table would set expectations.

---

**What's solid:**
- Backward compat via null cluster_id sentinel is exactly right
- kubeconfig write-only + encrypted at rest follows the Credential pattern correctly
- `session-affinity: colocated` annotation for gateway co-location is a clean UX
- Connection cache keyed by `(cluster_id, namespace)` correctly prevents cross-cluster key collisions
- The RBAC isolation guarantees (point 7 and 8 in identity boundaries) are strong

*— Amber*

jsell-rh (commented):
## 🤖 Amber Follow-up — commit `6bce271`

The `credential_id` FK refactor is the right call — reusing the existing Credential infrastructure (encrypted write-only storage, rotation via PATCH, RBAC via `credential:token-reader`) is strictly better than a parallel kubeconfig field. That design decision also resolves the kubeconfig rotation question from my prior review, since rotation now follows the standard `PATCH /credentials/{id}` path.

My two Major blockers from the prior review are still open and need spec language before this can merge:

---

**[Major — still open] Namespace provisioning on workload clusters**

`data-model.spec.md` describes session placement on remote workload clusters, but neither `control-plane.spec.md` nor `data-model.spec.md` says who creates the target namespace on the remote cluster. Today `ProjectReconciler.ensureNamespace()` handles this on the local cluster. For remote clusters the spec needs to answer explicitly: does `KubeReconciler` call `ensureNamespace()` on the remote `KubeClient` before provisioning pods, or does `ClusterReconciler` pre-create namespaces for all projects when a cluster registers? The silent failure path (missing namespace → pod creation fails with no user-visible condition) is the production footgun here.

**[Major — still open] In-flight session handling during cluster deregistration**

`control-plane.spec.md` still says "when a Cluster is deleted, the reconciler closes and evicts the client" without specifying what happens to sessions currently running on that cluster. The spec needs: drain vs. hard-evict semantics, whether the API server marks affected sessions `Failed` with a `ClusterDeregistered` condition, and whether there's a grace period. Without this, operators have no defined behavior to reason about during planned cluster decommissioning.

---

**[Minor — now partially resolved]** Kubeconfig expiry handling: the Credential PATCH path covers rotation, but there's still no mention of what the `ClusterHealthSyncer` does when it detects auth failure vs. network failure — a sentence distinguishing "kubeconfig expired" (`NotReady`, status_message="auth error: ...") from "unreachable" (`NotReady`, status_message="connection refused") would help operators triage.

Two quick wins on the heartbeat RBAC footnote and that ClusterHealthSyncer failure-mode distinction; the namespace provisioning and deregistration semantics are the gates.

*— Amber*

jsell-rh (changes requested):
🤖 **Amber Review — Multi-Cluster Scheduling Spec**

Good foundational spec overall. The credential_id FK approach for Cluster kubeconfigs (second commit) is the right call — it reuses the existing encrypted write-only Credential infrastructure cleanly. The RBAC table, backward-compat sentinel (`_local`), and identity-boundaries updates are solid.

Three issues need addressing before implementation begins. Two are material design gaps; one is a spec ambiguity that could cause a confusing implementation.

---

### [Major] Credential rotation → ClusterClientPool refresh is unspecified

The spec says the ClusterReconciler rebuilds the KubeClient when a Cluster resource changes. But there is no watch path for `Credential` changes. When a kubeconfig Credential is rotated via `PATCH /credentials/{id}`, the ClusterClientPool continues serving the old KubeClient indefinitely.

This is a security and reliability gap: the whole point of rotation is for the old credential to become invalid. The spec needs to define the refresh chain:

- Does the CP watch Credential events from the API server's gRPC stream?
- Or does the ClusterReconciler periodically re-fetch the Credential token and rebuild the KubeClient on a configurable interval?
- Or does rotating the Credential also require a PATCH to the Cluster to trigger reconciliation?

Any of these is acceptable — but the spec must pick one and describe it. The implementation will hit this immediately.

---

### [Major] No recovery path when the placed cluster goes NotReady between placement and provisioning

The spec says PlacementStrategy runs at session creation in `ignite_handler.go` and writes `cluster_id` before the session enters `Pending`. The control plane then picks it up asynchronously. If the chosen cluster transitions from `Ready` to `NotReady` in that window, the session is stuck in `Pending` with no stated recovery mechanism.

Suggested addition to the PlacementStrategy section:

```
Failure mode: if the placed cluster is NotReady when the reconciler processes the Pending session, the reconciler SHALL update the session with a Condition (type=PlacementFailed, reason=ClusterNotReady) and retry using the PlacementStrategy on the next reconcile cycle (which may select a different cluster). cluster_id is rewritten, not locked.
```

Without this, sessions silently stall in Pending with no user-visible explanation.

---

### [Major] `destination_cluster_id` on Application is architecturally ambiguous

The field is added to the Application model and listed in the coverage matrix, but the description is underspecified:

> "When set alongside `destination_ambient_url`, the Application syncs fleet definitions to the remote Ambient AND reconciles K8s resources to the specified cluster."

Who performs the K8s reconciliation? The local ApplicationReconciler talking directly to the remote cluster via `ClusterClientPool`? Or the remote Ambient's GatewayReconciler (triggered when the Application syncs a Gateway resource to the remote Ambient API)?

These are two very different architectures. The first couples the local CP to the remote cluster's K8s API. The second is the natural extension of "Application targets Ambient API, not K8s API" (a design decision the spec explicitly calls out). I believe the second is correct — but the spec needs to say so explicitly, and `destination_cluster_id` should probably be removed from the Application model if the answer is "the remote Ambient's reconcilers handle K8s."

---

### [Minor] Round-robin counter resets on pod restart

The RoundRobinPlacement description implies an in-memory counter, which resets on CP restart. Under HA or rolling deployments, this produces uneven distribution. Worth noting as a known limitation, or specify whether the counter should be persisted (e.g., in a ConfigMap or the `annotations` field on a Cluster).

---

### [Minor] Gateway annotation write-back mechanism unspecified

> "AND the annotation SHALL be written to the API server (PostgreSQL), not just to the Kubernetes resource"

How? The GatewayReconciler watches K8s events. Once it creates the external LoadBalancer/Route and observes its IP/hostname, it needs to make a REST call back to the API server to PATCH the Gateway's annotations. This is the same status write-back pattern used by the Session reconciler — worth calling out explicitly (e.g., "the GatewayReconciler calls `PATCH /api/ambient/v1/projects/{id}/gateways/{id}` to write the external URL annotation back to PostgreSQL after the external Service IP is assigned").

---

Once those are addressed, this is close to approvable. The overall design is sound.

— Amber

jsell-rh (changes requested):
## 🤖 Amber Review — PR #355: `spec: multi-cluster scheduling`

Prior feedback (credential watch, placement retry, Application cluster_id, RoundRobinPlacement prototype, gateway annotation write-back) — all addressed in the latest commit. Good work on those.

Two new gaps surfaced by reading the cross-cluster flow end-to-end:

---

### Critical — Cross-cluster mTLS credential delivery is unspecified

The spec describes the CP connecting to a gateway on a remote cluster via the `ambient-code.io/gateway-external-url` annotation. But it doesn't specify how the CP on the **workload cluster** obtains the gateway's **mTLS certificates** to establish the gRPC connection.

In same-cluster mode, the CP reads `openshell-server-tls` from the same namespace via its local K8s SA. In cross-cluster mode, that secret is on the **gateway cluster**, not the workload cluster's K8s API. The `ClusterClientPool` gives a `KubeClient` to talk to the remote K8s API — so the CP *could* read the secret from the gateway cluster using its remote `KubeClient`. But the spec doesn't describe this path.

Options to spec:
1. CP reads `openshell-server-tls` from the **gateway cluster** via `ClusterClientPool.ClientForCluster(gatewayClusterID)` at sandbox provisioning time.
2. GatewayReconciler syncs the TLS cert to the **workload cluster's** namespace at Gateway reconciliation time.
3. A new `ambient-code.io/gateway-ca-cert` annotation stores the PEM-encoded CA cert alongside the external URL.

The spec needs to define which approach is used, or cross-cluster mTLS will be silently broken. This is a security-critical gap.

---

### Major — `cluster_id` and `gateway_cluster_id` must be `readOnly` in OpenAPI

The spec says these fields are "set by PlacementStrategy at ignite time" (i.e., server-set). But if the OpenAPI schema doesn't mark them `readOnly: true`, clients can set them directly, bypassing placement logic entirely. A client could point sessions at clusters they don't have access to, or at `_local` to avoid scheduling to a GPU cluster.

The `SessionCreateRequest` and `SessionPatchRequest` schemas (once implemented) must explicitly exclude these fields from client-settable properties, or mark them `readOnly`. Please add a note to the spec's API reference section:

> `cluster_id` and `gateway_cluster_id` are server-managed fields. They SHALL be excluded from `SessionCreateRequest` and `SessionPatchRequest` schemas and marked `readOnly: true` in `Session`. Clients cannot directly set these fields; they are determined exclusively by `PlacementStrategy` at session ignite time.

---

### Minor — Partial provisioning on cluster-goes-NotReady between placement and reconcile

The spec says: _"the reconciler SHALL set a Condition (`type=PlacementFailed`, `reason=ClusterNotReady`) on the session and re-invoke the PlacementStrategy on the next reconcile cycle. The PlacementStrategy may select a different cluster — `cluster_id` is rewritten, not locked."_

This is correct for the pre-provisioning case. But what if the cluster goes NotReady **after** the namespace is created but **before** the Pod/Job is started? Rewriting `cluster_id` at that point would leave an orphaned namespace on the originally-selected cluster. Consider adding a note:

> If any K8s resources were already created on the original cluster before the `ClusterNotReady` condition is detected, the reconciler SHALL log a warning and proceed with cleanup of partially-created resources on the original cluster before retrying placement on a new cluster.

This is a corner case but worth documenting so implementers don't accidentally orphan namespaces.

---

**Confidence:** High (85%) — Spec-only change; the architecture is sound. The mTLS gap is the blocker; the readOnly schema note is a correctness guard that prevents accidental bypass.

*— Amber*

jsell-rh (approved):
## 🤖 Amber Review — PR #355: Round 5 (commit `a893d43`)

All three issues from my round 4 review are resolved. Approving.

---

### Verification against round 4 blockers

**[Critical — resolved] Cross-cluster mTLS credential delivery**
The spec now explicitly describes the path: CP reads `openshell-client-tls` from the gateway cluster via `ClusterClientPool.ClientForCluster(gatewayClusterID)`. TLS cache keyed by `(cluster_id, namespace)` — not namespace alone. Credential rotation on the Cluster resource evicts all TLS cache entries for that cluster. The cross-cluster mTLS scenario in `openshell-sandbox-provisioning.spec.md` is concrete and testable. ✅

**[Major — resolved] `cluster_id`/`gateway_cluster_id` readOnly**
ER diagram updated with `readOnly` in field comments. PlacementStrategy section now has explicit normative text: these are server-managed fields, excluded from create/patch request schemas, only writable by PlacementStrategy and KubeReconciler. ✅

**[Minor — resolved] Partial provisioning cleanup on re-placement**
`control-plane.spec.md` now specifies cleanup of partially-created resources on the original cluster before provisioning on the new one, with the correct behavior when the original cluster is unreachable (log warning, proceed — don't block re-placement). ✅

---

### CI
All applicable gates pass for this spec-only change. ✅

---

### What's solid across the full spec
- Credential watch → ClusterClientPool refresh chain (round 3 fix) ✅  
- PlacementStrategy retry on NotReady cluster (round 3 fix) ✅  
- `credential_id` FK reuse of existing Credential infrastructure (round 2 fix) ✅  
- Namespace provisioning on remote clusters via ProjectReconciler (round 2 fix) ✅  
- DELETE /clusters/{id} 409 on active sessions (round 2 fix) ✅  
- Round-robin prototype labelled as TODO for persistent counters / weighted scheduling ✅  
- GatewayReconciler write-back via REST PATCH described explicitly ✅  
- Application `destination_cluster_id` correctly removed ✅  
- Identity boundary guarantees (points 7, 8) still strong ✅  

This is a well-iterated spec. Solid work resolving five rounds of feedback without losing sight of the overall design coherence.

*— Amber*

jsell-rh (changes requested):
## 🤖 Amber — Implementation Review (commits `882040d`, `985bcf0`, `edda052`, `c382a3b`)

My prior spec approval stands. The implementation follows the spec's intent well and the plugin structure is clean. Three issues need addressing before this graduates from draft.

---

### [Major] `AutoMigrate` in `migration.go` will fail with `PreferSimpleProtocol` in testcontainers

`migration.go:202607150001` uses `tx.AutoMigrate(&Cluster{})`. This is the same issue I flagged in PR #363 for the OIDC migration — `AutoMigrate` fails with `PreferSimpleProtocol=true` which testcontainers PostgreSQL uses by default. Use explicit DDL instead:

```go
Migrate: func(tx *gorm.DB) error {
    if err := tx.Exec(`CREATE TABLE IF NOT EXISTS clusters (
        id              VARCHAR(36)  PRIMARY KEY,
        created_at      TIMESTAMPTZ,
        updated_at      TIMESTAMPTZ,
        deleted_at      TIMESTAMPTZ,
        name            TEXT         NOT NULL,
        description     TEXT,
        api_server_url  TEXT         NOT NULL,
        credential_id   VARCHAR(36),
        role            TEXT         NOT NULL DEFAULT 'hybrid',
        status          TEXT         NOT NULL DEFAULT 'Unknown',
        status_message  TEXT,
        labels          JSONB,
        annotations     JSONB,
        capacity        JSONB,
        last_heartbeat_at TIMESTAMPTZ
    )`).Error; err != nil {
        return err
    }
    return tx.Exec("CREATE UNIQUE INDEX IF NOT EXISTS idx_clusters_name ON clusters (name) WHERE deleted_at IS NULL").Error
},
```

This pattern is established by PR #363's fix (`ec533b5`). Integration tests in `integration_test.go` will fail on CI otherwise.

---

### [Major] `session_counter.go` silently returns 0 on column-not-found — could mask schema errors

`isColumnNotFoundError` is the right safety valve for when `cluster_id`/`gateway_cluster_id` don't exist yet on `sessions`. But returning `(0, nil)` silently means `Delete` proceeds even if the check was supposed to block it. At minimum add a warning log:

```go
if isColumnNotFoundError(err) {
    log := logger.NewLogger(ctx)
    log.Warnf("cluster_id column not found on sessions table — deregistration guard skipped: %v", err)
    return 0, nil
}
```

---

### [Minor] RBAC role seeding not visible in this PR

The commit message says "`cluster:admin`, `cluster:viewer` roles seeded via migration" but `migration.go` only creates the table. If these roles aren't seeded somewhere (RBAC plugin, separate migration), all Cluster API calls will fail `authzMiddleware.AuthorizeApi` at runtime. Please confirm where the seeding happens.

---

### What's solid ✅

- 8-file plugin pattern (`model`, `dao`, `service`, `handler`, `presenter`, `plugin`, `migration`, `session_counter`) ✅
- Advisory lock on `Replace` via `ClustersLockType` — prevents concurrent update races ✅
- `Delete` 409 guard via `sessionCounter.CountActiveSessionsOnCluster` — spec-compliant ✅
- `validIDPattern` on all path param reads — no injection risk ✅
- `validRoles` checked in both `Create` and `Patch` ✅
- No `panic()`, no token logging ✅
- Event emission on Create/Replace/Delete ✅
- Tests: service layer (8), presenter (3), integration (8) ✅
- `session_counter.go` queries both `cluster_id` and `gateway_cluster_id` ✅

**Confidence:** High — the `AutoMigrate` issue is reproducible from the established pattern. The sessionCounter gap is low blast-radius.

*— Amber*

jsell-rh (changes requested):
🤖 **Amber Review — PR #355: multi-cluster scheduling model (spec + Clusters API implementation)**

This is a large PR spanning spec updates across 6 files plus a full API server plugin implementation (Clusters CRUD, heartbeat, RBAC, tests). The spec work is thorough and the implementation follows established plugin patterns. Several issues block merge.

**Blockers:**

**1. `GetK8sClientsForRequest` not used in clusters handler**
The clusters API operates at global scope (not namespace-scoped per user), but the handler still needs to use the user's token for authorization checks. Looking at the plugin pattern — if `handler.go` makes any Kubernetes calls (e.g., to validate remote cluster connectivity), it must use `GetK8sClientsForRequest(c)`. If it doesn't make K8s calls at all, this is N/A, but confirm explicitly. The integration tests use `NewRandAccount()` + JWT auth, which is correct for the REST layer, but the service layer's `lockFactory` is set to `nil` in tests with a global `DisableAdvisoryLock = true` flag — this is a package-level side-effectful global that will bleed into other tests running in the same process. Use an interface or constructor injection instead.

**2. `DisableAdvisoryLock` package-level global is a test hygiene violation**
```go
DisableAdvisoryLock = true  // service_test.go
```
This is a mutable package-level global that permanently disables advisory locks for all tests in the process once set. This pattern can cause flaky tests if test order changes or tests run in parallel. Follow the existing pattern: pass a mock `lockFactory` via constructor injection rather than a global toggle.

**3. Integration tests bypass RBAC — missing `cluster:admin` role check**
`TestClusterCreate` creates a cluster using `NewRandAccount()` which is a regular user. The spec defines `cluster:admin` as required for create/delete. The integration test should verify that:
- A regular user without `cluster:admin` gets 403
- A user with `cluster:admin` succeeds
Currently the test only covers the happy path, not the RBAC enforcement. Per the ACP review checklist: "RBAC check performed before resource access."

**4. Author email is `u@example.com` in all commits**
Multiple commits on this PR (the spec and implementation commits) are authored by `user <u@example.com>`. This looks like a misconfigured git identity on the machine that ran Claude Code. Not a code issue, but worth flagging — squash commits before merge and ensure the final author is correct.

**Non-blocking:**

5. **Spec: `last_heartbeat_at` staleness-based health** — The spec defines `ClusterHealthSyncer` but doesn't specify what threshold makes a cluster transition from `Ready` → `Degraded` vs `Degraded` → `Offline`. Recommend defining: e.g., missed 2 heartbeats (2min) = Degraded, missed 5 (5min) = Offline. Otherwise implementers will have to decide this themselves.

6. **`session_counter.go` cross-plugin dependency** — `session_counter.go` queries the sessions table directly from the clusters plugin. Prefer going through the sessions service interface to avoid tight DB coupling between plugins. Low urgency for now but worth a TODO.

**What's good:**
- Full-stack spec coverage across 6 files is comprehensive
- 8-file plugin pattern followed correctly (dao, handler, migration, model, plugin, presenter, service, mock_dao)
- Deregistration guard (409 on DELETE with active sessions) is correctly implemented and tested
- Heartbeat endpoint design is clean
- `credential_id` FK over inline kubeconfig is the right call for security

**Confidence:** High — blockers 1-3 are clear violations; blocker 4 is cosmetic. Once resolved, this looks mergeable.

— Amber

jsell-rh (changes requested):
## 🤖 Amber Review — PR #355: Round 7 (commits `a8ad6c4`…`2acc52f`)

Good progress: `GetK8sClientsForRequest` in handler.go is confirmed non-issue (handler delegates entirely to service, no K8s calls at request time). The OpenAPI Clusters spec is clean with correct `readOnly` flags on server-managed fields. The deregistration guard is solid.

Two blockers from round 6 are still open, and two new issues surfaced in the latest commits.

---

### [Major — still open] `DisableAdvisoryLock` package-level global in tests

`service.go` declares:
```go
var (
    DisableAdvisoryLock     = false
    UseBlockingAdvisoryLock = true
)
```

`service_test.go`'s `newTestService` calls:
```go
DisableAdvisoryLock = true  // package-level mutation, never reset
```

This is a mutable package-level global that is permanently set to `true` for all tests in the process once touched — no `t.Cleanup`, no `defer`. Fix: add a `disableAdvisoryLock bool` field to `sqlClusterService` and set it via the constructor. The existing pattern elsewhere in the API server uses constructor injection for testability, not package-level toggles.

---

### [Major — still open] Integration tests don't verify RBAC enforcement

All 8 integration tests in `integration_test.go` use `NewRandAccount()` (a regular user with no `cluster:admin` role) and expect 2xx on Create/Patch/Delete. The spec and migration (`roles/202607150002`) define `cluster:admin` as required for write operations — but there's no test that verifies a regular user gets 403.

Add at minimum:
```go
func TestClusterCreateForbiddenWithoutClusterAdmin(t *testing.T) {
    account := h.NewRandAccount() // no cluster:admin
    jwtToken := h.NewAuthenticatedContext(account).Value(openapi.ContextAccessToken)
    // POST /clusters → expect 403
}
```

Per the ACP review checklist: "RBAC check performed before resource access." Without this, there is no proof the authz middleware is enforced on these routes.

---

### [Major — new] No DB-level FK constraint on `cluster_id` in sessions/gateways migrations

The `202607150003` and `202607150004` migrations add `cluster_id` as plain text columns with an index only:
```sql
ALTER TABLE sessions ADD COLUMN IF NOT EXISTS cluster_id TEXT
CREATE INDEX IF NOT EXISTS idx_sessions_cluster_id ON sessions(cluster_id)
```

There is no `REFERENCES clusters(id)` constraint. The deregistration guard in `ClusterService.Delete` is the only protection, but it only covers the application layer — if a cluster is deleted via direct DB access, or if the guard races with a concurrent session status change, `cluster_id` values in sessions/gateways will silently dangle.

Add a referential constraint:
```sql
ALTER TABLE sessions ADD COLUMN IF NOT EXISTS cluster_id TEXT REFERENCES clusters(id) ON DELETE SET NULL
```

`ON DELETE SET NULL` is the right semantic: deregistering a cluster nulls out the FK, which triggers the `IsNotFound` path in the reconciler cleanly without cascading deletes.

---

### [Minor — new] `cluster_id`/`gateway_cluster_id` missing from sessions/gateways OpenAPI schemas

`sessions/model.go` and `gateways/model.go` now have `ClusterId` and `GatewayClusterId` fields, but `openapi.sessions.yaml` and `openapi.gateways.yaml` are not updated. The spec says these are server-managed (`readOnly: true`), but clients have no visibility — they'll appear in JSON responses without being in the schema.

Add to the `Session` schema:
```yaml
cluster_id:
  type: string
  readOnly: true
  description: Workload cluster ID (server-managed, set by PlacementStrategy)
gateway_cluster_id:
  type: string
  readOnly: true
  description: Gateway cluster ID (server-managed, set by PlacementStrategy)
```

And to the `Gateway` schema:
```yaml
cluster_id:
  type: string
  readOnly: true
  description: Target cluster for cross-cluster gateway deployment
```

---

### Resolved ✅

- `GetK8sClientsForRequest` — handler.go makes zero K8s calls, delegate to service only. Confirmed non-issue.
- Spec rounds 1–5 blockers all remain resolved (mTLS delivery, credential_id FK, namespace provisioning, deregistration 409, readOnly field comments in ER diagram).

---

**Confidence:** High — the `DisableAdvisoryLock` and RBAC test issues are clear from the code; the missing FK constraint is a referential integrity gap the app-layer guard alone can't cover.

*— Amber*

jsell-rh (changes requested):
## 🤖 Amber Review — PR #355: Round 8 (commit `aff3f7a`)

The latest commit is pure SDK/OpenAPI client regeneration for the Cluster kind — no impact on the three blockers from round 7. All three remain unresolved.

---

### [Major — still open] `DisableAdvisoryLock` package-level global in tests

`service.go:14-17`:
```go
var (
    DisableAdvisoryLock     = false
    UseBlockingAdvisoryLock = true
)
```

`service_test.go:44-50` — `newTestService` sets `DisableAdvisoryLock = true` globally with no `t.Cleanup` or `defer` to reset it. This permanently mutates the package state for the entire test binary. Fix: add a `disableAdvisoryLock bool` field to `sqlClusterService` and thread it through `NewClusterService`. The constructor injection pattern is established elsewhere in the codebase.

---

### [Major — still open] Integration tests don't verify RBAC enforcement

`integration_test.go` — all 8 write operations (Create, Patch, Delete, Heartbeat) use `NewRandAccount()` which has no `cluster:admin` role. There is no test that sends a write request without `cluster:admin` and expects 403. Per the ACP review checklist: "RBAC check performed before resource access." Without a negative test, there is no CI proof the authz middleware is enforced on these routes.

Add at minimum:
```go
func TestClusterCreateForbiddenWithoutClusterAdmin(t *testing.T) {
    h, _ := test.RegisterIntegration(t)
    account := h.NewRandAccount() // no cluster:admin
    ctx := h.NewAuthenticatedContext(account)
    jwtToken := ctx.Value(openapi.ContextAccessToken)
    body, _ := json.Marshal(map[string]interface{}{
        "name": "forbidden-cluster", "api_server_url": "https://k8s.example.com:6443", "role": "hybrid",
    })
    resp, err := resty.R().
        SetHeader("Authorization", fmt.Sprintf("Bearer %s", jwtToken)).
        SetBody(body).Post(h.RestURL("/clusters"))
    Expect(err).NotTo(HaveOccurred())
    Expect(resp.StatusCode()).To(Equal(http.StatusForbidden))
}
```

---

### [Major — still open] No FK constraint on `cluster_id` in sessions/gateways migrations

`sessions/migration.go:clusterPlacementMigration()` adds `cluster_id TEXT` with an index but no `REFERENCES clusters(id)`. The application-layer deregistration guard in `ClusterService.Delete` can be bypassed (concurrent transaction race, direct DB access, future migration that removes the guard). 

Fix — change the sessions migration to:
```sql
ALTER TABLE sessions ADD COLUMN IF NOT EXISTS cluster_id TEXT REFERENCES clusters(id) ON DELETE SET NULL
ALTER TABLE sessions ADD COLUMN IF NOT EXISTS gateway_cluster_id TEXT REFERENCES clusters(id) ON DELETE SET NULL
```

`ON DELETE SET NULL` is the correct semantic: deregistering a cluster nulls the FK, which the reconciler handles via its existing `IsNotFound` path, without cascading deletes.

---

### [Minor — still open] `cluster_id`/`gateway_cluster_id` missing from Session OpenAPI schema

`openapi.sessions.yaml` — the `Session` schema has no `cluster_id` or `gateway_cluster_id` fields. These are returned in JSON responses (they're in `model.go`) but invisible to API clients. Add:

```yaml
cluster_id:
  type: string
  readOnly: true
  description: Workload cluster ID (server-managed, set by PlacementStrategy)
gateway_cluster_id:
  type: string
  readOnly: true
  description: Gateway cluster ID (server-managed, set by PlacementStrategy)
```

The SDK regeneration in this commit (`aff3f7a`) only covers `Cluster` endpoints — the Session SDK is untouched and won't expose these fields until the schema is updated.

---

### [Minor — carry forward] `clusters/migration.go` uses `AutoMigrate`

`clusters/migration.go:202607150001` still uses `tx.AutoMigrate(&Cluster{})`. This is the same pattern flagged in round 6 — it fails with `PreferSimpleProtocol=true` in testcontainers. Use explicit DDL as established by the fix in `ec533b5`. This will cause integration test failures in CI.

---

### What the SDK regeneration commit does ✅

The `aff3f7a` commit correctly regenerates Go SDK client methods for the 7 Cluster endpoints (`listClusters`, `createCluster`, `getCluster`, `patchCluster`, `deleteCluster`, `getClusterStatus`, `clusterHeartbeat`) and adds `ClusterList`, `ClusterPatchRequest`, `ClusterStatusResponse` types. SDK version bump across all existing client files is clean. No issues.

---

**Confidence:** High — all five blockers are directly observable in the changed files.

*— Amber*

PR #357, opened 2026-07-14
perf(kind): speed up inner-loop rebuild with change detection and parallelism

jeremyeder:
these numbers were gathered for real on my laptop.

## Verified impact — `make kind-rebuild` (Docker, arm64)

| Scenario | Baseline (main) | This PR | Improvement |
|---|---|---|---|
| 1-file UI change | 184s (rebuild all 3, serial) | 54s (rebuild only UI) | **−130s · ~3.4×** |
| 3-component change | 184s (serial) | 86s (3 in parallel) | **−98s · ~2.1×** |

## Changes

- **`kind-rebuild` change detection**: only rebuild components whose paths changed (`git diff` vs HEAD + staged + untracked). `ambient-sdk` changes fan out to all consumers; `ambient-api-server` changes also rebuild control-plane (it copies the api-server dir and imports its gRPC package); safe fallback to rebuilding all when nothing matches.
- **Parallelism**: `kind-rebuild` and `_kind-load-images` run reloads/image loads in parallel with per-PID exit-code propagation instead of serially.

## Codex review (scoped to this diff)

Re-reviewed clean. Two earlier findings both addressed:
- **Change-detection dependency — resolved.** api-server changes now also rebuild control-plane.
- **Default-overlay pull policy — resolved.** Dropped the `imagePullPolicy: IfNotPresent` change to the default Quay overlay (it could serve cached `:latest` on persistent clusters); the default path keeps `Always`, and `IfNotPresent` stays scoped to `kind-local`. Local iteration is unaffected — `kind-reload-*` uses uniquely-tagged images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1266,9 +1266,42 @@ screenshots-clean: ## Remove generated screenshots
 	@echo "$(COLOR_GREEN)✓$(COLOR_RESET) Screenshot output cleaned"
 
 kind-rebuild: check-kind check-kubectl check-local-context _kind-require-cluster ## Rebuild, reload, and restart all components in kind
-	@$(MAKE) --no-print-directory kind-reload-ambient-ui
-	@$(MAKE) --no-print-directory kind-reload-ambient-control-plane
-	@$(MAKE) --no-print-directory kind-reload-ambient-api-server
+	@_CHANGED=$$(git diff --name-only HEAD -- 2>/dev/null; git diff --name-only --cached -- 2>/dev/null; git ls-files --others --exclude-standard -- 2>/dev/null); \
+	_DO_UI=false; _DO_CP=false; _DO_API=false; \
+	if [ -z "$$_CHANGED" ]; then \
+		echo "$(COLOR_YELLOW)▶$(COLOR_RESET) No changed files detected — rebuilding all components"; \
+		_DO_UI=true; _DO_CP=true; _DO_API=true; \
+	else \
+		if echo "$$_CHANGED" | grep -q '^components/ambient-sdk/'; then \
+			echo "$(COLOR_YELLOW)▶$(COLOR_RESET) ambient-sdk changed — rebuilding all consumers"; \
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
## 🤖 Amber Review — `perf(kind): speed up inner-loop rebuild with change detection and parallelism`

**Verdict: Approved.** Solid, well-reasoned DX improvement with correct dependency tracking and safe fallback semantics.

**What's working well:**
- Change detection logic is accurate: `git diff HEAD + staged + untracked` covers all dirty states a developer will hit during iteration.
- The `ambient-api-server → control-plane` fan-out is explicitly handled and the inline comment explains why — good. This is the kind of subtle dependency that bites people silently.
- `ambient-sdk` fanning out to all consumers is correct given the SDK is generated from OpenAPI.
- Fallback to full rebuild when nothing matches is exactly right — opt for correctness over speed on unknown paths.
- PID-based exit-code propagation in the parallel loop is the correct POSIX approach; `wait $pid` returns the child's exit code.
- Parallelizing `_kind-load-images` removes a previously serial bottleneck — the ~3.4× improvement on a 1-file UI change is credible and matches my mental model.

**One minor observation** (not a blocker):

The `_CHANGED` variable collects untracked files via `git ls-files --others --exclude-standard`. This means a new file that hasn't been `git add`-ed yet (e.g. a freshly created `components/ambient-ui/src/foo.tsx`) correctly triggers a UI rebuild. That's the right behavior, but `git ls-files --others` can also pick up generated files, vendored artifacts, or temp files in component paths if `.gitignore` is incomplete. Given the project has a `.pre-commit-config.yaml` and `.gitignore` in place, this is low risk — just worth knowing if someone sees unexpected rebuild triggers.

**Constitution compliance:** No `panic()`, no secrets/tokens, no frontend `any` types. This is dev tooling (Makefile), so most production guardrails don't apply. The parallelism pattern correctly propagates errors rather than swallowing them.

*— Amber*

jsell-rh:

bsquizz (approved):

PR #358, opened 2026-07-14
perf(kind): speed up cold kind-up start

jeremyeder:
these numbers were gathered for real on my laptop. i havent seen any issues with keycloak not being ready in the reduced time.

## Verified impact — cold `make kind-up` (Docker, arm64)

| Metric | Baseline (main) | This PR | Improvement |
|---|---|---|---|
| Cold `kind-up` total | 250s | 190s | **−60s · −24%** |
| Keycloak → Ready | 62s | 23s | **−39s** |
| control-plane rollout revisions | 3 | 2 | −1 (redundant rollout removed) |

## Changes

- **No double-rollout**: inject `OPENSHELL_USE_GATEWAY` env before the single readiness wait and drop the redundant post-wait `kubectl rollout status` pair — eliminates a second full rollout of api-server + control-plane after they were already Ready.
- **Keycloak probes**: lower readiness `initialDelaySeconds` 60→15 and liveness 120→30 so Keycloak is marked Ready as soon as the realm responds (tolerances preserved via unchanged `periodSeconds`/`failureThreshold`).
- **Dead-infra cleanup**: remove `tests/infra/load-images.sh` (unreferenced, stale image set) and unused `overlays/kind` `ambient-api-server-db{,-notls}-patch.yaml`; drop the stale `load-images.sh` row from `tests/README.md`.

> Note: parallelizing `wait-for-ready.sh` was evaluated and **dropped** — no measurable speedup (`kubectl wait` only observes already-concurrent pod startup).

## Codex review (scoped to this diff)

No actionable regressions identified. The Kind overlay renders successfully with `kubectl kustomize`. (CodeRabbit unavailable — local CLI signed out.)


jsell-rh (approved):
## 🤖 Amber Review — `perf(kind): speed up cold kind-up start`

**Verdict: Approved.** Clean, targeted optimization with correct ordering logic and safe dead-infra cleanup.

**What's working well:**

- **Reordering `wait-for-ready.sh` after `kubectl set env`** is the right fix. The previous flow: (1) wait for pods ready, (2) inject env var, (3) rollout status — caused a redundant rollout because the env injection triggered another restart after readiness was already confirmed. Moving the wait to after the env injection means pods only need to reach Ready once. The benchmark numbers (−39s on Keycloak) are plausible.
- **Keycloak probe tuning**: `initialDelaySeconds 60→15` (readiness) and `120→30` (liveness) are aggressive but defensible — `failureThreshold: 10` with `periodSeconds: 10` gives 100s total tolerance for readiness (15s base + 100s retries), and `failureThreshold: 5` with `periodSeconds: 30` gives 150s total for liveness. Both are well above the observed 23s startup time the author reports. Conservative enough to not flap on slow hardware.
- **Dead file removal**: The deleted `ambient-api-server-db-{,-notls}-patch.yaml` files are NOT referenced in the current `kustomization.yaml` (I verified — the kustomization references `api-server-db-security-patch.yaml` and `api-server-security-patch.yaml` instead). Safe to delete.
- **`load-images.sh` deletion**: The script was already superseded by the `_kind-load-images` Makefile target. Removing it and the README row is correct housekeeping.

**One thing to verify** (not a blocker, but worth a quick check before merge):

The PR description mentions Keycloak was measured at 23s startup on the author's machine (Docker, arm64). On slower CI hardware or x86_64, Keycloak startup can be longer, especially on first pull when the JVM is cold. The `failureThreshold: 10` buffer provides ~100s of headroom past `initialDelaySeconds: 15` — that's 115s total, which should be fine for any reasonable environment. If CI starts seeing flaky Keycloak readiness failures after this merges, the immediate fix is to bump `initialDelaySeconds` back to 30 (not 60 — the old 60 was clearly over-provisioned).

**Constitution compliance:** No production code changes. Manifest cleanup is safe (verified kustomization references). Probe values preserve `failureThreshold`/`periodSeconds` (the "tolerances" the author references are intact).

*— Amber*

bsquizz (approved):

PR #359, opened 2026-07-14
feat(cli): add acpctl get gateway with connection info and --setup

JuanmaBM:
## Summary
- Add `acpctl get gateways` to list/get gateway resources with connection info (cluster DNS, SANs, TLS secret)
- Add `acpctl get gateway <name> --setup` to auto-configure the openshell CLI (extract certs, port-forward, register gateway)
- Add `acpctl delete gateway` support
- Gateway lookup works by both name and ID

### DIFF
--- a/components/ambient-cli/cmd/acpctl/get/cmd.go
+++ b/components/ambient-cli/cmd/acpctl/get/cmd.go
@@ -993,9 +993,8 @@ func printGatewayConnectionInfo(w io.Writer, gw *sdktypes.Gateway) {
 	if len(gw.ServerDnsNames) > 0 {
 		fmt.Fprintf(w, "  Server SANs:  %s\n", strings.Join(gw.ServerDnsNames, ", "))
 	}
-	fmt.Fprintf(w, "  TLS Secret:   openshell-server-tls (namespace: %s)\n", namespace)
 	fmt.Fprintf(w, "\nSetup openshell CLI:\n")
-	fmt.Fprintf(w, "  acpctl gateway setup-cli %s\n", gw.Name)
+	fmt.Fprintf(w, "  acpctl gateway setup-cli %s --gateway-url <url>\n", gw.Name)
 }
 
 func sessionChanged(old, current sdktypes.Session) bool {
--- a/components/ambient-cli/cmd/acpctl/get/cmd.go
+++ b/components/ambient-cli/cmd/acpctl/get/cmd.go
@@ -157,6 +157,9 @@ func run(cmd *cobra.Command, cmdArgs []string) error {
 	case "gateways":
 		return getGateways(ctx, client, printer, name)
 	default:
+		if args.setup {
+			return fmt.Errorf("--setup is only supported for the 'gateway' resource type")
+		}
 		return fmt.Errorf("unknown resource type: %s\nValid types: sessions, projects, project-agents, project-settings, users, agents, providers, policies, roles, role-bindings, credentials, gateways", cmdArgs[0])
 	}
 }
@@ -957,22 +960,28 @@ func getGateways(ctx context.Context, client *sdkclient.Client, printer *output.
 }
 
 func findGateway(ctx context.Context, client *sdkclient.Client, nameOrID string) (*sdktypes.Gateway, error) {
-	// Try by ID first
 	gw, err := client.Gateways().Get(ctx, nameOrID)
 	if err == nil {
 		return gw, nil
 	}
 
-	// Fall back to name search
-	opts := sdktypes.NewListOptions().Size(100).Build()
-	list, err2 := client.Gateways().List(ctx, opts)
-	if err2 != nil {
-		return nil, fmt.Errorf("list gateways: %w", err2)
-	}
-	for i := range list.Items {
-		if list.Items[i].Name == nameOrID {
-			return &list.Items[i], nil
+	page := 1
+	pageSize := 100
+	for {
+		opts := sdktypes.NewListOptions().Page(page).Size(pageSize).Build()
+		list, err2 := client.Gateways().List(ctx, opts)
+		if err2 != nil {
+			return nil, fmt.Errorf("list gateways: %w", err2)
 		}
+		for i := range list.Items {
+			if list.Items[i].Name == nameOrID {
+				return &list.Items[i], nil
+			}
+		}
+		if len(list.Items) < pageSize {
+			break
+		}
+		page++
 	}
 	return nil, fmt.Errorf("gateway %q not found", nameOrID)
 }
@@ -1037,11 +1046,11 @@ func setupOpenshellGateway(w io.Writer, gw *sdktypes.Gateway) error {
 	fmt.Fprintf(w, "Starting port-forward to openshell-gateway in %s...\n", namespace)
 	pfCmd := exec.Command("kubectl", "port-forward", "-n", namespace,
 		"statefulset/openshell-gateway", ":8080")
+	pfCmd.Stderr = os.Stderr
 	pfOut, err := pfCmd.StdoutPipe()
 	if err != nil {
 		return fmt.Errorf("create port-forward pipe: %w", err)
 	}
-	pfCmd.Stderr = pfCmd.Stdout
 	if err := pfCmd.Start(); err != nil {
 		return fmt.Errorf("start port-forward: %w", err)
 	}
@@ -1053,15 +1062,18 @@ func setupOpenshellGateway(w io.Writer, gw *sdktypes.Gateway) error {
 	// Wait for port-forward to print the assigned port
 	port := ""
 	scanner := bufio.NewScanner(pfOut)
-	deadline := time.After(15 * time.Second)
 	portCh := make(chan string, 1)
+	pfCtx, pfCancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer pfCancel()
 	go func() {
+		defer pfCancel()
 		for scanner.Scan() {
 			line := scanner.Text()
 			if idx := strings.Index(line, "Forwarding from 127.0.0.1:"); idx >= 0 {
 				rest := line[idx+len("Forwarding from 127.0.0.1:"):]
 				if end := strings.Index(rest, " "); end > 0 {
 					portCh <- rest[:end]
+					return
 				}
 			}
 		}
@@ -1069,7 +1081,7 @@ func setupOpenshellGateway(w io.Writer, gw *sdktypes.Gateway) error {
 
 	select {
 	case port = <-portCh:
-	case <-deadline:
+	case <-pfCtx.Done():
 		return fmt.Errorf("timeout waiting for port-forward to start")
 	}
 
--- a/components/ambient-cli/cmd/acpctl/get/cmd.go
+++ b/components/ambient-cli/cmd/acpctl/get/cmd.go
@@ -2,15 +2,11 @@
 package get
 
 import (
-	"bufio"
 	"context"
-	"encoding/base64"
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
 	"os/signal"
-	"path/filepath"
 	"strings"
 	"time"
 
@@ -27,7 +23,6 @@ var args struct {
 	limit        int
 	watch        bool
 	watchTimeout time.Duration
-	setup        bool
 }
 
 var Cmd = &cobra.Command{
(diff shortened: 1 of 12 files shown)


jsell-rh (changes requested):
## 🤖 Amber Review — PR #359: `feat(cli): add acpctl get gateway`

Good direction — gateway CLI support is a real gap. A few issues to address before this is ready.

**Summary:** Three actionable issues (one Critical, two Major). The overall structure is sound; the `--setup` flow is the part that needs the most attention.

**Confidence:** High (90%) — CLI-only change, no API surface changes.

---

### Critical

**Scanner goroutine leak in `setupOpenshellGateway`**

The scanner goroutine (`go func() { for scanner.Scan() ... }()`) has no termination signal when the `deadline` case fires. The goroutine will block indefinitely on `scanner.Scan()` after `setupOpenshellGateway` returns (or even while the port-forward is still running after the deadline path). Close `pfOut` or use a context to stop it:

```go
portCh := make(chan string, 1)
ctx2, cancel2 := context.WithTimeout(context.Background(), 15*time.Second)
defer cancel2()
go func() {
    defer cancel2()
    for scanner.Scan() {
        line := scanner.Text()
        if idx := strings.Index(line, "Forwarding from 127.0.0.1:"); idx >= 0 {
            rest := line[idx+len("Forwarding from 127.0.0.1:"):]
            if end := strings.IndexByte(rest, ' '); end > 0 {
                portCh <- rest[:end]
                return
            }
        }
    }
}()
select {
case port = <-portCh:
case <-ctx2.Done():
    return fmt.Errorf("timeout waiting for port-forward to start")
}
```

---

### Major

**`pfCmd.Stderr = pfCmd.Stdout` before `StdoutPipe` — Stderr output is silently dropped**

`pfCmd.Stdout` is `nil` at the point this line runs (before `StdoutPipe()` is called). `StdoutPipe()` sets `pfCmd.Stdout` internally *after* this assignment, so `Stderr` remains `nil` and all stderr from `kubectl port-forward` is discarded. This makes diagnosing port-forward startup failures much harder.

Either pipe stderr separately or route it to `os.Stderr`:
```go
pfCmd.Stderr = os.Stderr
pfOut, err := pfCmd.StdoutPipe()
```

---

### Major

**`findGateway` name search hardcoded at `Size(100)` — silently truncates**

If a project has >100 gateways, a name-based lookup will miss gateways past the first page. Either paginate or document the limitation. For a CLI tool this is unlikely to matter in practice today, but the silent truncation is the problem — at minimum add a comment:

```go
// NOTE: name search is limited to the first 100 results; use the gateway ID for large fleets.
opts := sdktypes.NewListOptions().Size(100).Build()
```

---

### Minor

**`--setup` flag registered globally but is gateways-only**

`acpctl get sessions --setup` silently accepts the flag and fails at runtime with a confusing error. Consider guarding earlier:

```go
case "gateways":
    return getGateways(ctx, client, printer, name)
default:
    if args.setup {
        return fmt.Errorf("--setup is only supported for the 'gateways' resource type")
    }
    return fmt.Errorf("unknown resource type: ...")
```

---

*— Amber*

bsquizz:
> Add acpctl get gateway <name> --setup to auto-configure the openshell CLI (extract certs, port-forward, register gateway)

Awesome! Let's mark this as a TODO in the code -- in future, once we support setup of the gateway using a Route/Ingress, we will want to use that instead of the port-forward

jsell-rh (approved):
## 🤖 Amber Re-Review — PR #359: `feat(cli): add acpctl get gateway`

All four issues from my prior review are resolved. Good turnaround.

**What was fixed:**
- **Critical — goroutine leak**: Correctly addressed. `pfCtx`/`pfCancel` with `defer pfCancel()` inside the scanner goroutine means the goroutine exits as soon as the context deadline fires or the port line is found. The outer `defer pfCmd.Process.Kill()` closes stdout, unblocking any remaining `scanner.Scan()` call. Clean.
- **Major — `pfCmd.Stderr = os.Stderr`**: Now runs before `StdoutPipe()`. stderr from `kubectl port-forward` is correctly routed.
- **Major — pagination in `findGateway`**: Loop with `page++` and `break` when `len(list.Items) < pageSize` is correct. Handles arbitrarily large gateway lists.
- **Minor — `--setup` guard**: The `default` case now returns an early error when `--setup` is set on an unknown resource type. `acpctl get sessions --setup` still silently accepts the flag (it's filtered before hitting the `default` case) — this is a minor rough edge, not a blocker.

**One open nit (not blocking):** @bsquizz asked for a `TODO` comment in `setupOpenshellGateway` about Route/Ingress as a future alternative to port-forward. That's not in the code yet — worth picking up as a follow-on to avoid the comment getting lost.

**Confidence:** High (95%) — CLI-only change, no API surface, all review blockers resolved.

*— Amber*

components/ambient-cli/cmd/acpctl/get/cmd.go
```diff
@@ -23,6 +27,7 @@ var args struct {
 	limit        int
 	watch        bool
 	watchTimeout time.Duration
+	setup        bool
```
squizzi: I think `--setup` should be extracted into a subcommand instead of a flag, based on @jeremyeder 's comments today on the call a getter should never setup config.

An `acpctl gateway setup` command seems appropriate here.

squizzi (dismissed):

jsell-rh:
## 🤖 Amber — Note on squizzi's review

Noting @squizzi's `CHANGES_REQUESTED` on commit `840da37`: the feedback is correct — `get` commands should be read-only and `--setup` is a mutation that configures local state. The precedent from @jeremyeder's call is right.

`acpctl gateway setup [name]` as a dedicated subcommand is the better UX:
- Follows the principle of least surprise (get = read, named verb = action)
- Makes the command discoverable via `acpctl gateway --help`
- Allows `--setup` to be removed from the shared get flag set, which eliminates the silent-accept-then-runtime-error for other resource types

My prior approval was on the implementation correctness of the flag-based approach and all the internals (goroutine lifecycle, stderr routing, pagination). The refactor to a subcommand is a CLI shape change — the guts of `setupOpenshellGateway` don't need to change, just the registration path. 

Deferring to squizzi + jeremyeder on the UX call. Happy to re-approve once the subcommand refactor lands.

*— Amber*

components/ambient-cli/cmd/acpctl/get/cmd.go
bsquizz: Can we go with something that makes it obvious it applies to the CLI: `acpctl gateway setup-cli`

jsell-rh (approved):
## 🤖 Amber Re-Review — PR #359: commits `b339d9d` + `e246f08`

Responding to squizzi's and bsquizz's feedback. Both commits are clean — approving.

**What changed:**
- `--setup` flag removed from `acpctl get gateway`; logic extracted to new `gateway/` package as `acpctl gateway setup-cli <name>` — correct call, get commands should be read-only
- Rename `setup` → `setup-cli` addresses bsquizz's naming request; makes it unambiguous the command configures the *local CLI*, not the gateway resource itself
- `printGatewayConnectionInfo` hint updated to match new command (`acpctl gateway setup-cli %s`) ✅

**Internals verified unchanged from `840da37`:**
- Goroutine leak fix (pfCtx + pfCancel) ✅
- `pfCmd.Stderr = os.Stderr` before `StdoutPipe` ✅
- `findGateway` pagination loop ✅
- `tls.key` written with `0600`, other certs `0644` ✅
- `defer pfCmd.Process.Kill()` + `pfCmd.Wait()` cleanup ✅

**Minor note (non-blocking):** `findGateway` is now duplicated between `get/cmd.go` and `gateway/setup.go`. Fine for now — extracting to a shared `pkg/gateways` helper would be a clean follow-up if the function diverges.

**Confidence:** High (95%) — refactor only, all previously-reviewed logic preserved, reviewer feedback fully addressed.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — Follow-up on commit `f1a04450`

Trivial change: `StringVar` → `StringVarP` adding `-U` as a short alias for `--api-url`. No logic change, no new flags registered, consistent with how other flags in the CLI are aliased.

Approval stands.

*— Amber*

squizzi:
Code has significantly changed, removing amber review approval

jsell-rh (changes requested):
## 🤖 Amber Re-Review — PR #359: commits `f5bc191` + `c59366a`

Two new commits since my last approval (Jul 14 @ `cf3e9d5`). The OIDC password grant flow, non-interactive gateway config writing, mTLS cert fetch, and deterministic Kind port-forwards are solid additions. One logic bug in `resolveLocalName` needs fixing before merge; everything else is a nit.

**Confidence:** High (92%) — CLI-only change, no API surface, no Go backend or frontend touched.

---

### Major

**`resolveLocalName` dead code — suffix loop is unreachable**

`components/ambient-cli/cmd/acpctl/gateway/setup.go`

```go
func resolveLocalName(project, apiGWName string) string {
	base := project + "-" + apiGWName
	existing := listOpenshellGateways()
	if existing[base] {
		return base      // ← exits if base IS registered
	}
	if !existing[base] {
		return base      // ← exits if base IS NOT registered (exhausts all cases)
	}
	// DEAD CODE — for i := 1; ; i++ { ... } is never reached
```

`existing[base]` is either true or false — the two branches cover every case. The suffix loop can never execute. The spec says _"with a numeric suffix added if a new registration is needed and the name is already taken"_, but the implementation never adds a suffix. The function is effectively `return project + "-" + apiGWName`.

If the suffix behavior is intentional (reserve for future multi-gateway per-project support), collapse to the simple form and add a comment. If it's a logic inversion bug, fix to:

```go
func resolveLocalName(project, apiGWName string) string {
    base := project + "-" + apiGWName
    // If already registered, reuse the name (re-auth path in setupOpenshellGateway)
    if listOpenshellGateways()[base] {
        return base
    }
    return base // Name is free — use as-is
}
```

Or if the spec really intends suffix-based deduplication for NEW registrations when `base` is taken by a *different* gateway:

```go
func resolveLocalName(project, apiGWName string) string {
    base := project + "-" + apiGWName
    existing := listOpenshellGateways()
    if existing[base] {
        return base  // already registered: re-auth, not new registration
    }
    // base not registered — but might be in use for a different endpoint
    // For now return base; suffix support is a future extension.
    return base
}
```

Either way, remove the dead loop.

---

### Nits (non-blocking)

**`buildAddArgs` uses `oidc.Audience` for both `--oidc-client-id` and `--oidc-audience`**

This is consistent with the gateway schema having only `Issuer` + `Audience`, so it's probably correct. If `GatewayOidc` ever gains a separate `ClientID` field, this will silently misbehave. Worth a comment:

```go
// Gateway schema stores a single Audience field; client-id and audience
// resolve to the same value until the schema adds a distinct ClientID.
"--oidc-client-id", oidc.Audience,
"--oidc-audience", oidc.Audience,
```

**`fetchClientTLS` kubectl calls have no timeout**

`exec.Command("kubectl", ...)` without context; if the cluster is unreachable this blocks indefinitely. The parent `runSetup` does set a timeout via `cfg.GetRequestTimeout()` but that ctx isn't threaded through. Non-fatal for a local dev tool but worth a follow-on.

**`--password` is exposed in shell history and `ps` output**

Worth a one-liner in the command help text: _"Avoid using --password on shared systems; prefer environment variable substitution."_ Not blocking — the flag is explicitly for local Kind dev and CI.

---

*— Amber*

jsell-rh:
**Inline note on `resolveLocalName` (setup.go:645–659):**

```go
func resolveLocalName(project, apiGWName string) string {
    base := project + "-" + apiGWName
    existing := listOpenshellGateways()
    if existing[base] {
        return base      // exits when base IS registered
    }
    if !existing[base] {
        return base      // exits when base IS NOT registered — covers all cases
    }
    for i := 1; ; i++ { // ← DEAD CODE, never reachable
        ...
    }
}
```

The two `if` branches are mutually exhaustive — the suffix loop is unreachable. Simplest fix: remove both `if` blocks and the loop, just `return project + "-" + apiGWName`.

*— Amber*

jsell-rh:
_Adding `amber/change-requested` label for the `resolveLocalName` dead-code bug._

jsell-rh (approved):
## 🤖 Amber Re-Review — PR #359: commits `71e810b` + `8122990` + `8ef7484`

All issues from my prior changes-requested review are resolved. Clean close.

**What changed:**
- **Major — `resolveLocalName` dead code**: Fixed in `8122990`. Both exhaustive `if` / `if !` branches collapsed to `return project + "-" + apiGWName`. Dead suffix loop removed. Correct.
- **Connectivity validation** (`71e810b`): `verifyGateway()` calls `openshell status -g <localName>` after setup. On failure, `cleanupGatewayConfig()` removes the written config dir for new registrations (preserves existing for re-auth). Spec updated + scenario added. Solid.
- **Makefile `kind-status`** (`8122990`): gateway port loop from `71e810b` simplified to a single-line `$(KIND_FWD_GATEWAY_BASE_PORT)+` static display. Both approaches are fine; the deterministic port scheme (from `f5bc191`) makes the static display accurate.
- **Error string cleanup** (`8ef7484`): Trailing `\n` removed from `fmt.Errorf` in `setupOpenshellGateway`. Correct — error strings should not have trailing newlines.

**Nits from prior review (status):**
- `buildAddArgs` OIDC audience comment — still not added, still non-blocking.
- `fetchClientTLS` kubectl timeout — still not added, still non-blocking follow-on.
- `--password` shell history warning in help text — still not added, non-blocking.

None of these are blocking. The PR is in good shape.

**Confidence:** High (95%) — CLI-only, no API surface, no Go backend or frontend touched, all Major findings resolved.

*— Amber*

PR #363, opened 2026-07-14
feat(gateway): optional OIDC authentication for OpenShell gateways

bsquizz:
## Summary

- Add optional OIDC authentication support to OpenShell gateways, configurable per-gateway via a nested `oidc` object
- When OIDC is enabled, the control plane injects `[openshell.gateway.oidc]` into gateway.toml and flips `allow_unauthenticated_users` to `false`
- Full-stack implementation across API server, control plane, all three SDKs (Go/Python/TS), CLI, and Kind test infrastructure

### API Server
- `GatewayOidc` OpenAPI schema with fields: issuer, audience, jwks_ttl, roles_claim, admin_role, user_role, scopes_claim
- JSONB storage column with migration `202607140001`
- PATCH support for updating OIDC config on existing gateways

### Control Plane
- OIDC config injection into gateway.toml TOML configuration
- Validation: issuer must be HTTP(S) URL, admin_role/user_role must both be set or both empty
- Change detection triggers gateway pod restart when OIDC config changes

### SDKs & CLI
- Go, Python, and TypeScript SDK types with builder methods
- Kustomize merge support for `oidc` field in resource definitions
- `acpctl apply` handles OIDC from kustomize resources on both create and update paths

### Kind Test Infrastructure
- `tenant-c` overlay with OIDC-enabled gateway pointing to Keycloak
- Keycloak `openshell-cli` public client with audience and realm-roles protocol mappers
- `openshell-admin` and `openshell-user` realm roles mapped to admin/developer users

## Spec
- [Gateway OIDC Authentication Spec](specs/platform/gateway-oidc.spec.md)

## Test plan
- [ ] `cd components/ambient-api-server && go build ./...`
- [ ] `cd components/ambient-control-plane && go build ./...`
- [ ] `cd components/ambient-cli && go build ./...`
- [ ] `cd components/ambient-sdk/go-sdk && go build ./...`
- [ ] `make kind-up` provisions tenant-c with OIDC gateway
- [ ] Verify gateway.toml contains `[openshell.gateway.oidc]` section in tenant-c gateway pod
- [ ] Verify Keycloak realm has openshell-cli client and realm roles
- [ ] PATCH gateway OIDC config and verify pod restart


jsell-rh (approved):
🤖 **Amber Review — Approved** (spec draft)

**Summary:** Well-structured spec for optional OIDC authentication on Gateway resources. Covers the API shape, reconciler behavior, validation constraints, database storage, and Kind testing setup comprehensively. This is spec-only — no implementation code to audit for constitution violations.

**Spec quality assessment:**

✅ **API design is sound** — `oidc` as an optional nullable object follows the same pattern as `labels`/`annotations`; backward compatibility is preserved (null = OIDC disabled, behavior unchanged)

✅ **Validation constraint is explicit** — both `admin_role` + `user_role` must be set together or both empty; scenario covers all three cases (valid RBAC, valid auth-only, invalid partial)

✅ **Change detection is specified** — OIDC changes trigger StatefulSet restart via annotation, consistent with how config field changes are handled today

✅ **Migration table is complete** — all consumers listed: OpenAPI, DB model, presenter, config struct, manifests, validation, reconciler, SDK (Go/Python/TS), Kind Keycloak realm, examples

✅ **JSONB storage is consistent** with the `labels`/`annotations` pattern; migration is additive (nullable column, existing rows get NULL)

✅ **Kind testing is scoped correctly** — `openshell-cli` public client, realm roles, user mappings, and a concrete example overlay; sufficient for local e2e validation

**One note for implementation:**

When the custom `config` TOML field and `oidc` are both set, the spec says OIDC "SHALL be appended to or merged into the custom config." The reconciler implementation should clarify whether it appends (simpler, risk of duplicate sections if user put `[openshell.gateway.oidc]` in their custom config) or does a structured merge. Worth a follow-up in the implementation PR.

**Referenced spec exists:** `gateway-provisioning.spec.md` ✅ | `gateway-rbac-policy.spec.md` ✅ (both in repo)

**Confidence:** High — spec is internally consistent and maps cleanly to existing patterns.

— *Amber*

jsell-rh (changes requested):
🤖 **Amber — Implementation Review (commit `67dfe61`)**

My spec approval stands as a signal of good design. The implementation is largely solid, but there's one meaningful correctness bug and one spec deviation to resolve before this graduates from draft.

---

## Major — `oidcConfigChanged` only checks `issuer`; other field changes won't trigger restart

`oidcConfigChanged` in `internal/gateway/reconciler.go` returns `true` (changed) only when the `[openshell.gateway.oidc]` section is missing or the `issuer` field differs:

```go
// Check if issuer value matches
if !strings.Contains(toml, fmt.Sprintf("issuer   = %q", desired.Issuer)) {
    return true, nil
}
return false, nil
```

This means changes to `audience`, `admin_role`, `user_role`, `roles_claim`, `jwks_ttl`, or `scopes_claim` will **not** trigger a pod restart. An operator patching `admin_role` from `openshell-admin` to `openshell-superadmin` will see the API accept the change, but the running gateway will silently continue enforcing the old role until manually restarted.

**Recommended fix:** Fold OIDC change detection into the existing `gatewayConfigTomlChanged` call. `ApplyConfigOverrides` already injects the OIDC section into the ConfigMap before the TOML comparison, so if you run the manifest overrides first and then call `gatewayConfigTomlChanged`, OIDC changes are detected automatically — no separate `oidcConfigChanged` function needed. The current separate function also duplicates the ConfigMap fetch, adding an unnecessary API call per reconcile loop.

---

## Minor — OIDC injection skipped when custom `config` is set (spec deviation)

In `manifests.go`:
```go
if kind == "ConfigMap" && ... && config.Config == "" {
```

The spec (`gateway-oidc.spec.md`) states:

> When OIDC is enabled AND a custom `config` field is set, the custom config SHALL be used as the base `gateway.toml`, and the `[openshell.gateway.oidc]` section SHALL be appended to or merged into the custom config.

The implementation silently skips OIDC injection when a custom config is present. Users who combine custom TOML with OIDC won't get OIDC applied. Either implement the append (check if the section is already present, append if not) or update the spec to document this as an explicit limitation. Either is acceptable; the current silent skip is not.

---

## Minor — `examples/overlays/tenant-c/gateway.yaml` uses `0.0.80`

The new example file references the old image tag. PR #364 is bumping everything to `0.0.83`. When this PR rebases after #364 merges, run `make update-openshell VERSION=v0.0.83`.

---

## What's good (no action needed)

✅ `unstructured.NestedMap` with `found` check — correct pattern throughout  
✅ `fmt.Errorf("context: %w", err)` error wrapping — consistent  
✅ OIDC validation in `validation.go` — URL prefix check and role-pairing constraint are clean  
✅ JSON marshal/unmarshal in `presenter.go` — matches the labels/annotations JSONB pattern  
✅ SDK types mirrored correctly across Go/Python/TypeScript  
✅ Keycloak realm JSON — audience mapper, realm-roles mapper, user→role assignments are all correct  
✅ `oidcFromResource` in CLI guards on `issuer == ""` before returning a non-nil struct  
✅ Migration is additive (nullable column, existing rows get NULL) — no breaking change  
✅ No `panic()`, no token logging, no security surface regression  
✅ Tenant-c overlay added to `OPENSHELL_TENANTS` in Makefile

**Confidence:** High on the diagnosis. The `oidcConfigChanged` fix is a small refactor — folding into the existing TOML comparison rather than maintaining a parallel check.

— *Amber*

components/ambient-control-plane/internal/gateway/reconciler.go
```diff
@@ -241,6 +252,41 @@ func gatewayConfigTomlChanged(ctx context.Context, dynamicClient dynamic.Interfa
...
+	if !hasOidcSection {
+		return true, nil
+	}
+
+	// Check if issuer value matches
+	if !strings.Contains(toml, fmt.Sprintf("issuer   = %q", desired.Issuer)) {
+		return true, nil
+	}
+
+	return false, nil
```
squizzi: `oidcConfigChanged` only compares the `issuer` field — changes to `audience`, `roles_claim`, `admin_role`, `user_role`, `jwks_ttl`, or `scopes_claim` won't be detected. The ConfigMap will be updated by `ApplyConfigOverrides` with the new values, but `needsRestart` stays false (assuming nothing else changed), so gateway pods keep running with the stale config until an unrelated change triggers a restart.

Consider generating the full desired OIDC TOML snippet here and doing a single substring comparison, similar to how `gatewayConfigTomlChanged` compares the entire config string.

components/ambient-control-plane/internal/gateway/manifests.go
```diff
@@ -136,6 +136,51 @@ func ApplyConfigOverrides(obj *unstructured.Unstructured, config GatewayConfig)
 		}
 	}
 
+	// Inject OIDC configuration into gateway.toml (only when no custom config override)
+	if kind == "ConfigMap" && obj.GetName() == "openshell-gateway-config" && config.Oidc != nil && config.Oidc.Issuer != "" && config.Config == "" {
```
squizzi: The `config.Config == ""` guard skips OIDC injection when a custom config is provided. The spec added in this PR explicitly requires OIDC to be merged into custom configs:

> **Scenario: OIDC section injected into custom gateway.toml** — the custom config SHALL be used as the base and the `[openshell.gateway.oidc]` section SHALL be appended.

If this is intentionally deferred, the spec scenario should be removed or marked as not-yet-implemented to avoid a contract mismatch between the spec and code.

squizzi (commented):
Looked over the spec, lgtm, I ran a different reviewer agent over the code as well and found a couple issues which are in the comments below.

jsell-rh (changes requested):
🤖 **Amber — Follow-up Review on `fac2bdd`**

New commit adds OIDC kind CLI registration via `setup-gateway-cli.sh` — that logic looks correct and is a nice addition. However, the two blockers from my previous review remain unaddressed in the current HEAD.

---

## Still Blocking

**1. `oidcConfigChanged` only compares `issuer` (reconciler.go ~L270)**

No change from `67dfe61`. The function still returns `false` (no restart needed) when `audience`, `admin_role`, `user_role`, `roles_claim`, `jwks_ttl`, or `scopes_claim` change. squizzi's inline comment (`PRRT_kwDOS2zRNs6Q3Klp`) identifies the same issue.

The fix I recommended in my last review still stands: `ApplyConfigOverrides` already writes the full OIDC section into the ConfigMap before any change-detection runs. You can fold OIDC detection into `gatewayConfigTomlChanged` by running overrides first and comparing the full rendered TOML, eliminating the partial `oidcConfigChanged` function entirely.

**2. OIDC injection skipped when `config.Config != ""` (manifests.go ~L140)**

Also unchanged. squizzi's inline comment (`PRRT_kwDOS2zRNs6Q3Ktr`) captures this too. The spec says OIDC SHALL be merged into a custom config when both are set; the code silently skips it. Either implement the append (check for existing `[openshell.gateway.oidc]` section, append if absent) or explicitly document in the spec and code that this combination is unsupported.

---

## New commit review (`fac2bdd` — `setup-gateway-cli.sh` + `Makefile`)

✅ OIDC issuer detection from ConfigMap TOML is correct — `grep '[openshell.gateway.oidc]'` + `sed` extraction  
✅ Keycloak issuer URL rewrite (`svc.cluster.local` → `localhost:$KEYCLOAK_PORT`) handles the dual-issuer pattern correctly  
✅ `--gateway-insecure` flag appropriate for kind local dev (TLS not enforced in local Keycloak flows)  
✅ Graceful fallback to mTLS path when no OIDC section detected  
✅ `KEYCLOAK_PORT` passed through Makefile correctly  

One minor note: if `OIDC_ISSUER` is extracted but `KEYCLOAK_PORT` is unset (e.g., someone calls the script directly without going through `make`), the issuer rewrite is skipped and the raw `svc.cluster.local` URL is passed to `openshell gateway add`. That will silently fail at CLI login time. Worth a `[ -z "$KEYCLOAK_PORT" ] && echo "WARN: KEYCLOAK_PORT unset, OIDC issuer may not be reachable from localhost"` guard, but this is a minor dev-UX issue, not a blocker.

---

**Summary:** Two blockers from the prior review still need resolution before this graduates from draft. The new commit is clean and doesn't introduce new issues.

— *Amber*

jsell-rh (changes requested):
🤖 **Amber — Follow-up Review on `9e6b6a9`**

Reviewing the new commit against my two prior blockers.

---

## Blocker 2 — Resolved ✅

**OIDC injection skipped when `config.Config != ""`**

The spec in `gateway-oidc.spec.md` now explicitly documents this as intentional ("Scenario: Custom config overrides bypass OIDC injection — the custom `config` SHALL be used verbatim... the user is responsible for including OIDC settings directly in the custom TOML"). The spec and code are now in agreement. squizzi's thread `PRRT_kwDOS2zRNs6Q3Ktr` is resolved by this spec update — the contract mismatch that made this a blocker no longer exists.

---

## Blocker 1 — Still Open ❌

**`oidcConfigChanged` only compares `issuer`** (`reconciler.go` L283)

`reconciler.go` was not touched in `9e6b6a9`. The function still returns `false` (no restart needed) when `audience`, `admin_role`, `user_role`, `roles_claim`, `jwks_ttl`, or `scopes_claim` change. This is squizzi's open thread `PRRT_kwDOS2zRNs6Q3Klp` and my prior review's Major blocker — both still unresolved.

The same recommended fix applies: generate the full desired OIDC TOML snippet and compare it as a whole string, rather than checking only `issuer`. This mirrors how `gatewayConfigTomlChanged` compares the entire config string and avoids maintaining a parallel field-by-field comparison that will drift as the OIDC schema grows.

---

## New commit review (`9e6b6a9` content)

✅ `client_ca_path` removal when OIDC enabled — correct, mTLS and Bearer token auth are mutually exclusive  
✅ Fixed-port kind cluster — cleaner than hash-derived offsets for local dev  
✅ Keycloak issuer URL fixed to use service DNS with port `11880` — consistent with the `/etc/hosts` entry added in CI  
✅ `SKIP_TENANT_SETUP` support and sub mapper — correct additions  
✅ `setup-gateway-cli.sh` refactor to print commands instead of auto-registering — right call, avoids OIDC browser login in automation  
✅ `manifests_test.go` updated to cover the mTLS disable path  

One minor nit: `examples/overlays/tenant-c/gateway.yaml` image tag — worth verifying this was updated to match the current openshell release when you rebase (prior round flagged `0.0.80`).

---

**Summary:** One blocker remaining (partial OIDC change detection in `oidcConfigChanged`). Everything else in this commit is clean.

— *Amber*

components/ambient-control-plane/internal/gateway/reconciler.go
bsquizz: Fixed. Removed `oidcConfigChanged` entirely and replaced it with `renderedConfigChanged`, which builds the fully-rendered `gateway.toml` by running `ApplyConfigOverrides` on a copy of the current ConfigMap and comparing the result against the live TOML. This catches changes to any OIDC field (audience, roles_claim, admin_role, user_role, jwks_ttl, scopes_claim) — not just `issuer`. It also eliminates the duplicate ConfigMap fetch that the separate function was doing.

components/ambient-control-plane/internal/gateway/manifests.go
bsquizz: Fixed. Removed the `config.Config == ""` guard so OIDC injection now runs for both default and custom configs, matching the spec. Added a `strings.Contains(toml, "[openshell.gateway.oidc]")` check to avoid duplicating the section if the custom config already defines one.

jsell-rh (approved):
🤖 **Amber — Follow-up Review on `92678b96`**

All three blockers from my previous review are resolved. Approving.

---

## Blockers Resolved ✅

**1. `oidcConfigChanged` only compared `issuer`**

Replaced entirely with `renderedConfigChanged`, which builds the fully-rendered `gateway.toml` by running `ApplyConfigOverrides` on a copy of the current ConfigMap and diffing the result against the live TOML. Changes to any OIDC field (`audience`, `admin_role`, `user_role`, `roles_claim`, `jwks_ttl`, `scopes_claim`) now trigger a pod restart. The approach also subsumes the old `gatewayConfigTomlChanged` call (custom config + OIDC + serverDnsNames all detected in one unified comparison), eliminating the duplicate ConfigMap fetch. This is cleaner than what I suggested. ✅

**2. OIDC injection skipped when `config.Config != ""`**

`config.Config == ""` guard removed. OIDC is now injected for both default and custom configs, with a `strings.Contains(toml, "[openshell.gateway.oidc]")` guard to avoid duplicating the section if the custom config already defines one. Spec and code are in agreement. ✅

**3. Migration `AutoMigrate` fails with `PreferSimpleProtocol`**

Replaced with explicit `ALTER TABLE gateways ADD COLUMN IF NOT EXISTS oidc JSONB` / `DROP COLUMN IF EXISTS`. ✅

---

## Minor (no action required)

`gatewayConfigTomlChanged` is now dead code — its only call site was replaced by `renderedConfigChanged`. Safe to remove in a cleanup commit, but not a blocker.

---

## CI Status (commit `92678b96`)

All critical gates green: lint, unit tests, CodeQL, API server integration tests, all component builds (amd64 + arm64). `test-local-dev-simulation` is still in progress, which is expected at this stage.

---

When you promote from draft, the outstanding threads from squizzi are both outdated (both issues are addressed in this commit as confirmed in your inline replies). Worth resolving them before requesting final human review.

**Confidence:** High — the `renderedConfigChanged` approach is correct and complete.

— *Amber*

jsell-rh (approved):
🤖 **Amber — Follow-up on `34f8f1b` + `36cac18`**

Two small SDK cleanup commits since my last approval. Both are clean.

**`34f8f1b` — GatewayOidc companion type file**

Correct pattern. `GatewayOidc` was hand-edited into the generated `gateway.go`, causing regeneration drift. Extracting to `gateway_oidc.go` follows the same convention as `sandbox_template.go` and `payload.go` — non-generated companion types that sit alongside generated files without being overwritten. The regenerated `gateway.go` now has the `Oidc` and `ProjectID` builder methods in alphabetical order (minor reordering, no logic change). ✅

**`36cac18` — JwksTTL → JwksTtl field casing**

One-line fix to match the generator's `toGoName` output. `TTL` is not in the acronym map, so the generator produces `Ttl`, not `TTL`. Having the companion type disagree with what the generator would produce for that field would cause a drift check failure. Fix is correct. ✅

**CI** (on `353762d` — merge from main):
- Lint, Unit Tests, API Server Integration Tests, CLI Build, all component builds (amd64 + arm64): success ✅
- CodeQL: success ✅
- `test-local-dev-simulation`: in progress (expected at this stage) ✅

Prior approval stands. All blockers from prior rounds remain resolved.

— *Amber*

jsell-rh (commented):
🤖 **Amber — Follow-up on `9ba1920` (CI footprint reduction)**

New commit is CI/test infrastructure only — all production code from prior commits unchanged. My existing approval stands.

**New commit review (`9ba1920`):**

✅ `OPENSHELL_TENANTS="tenant-a"` in `test-local-dev.yml` — correct. Single tenant is all the e2e test exercises; removing 4 idle StatefulSets saves real cluster headroom without changing test coverage.

✅ `_cleanup_sandboxes()` helper — `|| true` makes it non-fatal, `SKIP_CLEANUP` guard is respected. The three mid-section cleanup calls are placed at correct logical boundaries (after section 11, after section 12, after network policy tests). No resource leak from the helper itself.

✅ Final cleanup refactor (for loop over session IDs) — functionally equivalent to the prior four `if [ -n ]` blocks, cleaner.

**CI status — needs investigation:**

`test-local-dev-simulation` is **failing** on this commit. All other gates (lint, unit tests, API integration, builds, CodeQL) are green.

The changes in this commit should reduce CI resource pressure, not increase it — the failure is likely pre-existing infra instability or a race in the e2e test that the sandbox cleanup exposed (e.g., deleting sandboxes mid-section while another operation is in flight). Worth checking the job logs for the failure message before re-triggering.

One thing to verify: `_cleanup_sandboxes` uses `$TENANT` which must be set at call sites. Confirm `$TENANT` is always initialized before the first `_cleanup_sandboxes` call in the script (section 11 → 12 boundary).

— *Amber*

jsell-rh (approved):
🤖 **Amber — Follow-up on `edbd76d` + `80dcc357`**

Two new commits since my last review. Both are clean — prior approval stands.

---

## `80dcc357` — `image_pull_policy` idempotency fix (production code change)

**Root cause was real and well-diagnosed.** `ApplyConfigOverrides` was non-idempotent: every reconcile cycle (30s) unconditionally inserted an `image_pull_policy` line, causing `renderedConfigChanged` to always return true and restart the StatefulSet in a tight loop.

Fix is correct:
```go
if !strings.Contains(toml, "image_pull_policy") {
    // insert logic
}
```
✅ Simple substring guard makes insertion idempotent — if the line is already present, skip. This is the standard TOML injection pattern.
✅ The `data["gateway.toml"]` assignment is now correctly inside the `if` block — when skipped, the existing toml is left unchanged (no data update call either, which avoids a no-op ConfigMap patch).
✅ Constitution checklist: no `panic()`, no token logging, proper `unstructured.SetNestedMap` error handling preserved from surrounding code.

One minor observation (no action needed): `strings.Contains(toml, "image_pull_policy")` would also match a user who set `image_pull_policy` in their custom `config` field, skipping the Kind-only `IfNotPresent` injection. That's actually the correct behavior (user config takes precedence). Happy accident, not a bug.

---

## `edbd76d` — CI gateway readiness waits

CI infrastructure only, no production code changed.

✅ TLS secret wait loop (60×5s) before `setup-gateway-cli.sh` — correct fix for the certgen race.
✅ `kubectl wait --for=condition=ready pod ... --timeout=120s || echo "Warning: ..."` — non-fatal, correct.
✅ Early gateway registration after section 8 health check with `skip` fallback.
✅ Pre-network-test readiness wait (30×2s) + port-forward refresh on restart.
✅ Brief readiness wait between locked-down/permissive phases (15×2s).

The `_delete_session` helper from `80dcc357` combined with these waits cleanly addresses all three race conditions described in the commit message.

---

## Summary

All blockers from prior rounds remain resolved. The infinite-restart loop fix is an important correctness improvement that would have been painful to debug in a deployed cluster. CI gates (lint, unit tests, CLI, UI, builds, CodeQL, SDK drift) are all ✅ on the latest commit.

Prior approval stands. When ready to promote from draft, the stale review threads from earlier rounds can be resolved before requesting final human review.

— *Amber*

jsell-rh (approved):
🤖 **Amber — Follow-up on `f4f5cf13` + `da1d92f9`**

Two new CI-only commits since my last approval. Both are clean — prior approval stands.

---

## `f4f5cf13` — Restart port-forwards before RBAC and scheduled session e2e tests

✅ Root cause is real: port-forwards established in the gateway e2e step can die during sandbox cleanup, leaving the RBAC and scheduled-session tests trying to hit ports that have no listener.

✅ The fix (background port-forward + `sleep 3` grace period) is the same pattern already used in the gateway e2e step — consistent with how the workflow handles this elsewhere.

✅ No production code touched.

## `da1d92f9` — Kill stale port-forwards before restarting

✅ Correct follow-up. Without `fuser -k` first, the new `kubectl port-forward` would fail with `bind: address already in use` if the prior process was still alive. `|| true` makes it non-fatal when no process holds the port (the common case in CI).

✅ `fuser -k <port>/tcp` is available on the ubuntu-24.04 runner (part of `psmisc`). No dependency concern.

✅ CI-only, no production code touched.

---

**Summary:** Both commits address a real reliability regression identified in the previous round. The pattern is consistent throughout the workflow. Prior approval stands.

— *Amber*

jsell-rh (approved):
🤖 **Amber — Follow-up on `7228be2b` (JWKS retry loop)**

CI-only change — no production code touched. Prior approval stands.

✅ Root cause is correct: Keycloak's OIDC discovery endpoint has a longer initialization time than the pod readiness probe after `KC_HOSTNAME` is patched and the pod restarts. A single-shot `curl` on a fresh rollout will consistently lose this race in CI.

✅ Fix is correct: 30×2s retry loop (60s total budget) with `[[ -n "$KC_JWKS" && "$KC_JWKS" != "null" ]] && break` — matches the pattern used elsewhere in the e2e suite for service readiness. The subsequent FATAL guard is unchanged, so a genuine Keycloak outage still fails fast after the retry budget expires.

✅ No production code changed; all previously-reviewed blockers remain resolved.

— *Amber*

jsell-rh:
🤖 **Amber — CI Status Note (commit `7228be2b`)**

`test-local-dev-simulation` is still failing on the latest commit. All other gates are green (lint, unit tests, API integration, all builds amd64/arm64, CodeQL).

This is the same e2e flakiness we've been tracking. Given the JWKS retry fix in `7228be2b` was specifically targeting one failure mode, the current failure may be a different race or a pre-existing infra issue. Worth checking the job logs to see if it's the JWKS fetch specifically or something else in the test suite.

My code review approval stands — this is a CI reliability issue, not a correctness issue with the implementation.

— *Amber*

jsell-rh (approved):
🤖 **Amber — Follow-up on `6042e6c` (Keycloak port-forward port fix)**

CI-only change — one line in `tests/e2e/rbac_e2e_test.sh`. No production code touched.

✅ Root cause is real: `kubectl port-forward svc/...` requires the **service port**, not the container port. Keycloak's service port is `11880`; the port-forward was incorrectly targeting `8080` (container port), causing the re-established forward to fail silently after Keycloak rollout.

✅ Fix is minimal and correct — single-line change to match the service port convention already used everywhere else in this PR.

Prior approval stands. All blockers from prior rounds remain resolved.

— *Amber*

PR #364, opened 2026-07-14
chore(deps): Update OpenShell to v0.0.83 and add one-shot dependency update tooling

maknop:
# Summary

- Update OpenShell from v0.0.80 to v0.0.83 — vendored protos, regenerated gRPC stubs, and bumped all gateway/supervisor image tags
- Confirm agent-sandbox is already at v0.5.1 (latest) with v1beta1 API — no alpha references remain
- Fix vendor-proto.sh macOS compatibility (herestring/heredoc conflict, GNU mktemp --suffix, missing import path rewriting)
- Add one-shot update scripts and Makefile targets: make update-openshell VERSION=<tag> and make update-agent-sandbox VERSION=<tag>

# Detail

## OpenShell v0.0.80 → v0.0.83

Proto definitions are unchanged between versions; the update picks up bug fixes in the gateway and supervisor (cert gen cross-device rename fix, fail-closed credential rewriting, PVC subPath support, sidecar supervisor topology). Updated across:
- Vendored proto sources + generated Go gRPC stubs
- Gateway image defaults in internal/gateway/reconciler.go and internal/reconciler/gateway_reconciler.go
- Supervisor image in manifests/gateway/configmap.yaml
- All example gateway YAMLs under examples/

## vendor-proto.sh fixes

Three macOS-incompatible patterns that caused the script to fail:
1. Simultaneous <<< herestring and <<HEREDOC both targeting stdin — replaced with pipe
2. mktemp --suffix=.py (GNU-only) — replaced with mktemp -t prefix
3. Upstream protos use bare imports (import "datamodel.proto") but local layout nests them (openshell/datamodel/v1/datamodel.proto) — added a post-fetch import rewriting pass derived from the file table

## One-shot update tooling

`make update-openshell VERSION=v0.0.84`: Vendors protos, regenerates stubs, updates gateway + supervisor image tags across Go source, configmap, and examples.
`make update-agent-sandbox VERSION=v0.6.0`: Validates the release exists, updates version in Makefile, setup script, README, and provisioning spec.

---


jsell-rh (approved):
🤖 **Amber Review — Approved**

**Summary:** Clean OpenShell v0.0.80 → v0.0.83 bump with macOS-compatible vendor tooling fixes and new one-shot update scripts. Proto vendoring is correct, image tags are consistent across all files, and the generated gRPC stubs reflect the new upstream SHA. Good tooling investment.

**What I checked:**

✅ **Proto vendoring correct** — vendored commit `e3d26dd3` matches all `.pb.go` header comments; `VENDOR.md` updated; import reordering in `openshell.proto` (`go_package` moved before imports) is the generated output change from the new upstream commit.

✅ **Image tags consistent** — gateway (`0.0.83`) and supervisor (`0.0.83`) updated in `reconciler.go`, `gateway_reconciler.go`, `configmap.yaml`, and all example YAMLs.

✅ **macOS mktemp fix is correct** — `--suffix=.py` is Linux-only; `$(mktemp -t vendor_proto).py` works cross-platform.

✅ **Import path rewriting logic** — `vendor-proto.sh` now rewrites bare upstream imports (`import "sandbox.proto"`) to local nested paths (`openshell/sandbox/v1/sandbox.proto`). The sed expression builds and applies correctly.

✅ **Update scripts** — `update-openshell.sh` and `update-agent-sandbox.sh` are well-structured: idempotency check (skip if already at target), version detection from source-of-truth files, `set -euo pipefail` guards, clear progress output.

**Two nits (non-blocking, fix before marking ready):**

1. **Minor temp file leak** — `PYSCRIPT="$(mktemp -t vendor_proto).py"` creates a temp file `/tmp/vendor_protoXXXXXX` via `mktemp`, then appends `.py` to that path string. The `trap` cleans up `$PYSCRIPT` (the `.py` path) but not the original empty file created by `mktemp`. It's a 0-byte file and gets cleaned on OS reboot, but it's avoidable. The script doesn't require the `.py` extension for `buf protoc --plugin` — the file just needs to contain valid Python.

2. **Import rewrite condition always true** — The guard `[[ "$upstream_basename" != "$local_sub" ]]` is always true when `local_sub` has directory components (which it always does — e.g., `openshell/sandbox/v1/sandbox.proto`). The condition body is always entered. Functionally correct (sed substitution is idempotent), but the logic is misleading — a reader might think the condition gates something meaningful.

**One merge-order awareness note:** `examples/overlays/tenant-c/gateway.yaml` (added by PR #363) uses `0.0.80`. When either PR merges and the other rebases, `make update-openshell VERSION=v0.0.83` will catch it since the script searches `examples/**/*.yaml`. Not a blocker.

**Constitution compliance:** No production Go logic changed — only vendored generated code and tooling scripts. No `panic()`, no `any` types, no security surface changes.

**Confidence:** High — straightforward version bump with good cross-platform tooling fixes.

— *Amber*

PR #366, opened 2026-07-15
spec: acp-control-plane using OIDC authentication to gateways

bsquizz:
## Summary

The purpose of this PR is to enable ACP's control-plane to authenticate with a gateway that has been set up in "OIDC auth mode" as opposed to the way things currently work (using 'mtls auth mode'), so that it can manage that gateway.

- Adds `specs/platform/gateway-oidc-cp-auth.spec.md` defining how the control plane authenticates to OIDC-enabled gateways using OAuth2 `client_credentials` grant instead of mTLS
- Builds on PR #363 which configures the gateway side of OIDC; this spec covers the control plane side
- Covers per-gateway auth strategy selection, token acquisition/refresh, server-only TLS, credential Secret schema, and Kind Keycloak client setup

## Test plan

- [ ] Review spec requirements and scenarios for completeness
- [ ] Verify alignment with PR #363's gateway OIDC configuration
- [ ] Confirm backward compatibility claims against current `GatewayClient` implementation

### DIFF
--- a/specs/platform/gateway-oidc-cp-auth.spec.md
+++ b/specs/platform/gateway-oidc-cp-auth.spec.md
@@ -1,0 +1,317 @@
+# Control Plane OIDC Authentication to Gateways
+
+**Date:** 2026-07-14
+**Status:** Design
+**Depends on:** PR #363 (`feat(gateway): optional OIDC authentication for OpenShell gateways`)
+**Related:** `gateway-oidc.spec.md` — gateway-side OIDC configuration; `gateway-provisioning.spec.md` — gateway lifecycle; `sso-authentication.spec.md` — platform SSO model; `control-plane.spec.md` — CP architecture
+
+---
+
+## Purpose
+
+When a Gateway resource has OIDC enabled (non-empty `oidc.issuer`), the control plane's gRPC connection to that gateway SHALL authenticate using an OIDC bearer token obtained via the OAuth2 `client_credentials` grant, instead of mTLS client certificates.
+
+PR #363 configures the gateway side: OIDC-enabled gateways validate `Authorization: Bearer <JWT>` headers and disable mTLS client certificate verification (`client_ca_path` is removed). However, the control plane currently authenticates to all gateways using a fixed strategy — mTLS transport credentials resolved via `TLSResolver` plus a Kubernetes ServiceAccount token injected as gRPC metadata by `authContext()`. Neither credential is valid against an OIDC-enabled gateway: the gateway no longer checks client certificates, and the SA token is not a JWT issued by the gateway's configured OIDC issuer.
+
+This specification closes that gap. When the control plane connects to an OIDC-enabled gateway, it SHALL:
+
+1. Use server-only TLS (verify the gateway's server certificate, but do not present a client certificate)
+2. Obtain a JWT from the gateway's OIDC issuer via the `client_credentials` grant
+3. Inject the JWT as `Authorization: Bearer <token>` in gRPC metadata
+4. Refresh the token transparently before expiry so that gRPC connectivity is never interrupted
+
+The OIDC client credentials are sourced from a Kubernetes Secret named `acp-control-plane-oidc` in the control plane's runtime namespace, mounted via `valueFrom`/`secretKeyRef` on the control plane Deployment. In the Kind test environment, a dedicated Keycloak confidential client provides these credentials. In production, the OIDC provider is whatever issuer URL is configured on the Gateway resource.
+
+---
+
+## Requirements
+
+### Requirement: Per-Gateway Auth Strategy Selection
+
+The control plane SHALL select its authentication strategy for each gateway based on whether the Gateway resource has OIDC enabled. The auth strategy determines both the transport credentials (mTLS vs. server-only TLS) and the application-layer token (SA token vs. OIDC JWT).
+
+| Gateway OIDC state | Transport | Application-layer auth |
+|---|---|---|
+| OIDC disabled (default) | mTLS (client cert from TLS Secret) | SA token from `saTokenPath` |
+| OIDC enabled (`oidc.issuer` non-empty) | Server-only TLS (verify server cert, no client cert) | OIDC JWT from `client_credentials` grant |
+
+The `GatewayClient` SHALL expose a method to register OIDC configuration per namespace (e.g., `SetOIDCConfig(namespace string, oidcCfg *OidcConfig)`). The reconciler SHALL call this method when it processes a Gateway with OIDC enabled. When `getOrCreateConn()` creates a connection for a namespace, it SHALL check the registered OIDC state to select the correct transport and auth strategy.
+
+If the OIDC state of a gateway changes (enabled → disabled or vice versa), the reconciler SHALL call `SetOIDCConfig` with the updated state, which SHALL evict the cached gRPC connection for that namespace so the next call re-establishes it with the correct auth strategy.
+
+#### Scenario: Control plane connects to an OIDC-enabled gateway
+
+- GIVEN a Gateway resource in namespace `tenant-a` with `oidc.issuer = "https://keycloak.example.com/realms/ambient-code"`
+- AND the `acp-control-plane-oidc` Secret exists with valid client credentials
+- WHEN the control plane makes a gRPC call to the gateway (e.g., `CreateSandbox`)
+- THEN the gRPC connection SHALL use server-only TLS (no client certificate presented)
+- AND the gRPC metadata SHALL contain `authorization: Bearer <jwt>` where the JWT is obtained from the gateway's OIDC issuer via `client_credentials` grant
+- AND the gateway SHALL accept the request
+
+#### Scenario: Control plane connects to a non-OIDC gateway
+
+- GIVEN a Gateway resource in namespace `tenant-b` with no `oidc` configuration
+- WHEN the control plane makes a gRPC call to the gateway
+- THEN the gRPC connection SHALL use mTLS (client certificate from TLS Secret)
+- AND the gRPC metadata SHALL contain `authorization: Bearer <sa-token>` from the SA token file
+- AND behavior SHALL be identical to the current implementation
+
+#### Scenario: Gateway OIDC state changes
+
+- GIVEN a cached gRPC connection to a gateway in namespace `tenant-a` using mTLS
+- WHEN the Gateway resource is updated to enable OIDC
+- AND the GatewayReconciler processes the updated Gateway
+- THEN the reconciler SHALL call `GatewayClient.SetOIDCConfig("tenant-a", oidcCfg)` with the new OIDC configuration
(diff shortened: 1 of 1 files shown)


jsell-rh (commented):
🤖 **Amber — Spec Review: PR #366 (gateway-oidc-cp-auth.spec.md)**

**Summary:** Solid, implementation-grounded spec. I validated it against the actual `gateway_client.go`, `tls_resolver.go`, `token_provider.go`, and `config.go`. Most design decisions are well-reasoned and consistent with existing patterns. A few issues worth addressing before implementation starts.

---

### Issues

**[Major] Audience injection not supported by current `OIDCTokenProvider`**

The spec says: _"The `oidc.audience` from the Gateway resource SHALL be included as a token request parameter"_, and separately: _"Reuse `OIDCTokenProvider` as-is for gateway token acquisition."_

These two requirements are in tension. The current `OIDCTokenProvider` (`internal/auth/token_provider.go:40-49`) constructs a `clientcredentials.Config` with only `ClientID`, `ClientSecret`, and `TokenURL` — no `EndpointParams`. To send `audience=openshell-cli` in the `client_credentials` POST, you'd need:

```go
cfg: clientcredentials.Config{
    ...
    EndpointParams: url.Values{
        "audience": {gatewayAudience},
    },
},
```

Since `audience` is per-gateway (different gateways can have different OIDC issuers and audiences), `OIDCTokenProvider` can't be instantiated once and reused for all gateways. You'd need one instance per unique `(tokenURL, clientID, audience)` tuple, or modify the provider to accept `audience` as a `Token(ctx, audience)` parameter. The spec should explicitly address this — right now the migration table says "Reused as-is" which will mislead implementors.

**Suggested resolution:** State that a new `OIDCTokenProvider` is instantiated per gateway namespace (caching by namespace), or add an `audience` parameter to the provider. Either works — just needs to be explicit.

---

**[Minor] Dependency on unmerged PR #363 and #345 not flagged as implementation prerequisite**

The spec's migration table lists `ExecSandboxStreaming` as returning `(<-chan error, error)`. That signature change is in PR #345 (not yet merged). The current `gateway_client.go:290` returns `error`. The spec should note this is an assumed prerequisite, not current state — otherwise implementors may be confused about the baseline.

Similarly, `gateway-oidc.spec.md` (referenced in the Related section) doesn't exist in main yet — it ships with PR #363. This is fine for a spec that explicitly says `Depends on: PR #363`, but a sentence acknowledging this in the Migration section would help.

---

**[Minor] `SetOIDCConfig` API surface — consider making it idiomatic with the existing `evictConn` pattern**

The spec proposes `GatewayClient.SetOIDCConfig(namespace string, oidcCfg *OidcConfig)` which "SHALL evict the cached gRPC connection." Looking at `gateway_client.go:99-112`, `evictConn` already exists and handles the eviction + close. The implementation can simply call `evictConn` from `SetOIDCConfig` — worth calling this out in the spec so implementors don't duplicate that logic.

---

### Strengths

- Auth strategy table (mTLS vs. server-only TLS) matches the real `TLSResolver.CredentialsForNamespace` behavior. The spec correctly identifies that only `ca.crt` needs to be loaded for OIDC connections (not `tls.crt`/`tls.key`).
- The 30-second refresh buffer in `OIDCTokenProvider` (`token_provider.go:55`) exactly matches the spec's requirement — this is already implemented correctly.
- `optional: true` on all `secretKeyRef` entries is the right call — the control plane must start without the Secret in non-OIDC deployments.
- The `acp-control-plane-oidc` Secret name clearly separates gateway OIDC credentials from the API-server OIDC credentials (`OIDC_CLIENT_ID`/`OIDC_CLIENT_SECRET`). This is the correct design.
- Scenarios are complete and cover the failure paths (Secret missing, token refresh failure, stream survival during refresh).

---

**Confidence:** High — reviewed against actual implementation files.

This is a well-structured spec that clearly closes the gap left by PR #363. The audience injection issue is the main thing that needs resolution before implementation; everything else is clarification-level.

— *Amber*

jsell-rh (changes requested):
## 🤖 Amber Review — PR #366: Control Plane OIDC Auth to Gateways

This spec is a clean companion to PR #363 and the overall design intent is sound — using `client_credentials` for the CP→gateway auth hop instead of mTLS when OIDC is enabled is the right call. The secret schema, per-gateway strategy table, and `optional: true` deployment pattern are all correct.

Three issues need resolving before this is ready:

---

### [Major] `OIDCTokenProvider` instance lifecycle is underspecified for multi-issuer environments

The Migration table says:
> `OIDCTokenProvider` (`internal/auth/token_provider.go`) — Reused as-is for gateway token acquisition — same `client_credentials` grant, different credentials

This understates the complexity. The current `OIDCTokenProvider` is presumably instantiated once at startup, pointed at the API server's OIDC issuer. But different gateways can have **different** `oidc.issuer` values (each tenant could use a separate Keycloak realm). You need one `OIDCTokenProvider` instance per distinct `(issuer + client_credentials)` tuple — not one per gateway, but also not one global singleton.

The spec needs to describe instance management:

> The control plane SHALL maintain an `OIDCTokenProviderCache` (map from `token-url+client-id` → `OIDCTokenProvider`). `SetOIDCConfig` looks up or creates a provider for the gateway's `(token-url, client-id, audience)` tuple. Providers are evicted when no gateway references that configuration. This avoids both the N-provider-per-gateway explosion and the single-provider assumption.

Without this, the implementation will either create a new token provider on every gRPC call (cache miss every time, no refresh benefit) or share one provider across gateways with different issuers (wrong token for some gateways).

---

### [Major] "Existing gRPC streams survive token refresh" scenario is architecturally incorrect

From the spec:
> GIVEN an active gRPC streaming call (e.g., `WatchSandbox`) to an OIDC-enabled gateway  
> AND the OIDC token expires during the stream  
> WHEN the stream continues to receive events  
> THEN the stream SHALL NOT be interrupted by token expiry

This is not achievable with standard gRPC metadata-based token injection. gRPC metadata headers are set at the time the **stream is opened** — the gateway sees the token from the initial `WatchSandbox` call and there is no mechanism to refresh it mid-stream without closing and reopening the connection.

The spec should instead say:

> The control plane refreshes tokens proactively (30s before expiry) so that *new* gRPC calls and *new* streams use a fresh token. Existing long-lived streaming calls (e.g., `WatchSandbox`) may be established before a token refresh and will use the token that was current at stream open time. If the gateway closes the stream due to token expiry, the control plane's existing reconnection logic (gRPC backoff with re-dial) will establish a new stream with a fresh token. The 30-second proactive refresh window is sized to be shorter than the stream's expected lifetime between reconnects.

This is a "correct by reconnect" guarantee rather than a "transparent mid-stream refresh" guarantee. The spec should be honest about which it provides — implementers will discover this constraint immediately.

---

### [Minor] CP restart doesn't rebuild OIDC state from Gateway resources

The spec defines `SetOIDCConfig(namespace, *OidcConfig)` as the mechanism for registering per-namespace OIDC state. But on CP restart, `SetOIDCConfig` is never called until the reconciler processes a MODIFIED event. If no events arrive (no Gateway changes since startup), the in-memory OIDC state map is empty and all connections fall back to mTLS — breaking connections to OIDC-enabled gateways.

Add a requirement:
> On startup (or when the GatewayReconciler initializes), the reconciler SHALL perform an initial LIST of all Gateway resources and call `SetOIDCConfig` for each with a non-empty `oidc.issuer`, before the watch loop begins. This ensures OIDC state is rebuilt from current cluster state on every restart.

This is the same reconciler initialization pattern used elsewhere in the CP (listing existing resources before establishing the watch).

---

### What's solid ✅

- Per-gateway auth strategy table (OIDC vs mTLS) — exactly right
- `optional: true` on all `secretKeyRef` entries — no dummy Secrets in non-OIDC deployments
- Separate `acp-control-plane-oidc` Secret from `OIDC_CLIENT_ID`/`OIDC_CLIENT_SECRET` — correct credential isolation
- Server-only TLS for OIDC gateways (verify server cert, no client cert) — matches how OIDC-enabled gateways work per PR #363
- Token URL derivation fallback (`oidc.issuer + /protocol/openid-connect/token`) — right Keycloak convention
- `openshell-admin` role for CP service account — correct (CP manages sandbox lifecycle)
- 5-minute token lifetime + 30-second refresh buffer — matches existing `OIDCTokenProvider` pattern
- Backward compat: non-OIDC gateways unchanged, Secret absence is safe

**Confidence:** High — The OIDCTokenProvider instance scope and stream-survive semantics are implementation constraints that will be hit immediately.

*— Amber*

jsell-rh (commented):
🤖 **Amber Review — PR #366: spec: control-plane OIDC authentication to gateways**

This is a spec-only PR (draft), so no code to check against the Go/TS checklists. The document is well-structured and internally consistent. A few observations:

**Strengths:**
- Per-gateway auth strategy selection is clearly specified with the right invariant: both mTLS path and OIDC path are preserved, controlled by `oidc.issuer` presence.
- `optional: true` on all `secretKeyRef` entries is correct — prevents control plane startup failure when no OIDC gateways exist.
- The 30-second refresh buffer and 5-minute token lifetime are consistent with existing `OIDCTokenProvider` patterns.
- Distinguishing `OIDC_GATEWAY_CLIENT_ID` from `OIDC_CLIENT_ID` (API server auth) is the right call.

**Questions/gaps to resolve before implementation:**

1. **`SetOIDCConfig` race window** — The spec says evicting the cached gRPC connection on OIDC state change causes "the next call" to re-establish with correct auth. But if a streaming call (`WatchSandbox`) is in-flight on that connection, eviction races with the stream. The spec should say whether in-flight streams are drained, interrupted, or left until they naturally close. (The reconciler already handles reconnection, but the spec should be explicit.)

2. **Token refresh in streaming context** — Requirement says "existing gRPC streams SHALL NOT be interrupted by token expiry." How? gRPC stream tokens are set at dial time (for metadata interceptors) or per-call. If the token is set as a per-call credential, long-lived streams will use the token from when the stream opened. If the token is not re-injected into an in-flight stream, it will eventually get `UNAUTHENTICATED` from the gateway as the token expires. The spec needs to address whether: (a) stream auth is re-negotiated via gRPC keep-alive + re-auth, (b) the stream is reconnected when the token expires, or (c) some other mechanism. This is a non-trivial implementation detail.

3. **`DisableAdvisoryLock` global in tests** — Not a spec issue, but note this for the implementation PR: if the Go implementation follows the clusters pattern (PR #355), the test harness should not use a package-level `DisableAdvisoryLock` global for OIDC token provider. Use dependency injection or interface instead.

4. **Kind Secret provisioning** — The spec says the Secret should be created during `kind-up`. What script creates it? `setup-kind-sso.sh` is the natural place, but the spec should call this out explicitly so the implementation PR author doesn't have to infer it.

**Overall:** Good spec. Resolve the streaming-auth question (point 2) before it becomes a thorny implementation surprise. The rest can be addressed during implementation.

This is a draft PR — leaving it as `amber/change-requested` until the streaming question is resolved and PR #363 lands.

— Amber

markturansky:
implemented in https://github.com/openshift-online/agent-control-plane/pull/404

PR #368, opened 2026-07-15
feat: sandbox lifecycle cleanup and stop_on_run_finished

JuanmaBM:
## Summary

- Add `stop_on_run_finished` boolean field to Session across the full stack (OpenAPI, API server + migration, Go/Python/TS SDKs, UI checkbox)
- Remove the gateway-mode guard on PhaseCompleted deprovision so sandboxes are cleaned up on completion regardless of provisioning mode
- Pass `STOP_ON_RUN_FINISHED` env var to gateway sandboxes based on the session field so the runner auto-stops when finished
- Add `ExecExitError` sentinel type to distinguish non-zero exit codes from transient gRPC errors (no retry on real failures)
- Use independent contexts per step in `deprovisionSessionSandbox` to prevent timeout cascades
- Show Live/Inactive indicator in chat based on session phase
- Add e2e tests for long-running and short-running sandbox lifecycle: long-running sessions poll for 2 minutes to confirm the sandbox stays up after the LLM responds; short-running sessions set `stop_on_run_finished=true` and verify the session transitions to Completed and the sandbox is terminated
- Add `--test NAME` flag to the gateway e2e test script to run individual test sections. Sections 1-8 (setup/prerequisites) always run; test sections 9+ are gated by short name: `long_running`, `short_running`, `repo_payload`, `network_policy`. Usage: `./tests/e2e/gateway-e2e-test.sh --test short_running`

Closes https://redhat.atlassian.net/browse/ENGPROD-10186


jsell-rh (changes requested):
## 🤖 Amber Review — PR #368: `feat: sandbox lifecycle cleanup and stop_on_run_finished`

**Overall:** This is a solid full-stack feature that resolves the longstanding FIXME(#223). The `ExecExitError` sentinel type is exactly the right pattern for distinguishing non-zero exits from transient gRPC errors, the `deprovisionSessionSandbox` independent contexts are a clear correctness improvement, and the Live/Inactive indicator is a nice UX touch. The multi-SDK propagation is thorough.

Three issues need addressing before merge: one Shadcn convention violation (blocker per ACP checklist), one behavioral regression risk for existing gateway sessions (Major), and one TypeScript type safety gap (Major).

---

### [Blocker] UI uses raw `<input type="checkbox">` — Shadcn convention violation

`create-session-sheet.tsx:184-192` uses a plain HTML `<input type="checkbox">` and `<label>`. Per ACP conventions and the review checklist: **Shadcn UI components only — no custom buttons, inputs, dialogs.** Replace with Shadcn `<Checkbox>` + `<Label>`:

```tsx
import { Checkbox } from '@/components/ui/checkbox'
import { Label } from '@/components/ui/label'

// ...
<div className="flex items-center gap-2">
  <Checkbox
    id="session-stop-on-finish"
    checked={stopOnRunFinished}
    onCheckedChange={(checked) => setStopOnRunFinished(checked === true)}
  />
  <Label htmlFor="session-stop-on-finish">Stop sandbox on run finished</Label>
</div>
```

---

### [Major] Existing gateway sessions default to `stop_on_run_finished = true` after upgrade — undocumented breaking behavioral change

The migration adds the column as `BOOLEAN` (nullable, no DEFAULT). All existing sessions have `NULL` in the DB. The REST-fetch path in `provisionSessionSandbox`:

```go
stopOnRunFinished := true  // ← default
if freshSession.StopOnRunFinished != nil {
    stopOnRunFinished = *freshSession.StopOnRunFinished
}
```

Pre-PR, the gateway FIXME(#223) meant gateway sessions **never** auto-stopped. Post-merge, all pre-existing sessions (nil field) will silently begin auto-stopping and deprovisioning on completion. This is a behavioral change that operators running the platform need to know about.

Suggested mitigations (pick one):
1. Add `DEFAULT false` to the migration, preserving the pre-PR behavior for existing sessions.
2. Keep nullable but flip the code default: `stopOnRunFinished := false` and set `true` explicitly via the new field.
3. Document the change explicitly in the PR body and ensure ops/release notes reflect it.

I'd lean toward option 1 or 2 — making existing sessions default to the old behavior (no auto-stop) is the least-surprise choice for operators.

---

### [Major] TypeScript SDK: `stop_on_run_finished` is non-optional in `Session` type but nullable from DB

`ts-sdk/src/session.ts`:
```typescript
export type Session = ObjectReference & {
  // ...
  stop_on_run_finished: boolean;  // ← non-optional
```

The DB column is nullable (no DEFAULT), so existing sessions return `null` from the API. A non-optional `boolean` type will fail type-checks or cause silent runtime coercion. Should be:
```typescript
stop_on_run_finished?: boolean | null;
```

The `SessionCreateRequest` and `SessionPatchRequest` correctly use `stop_on_run_finished?: boolean` — the `Session` response type needs to match the actual nullable behavior.

---

### [Minor] Context cancellers not deferred

`kube_reconciler.go` — in both `execAfterReady` (completion block) and `deprovisionSessionSandbox`, context cancellers are called at the end rather than deferred. This is technically safe here (no early returns between creation and cancel), but it's non-idiomatic Go and a maintenance risk if someone adds an early return later:

```go
// Current (fragile):
completeCtx, completeCancel := context.WithTimeout(context.Background(), 10*time.Second)
// ... code ...
completeCancel()

// Preferred:
completeCtx, completeCancel := context.WithTimeout(context.Background(), 10*time.Second)
defer completeCancel()
```

Same for `snapCtx/snapCancel`, `deleteCtx/deleteCancel`, `phaseCtx/phaseCancel` in `deprovisionSessionSandbox`.

---

### What's solid ✅

- `ExecExitError` sentinel type + `errors.As` distinction is exactly right — clean, testable, idiomatic
- Tests for `ExecExitError` (3 cases) and `TestPhaseCompleted_TriggersDeprovision` are appropriate
- Removing the gateway-mode guard on `PhaseCompleted` deprovision resolves FIXME(#223) cleanly
- Independent contexts per step in `deprovisionSessionSandbox` prevents timeout cascades — a real correctness fix
- TODO comment acknowledging the REST-fetch workaround (`// TODO(#223): add stop_on_run_finished to gRPC Session proto`) is appreciated
- `Clone()` correctly propagates `StopOnRunFinished` from source session
- Go/Python/TS SDK coverage is thorough and consistent
- DB migration includes rollback — well done
- `LiveIndicator` receiving `isLive` prop and showing Inactive state is a good UX improvement; always showing when `chatItemCount > 0` (not just when at bottom) is the right call

**Confidence:** High (90%) — The Shadcn and default-value issues are clear convention/correctness gaps. The TypeScript nullability is a direct type-system mismatch.

*— Amber*

jsell-rh (changes requested):
🤖 **Amber Review — PR #368: sandbox lifecycle cleanup + `stop_on_run_finished`**

The core logic here is solid: `ExecExitError` sentinel type, independent contexts in `deprovisionSessionSandbox`, and the full-stack field propagation all look correct. Tests cover the new error type and phase transitions. Three issues need attention before merging.

**Blocking — must fix:**

1. **Shadcn violation in `create-session-sheet.tsx`** — Native `<input type="checkbox">` is used instead of Shadcn's `<Checkbox>` component. CLAUDE.md is unambiguous: "Shadcn UI components only — no custom buttons, inputs, dialogs." The Shadcn Checkbox is a drop-in replacement; this should take ~5 minutes to fix.

2. **TypeScript SDK `Session` type has `stop_on_run_finished: boolean` (non-optional)** — The DB column is `BOOLEAN` (nullable), the OpenAPI spec defines it as optional (`*bool` in Go), and the Python SDK uses `Optional[bool]`. The TS type should be `stop_on_run_finished?: boolean` to match. Old sessions without this field set will get `undefined`, not a `boolean`, causing potential type errors in consumers.

**Non-blocking — nice to fix:**

3. **Context cancel not deferred in `execAfterReady`** — The `completeCancel()` on the UpdateStatus context should be `defer completeCancel()` to ensure it fires on all exit paths. Currently it only fires if the `if !stopOnRunFinished` branch doesn't return first. Low risk since `stopOnRunFinished` guard is right above, but `defer` is the correct idiom.

<details>
<summary>Inline details</summary>

**`create-session-sheet.tsx` ~line 186** — replace `<input type="checkbox">` with:
```tsx
import { Checkbox } from '@/components/ui/checkbox'
// ...
<div className="flex items-center gap-2">
  <Checkbox
    id="session-stop-on-finish"
    checked={stopOnRunFinished}
    onCheckedChange={(checked) => setStopOnRunFinished(checked === true)}
  />
  <label htmlFor="session-stop-on-finish" className="text-sm font-medium">Stop sandbox on run finished</label>
</div>
```

**`ts-sdk/src/session.ts`** — change:
```ts
stop_on_run_finished: boolean;  // ← wrong, non-optional
```
to:
```ts
stop_on_run_finished?: boolean;  // ← matches OpenAPI + DB nullable
```

**`kube_reconciler.go` in `execAfterReady`** — change:
```go
completeCtx, completeCancel := context.WithTimeout(context.Background(), 10*time.Second)
// → add:
defer completeCancel()
```

</details>

**Confidence:** High (95%) — the backend and control-plane changes are correct; these are frontend/SDK consistency issues.

— Amber

jsell-rh (changes requested):
🤖 **Amber Follow-up — PR #368 (commit `d6fdb4e`)**

The new commit is a solid fix: detecting `codes.NotFound` in the exec retry loop and exiting cleanly prevents the spurious `Failed` marking when a session is manually stopped. The session-phase fetch for diagnostic logging is a nice touch. ✅

**Two prior blockers remain open.** Neither was addressed in the new commit.

---

### [Blocker — still open] Shadcn checkbox violation in `create-session-sheet.tsx`

The raw `<input type="checkbox">` and `<label>` are still there. CLAUDE.md is unambiguous: "Shadcn UI components only — no custom buttons, inputs, dialogs." This has been flagged in two prior reviews. Replace with:

```tsx
import { Checkbox } from '@/components/ui/checkbox'
import { Label } from '@/components/ui/label'

<div className="flex items-center gap-2">
  <Checkbox
    id="session-stop-on-finish"
    checked={stopOnRunFinished}
    onCheckedChange={(checked) => setStopOnRunFinished(checked === true)}
  />
  <Label htmlFor="session-stop-on-finish">Stop sandbox on run finished</Label>
</div>
```

This is a ~5 minute fix.

---

### [Major — still open] TypeScript SDK `Session` type: `stop_on_run_finished` is non-optional

`ts-sdk/src/session.ts` still declares:
```typescript
stop_on_run_finished: boolean;   // ← non-optional, required field
```

The DB column is `BOOLEAN` (nullable, no DEFAULT). Existing sessions return `null`/`undefined` from the API — a non-optional `boolean` will cause type errors in consumers or silent coercion. Should be:
```typescript
stop_on_run_finished?: boolean | null;
```

The Python SDK correctly uses `Optional[bool] = None`. The Go SDK uses `*bool`. The TS type should match.

---

### New commit assessment ✅

The `codes.NotFound` handler in `execAfterReady` is correct. One minor pattern note: `checkCancel()` should be `defer checkCancel()` to ensure cleanup on all paths — same pattern as the other cancel calls in the PR. Low risk here since there's no early return between creation and use, but `defer` is idiomatic.

---

Two changes before this can merge: Shadcn checkbox + TS type nullability. Both are small and isolated.

*— Amber*

jsell-rh (approved):
🤖 **Amber — PR #368: Approved (commit `dc3206e`)**

All blockers from my three prior reviews are resolved in the latest commit.

**Verified fixes:**
- ✅ **Shadcn Checkbox** — `<input type="checkbox">` replaced with `import { Checkbox } from '@/components/ui/checkbox'` and correct `onCheckedChange` handler
- ✅ **TS SDK nullability** — `stop_on_run_finished?: boolean` now optional, matching the nullable DB column and Go/Python SDKs
- ✅ **`defer completeCancel()`** — context canceller is now correctly deferred in `execAfterReady`

**Still noting for awareness (non-blocking):** The migration column has no `DEFAULT` — existing sessions will have `NULL`, and the code defaults to `stopOnRunFinished = true` for `NULL`. This means existing gateway sessions will begin auto-stopping after upgrade, which is a behavior change. Recommend documenting in release notes so operators know what to expect.

**What's solid across the full PR:**
- `ExecExitError` sentinel + `errors.As` pattern is clean and testable
- Independent contexts per step in `deprovisionSessionSandbox` — real correctness fix
- `codes.NotFound` detection in exec retry loop prevents spurious `Failed` on manual stop
- Full-stack field propagation (OpenAPI → DB → Go/Python/TS SDKs → UI) is thorough
- Tests cover the new error type and phase transitions

Good work iterating quickly on the feedback.

*— Amber*

jsell-rh (approved):
🤖 **Amber — PR #368 Follow-up (commits `bfde7de` + `82575d6`)**

Two new commits since my approval at `68ca399` — both look clean.

**`bfde7de` — SDK header regeneration:** Diff is SHA256/timestamp metadata only across Go SDK files; no type definition changes. The TS SDK fix (`stop_on_run_finished?: boolean`) and Python SDK (`Optional[bool]`) from `dc3206e` are untouched. ✅

**`82575d6` — e2e sandbox lifecycle tests:** Solid coverage:
- Long-running: 12 × 10s polling to confirm session + pod stay `Running` after LLM response ✅
- Short-running: start → PATCH `stop_on_run_finished=true` → wait for LLM → verify `Completed` + sandbox pod gone ✅
- `SHORT_SESSION_ID` guarded with `${SHORT_SESSION_ID:-}` in cleanup loops ✅
- Section renumbering (12–13 → 13–14) is clean ✅

**One minor note:** The `SHORT_SBX_NAME` derivation (`session-$(echo "${SHORT_SESSION_ID:0:40}" | tr ...)`) matches the existing pattern for `CREATED_SESSION_ID` — consistent. The 120s deprovisioning window is realistic given control plane reconcile latency.

Approval stands. No new issues introduced.

*— Amber*

jsell-rh (approved):
🤖 **Amber — PR #368 Follow-up (commit `719b8de`)**

New commit since my last approval at `82575d68` — looks clean. Approval stands.

**What changed:**

1. **Re-read at decision time in `execAfterReady`** — Adds a second SDK fetch of `stop_on_run_finished` *after* the exec stream finishes, so a `PATCH` that arrives mid-session is honoured. The design is correct: the initial read at provisioning time sets the env var and drives the initial exec call; the late read drives the actual stop decision. Using `context.Background()` for the late fetch is appropriate since the session context is done by that point.

2. **Non-pointer SDK usage** — `freshSession.StopOnRunFinished` (plain `bool`, not `*bool`) throughout, consistent with the SDK regeneration in `bfde7de`. The `buildEnv` simplification (`if session.StopOnRunFinished {`) is correct.

3. **Implicit default-to-false for NULL DB rows** — Worth noting as a positive: existing sessions with a NULL DB column will now get `false` from the API (bool zero value), overriding the `stopOnRunFinished := true` initial default. This resolves the behavioral concern I raised in my round-one review about existing gateway sessions unexpectedly auto-stopping after upgrade. The only case where the `true` default still applies is when the SDK fetch itself errors — which is a good fail-safe (auto-stop on error rather than leaving a sandbox running indefinitely).

**No new issues introduced.** The double-read pattern is clean, no data races (provisioning completes before the goroutine reaches the decision point), and no API surface changes.

*— Amber*

jsell-rh (approved):
🤖 **Amber — PR #368 Follow-up (commits `92f4f5e`, `1ad2e3d`)**

Two new commits since my last approval at `c7eff8fb` — both are clean correctness fixes. Approval stands.

**`92f4f5e` — parse start request body before `handlers.Handle` consumes it**

Root cause analysis is correct: `handlers.Handle()` does an `io.ReadAll` on `r.Body` before invoking the `Action` closure, so any `json.Decode` inside the closure reads an empty body and silently fails. Moving the body decode *before* the `handlers.Handle` call and inlining the handler logic is the right fix — not a workaround, it's the architecturally correct approach.

Specific callouts:
- Each error path now calls `handlers.HandleError(ctx, w, err)` and returns immediately — consistent with the rest of the codebase. ✅
- `_ = json.NewEncoder(w).Encode(resp)` — discarding the encode error is acceptable here since the response header (`StatusCreated`) is already committed at that point; there's nowhere useful to send an error. ✅
- No tokens in logs, no `panic()`, no security concerns introduced. ✅

**`1ad2e3d` — duplicate migration ID fix**

Bumping `202607150001` → `202607150005` for the `stop_on_run_finished` session migration correctly resolves the gormigrate startup fatal from the ID collision with the clusters migration. The chosen ID (`...0005`) leaves room between existing migrations. ✅

No new issues introduced across either commit.

*— Amber*

PR #370, opened 2026-07-15
feat: gateway route exposure, acpctl gateway CLI, and localhost cert SANs

bsquizz:
## Summary

Gateway route exposure via GRPCRoute + BackendTLSPolicy, new `acpctl gateway` CLI subcommands, localhost SAN support for port-forward access, and OpenShift Local (CRC) deployment tooling.

### acpctl Gateway CLI

New `acpctl gateway` subcommand tree for managing gateway connections from the command line:

- **`acpctl gateway setup-cli [name]`** — configures the local openshell CLI to connect to a gateway. Detects whether the gateway uses mTLS or OIDC authentication and takes the appropriate setup path:
  - **mTLS gateways**: writes config directly without triggering an interactive OIDC browser prompt
  - **OIDC gateways**: reuses existing acpctl credentials for non-interactive setup when available, otherwise falls back to interactive `openshell gateway add/login`
- **`acpctl gateway remove-cli [name]`** — removes a local openshell registration
- **`acpctl get gateways`** (aliases: `gateway`, `gw`) — lists gateways with name, image, DNS names, route address, and age. Single-gateway view includes connection info and setup hints.

**setup-cli flags:**
| Flag | Description |
|------|-------------|
| `--gateway-url` | Explicit gateway URL; if omitted, uses the gateway's `routeAddress` from the API |
| `--project` | Project/namespace to look up the gateway in |
| `--name` | Override the local registration name (default: `<project>-<gateway-name>`) |
| `--print` | Dry-run: print openshell commands instead of executing them |
| `--kubectl` | Fall back to `kubectl port-forward` + cert extraction when no route address exists |

**Auth detection (`gatewayAuthMode()`):** inspects the gateway's OIDC config from the API response — if `oidc.issuer` is present, mode is `oidc`; otherwise `mtls`.

### Certificate Management: Localhost SAN for Port-Forward Access

The gateway server TLS certificate previously only included the in-cluster DNS name (`openshell-gateway.<namespace>.svc.cluster.local`), which caused TLS hostname verification to fail when connecting via `kubectl port-forward` to `localhost`. Two independent certificate generation paths were updated:

1. **Certgen Job** (`internal/gateway/manifests.go`): `--server-san=localhost` is now unconditionally prepended to the certgen container args, with deduplication against user-provided `ServerDnsNames`
2. **cert-manager Certificate** (`internal/reconciler/gateway_reconciler.go`): `localhost` is hardcoded in the `dnsNames` list alongside the service FQDN when building the cert-manager `Certificate` resource

This enables `acpctl gateway setup-cli --kubectl` to work without `--gateway-insecure` in test environments where `kubectl port-forward` is the access method. The `OPENSHELL_GATEWAY_INSECURE` env var workaround and `--gateway-insecure` flag are no longer needed for localhost connections.

### Gateway Route Exposure

The control plane now automatically exposes OpenShell gateways via Kubernetes Gateway API GRPCRoute resources when a compatible networking Gateway is available on the cluster:

- **GRPCRoute + BackendTLSPolicy** — per-tenant GRPCRoute, BackendTLSPolicy (re-encrypt TLS), and CA ConfigMap created with OwnerReferences to the gateway StatefulSet
- **Dual platform detection** — independently detects `route.openshift.io` (for OpenShift SCC bindings) and `gateway.networking.k8s.io/v1` GRPCRoute CRD + networking Gateway (for GRPCRoute provisioning). Each is a graceful no-op when absent.
- **Auto-populated `routeAddress`** — derived from GRPCRoute hostname and networking Gateway listener protocol, patched onto the Gateway API resource
- **Gateway API prerequisites scripts** — `setup-gateway-api.sh` installs the `acpgw` networking Gateway, generates wildcard TLS certs, and creates a passthrough Route; `setup-gateway-trusted-ca.sh` extracts CRC ingress CA for OIDC endpoint trust
- **RBAC** — ClusterRole updated for `gateway.networking.k8s.io` (grpcroutes, backendtlspolicies, gateways) and `cert-manager.io` (issuers, certificates)
- **Data model** — `route` (JSONB) and `route_address` (TEXT) columns added to gateways table; OpenAPI schema extended with `GatewayRoute` and `routeAddress`

### OpenShift Local (CRC) Deployment Tooling

New `make crc-up` target deploys the full ACP stack to OpenShift Local:

- **`make crc-up`** — end-to-end: namespace, Keycloak with custom branded theme, PostgreSQL with RHEL image, all ACP components, Gateway API prerequisites, trusted CA injection
- **`make crc-down`** — teardown
- **`make crc-up LOCAL_IMAGES=true`** — builds all component images from source and pushes to the CRC internal registry
- **`crc-reload-component` / `crc-reload-images`** — inner-loop rebuild and push for individual or all components
- **OpenShift Local overlay** (`manifests/overlays/openshift-local/`) — kustomization, API server and UI Routes, Keycloak deployment, PostgreSQL init scripts, LimitRange (50m CPU default for constrained CRC nodes), credentials
- **CA cert trust instructions** printed at end of `make crc-up` (Fedora/RHEL + macOS)

### Specs

- **New:** `gateway-route-exposure.spec.md` — GRPCRoute architecture, requirements, detection, config, RBAC, CRC prerequisites, E2E testing
- **Updated:** `data-model.spec.md` (Gateway fields: oidc, route, route_address), `gateway-provisioning.spec.md` (BackendTLSPolicy, cert-manager), `gateway-cli.spec.md` (ROUTE column, connection info), `gateway-rbac-policy.spec.md` (route access control), `identity-boundaries.spec.md`, `control-plane/conventions.spec.md` (seccompProfile)

### Other

- **E2E tests** — `route-e2e-test.sh` (new) for GRPCRoute connectivity; `gateway-e2e-test.sh` expanded
- **mock-llm** — OpenShift SCC compatible (no hardcoded UID); non-root user added for Kind
- **OwnerReferences** — removed `blockOwnerDeletion` to prevent GC stalls; added `controller: true` to session Role/RoleBinding refs
- **SecurityContext** — restricted context with emptyDir `/tmp` on CRC api-server overlay
- **Test token** lengthened to meet SDK 20-char minimum
- **api-server Dockerfile** — removed unnecessary `util-linux` package (unavailable on arm64)
- **CLI renames** — `--project-id` → `--project`, `--confirm` → `--yes`/`-y` globally

## Test plan

- [x] E2E route tests passing (9/9)
- [x] `setup-cli --print` shows correct commands for mTLS gateway
- [x] `acpctl apply -f examples/overlays/tenant-c --project tenant-c` creates gateway with route field
- [ ] `setup-cli --kubectl` detects mTLS and configures without OIDC prompt
- [ ] `setup-cli` with OIDC gateway still works as before
- [ ] `make crc-up` completes successfully on OpenShift Local
- [ ] `make crc-up LOCAL_IMAGES=true` builds and pushes local images
- [ ] mock-llm pod starts under OpenShift restricted SCC
- [ ] GRPCRoute creation on OpenShift cluster with Gateway API
- [ ] Graceful no-op on Kind (no GRPCRoute API or networking Gateway)
- [ ] `kubectl port-forward` + `acpctl gateway setup-cli --kubectl` works without `--gateway-insecure` (localhost SAN in cert)


jsell-rh (commented):
## 🤖 Amber Review — PR #370: Gateway Route Exposure Spec

**Summary:** Well-structured spec with solid TLS rationale and good BDD scenario coverage. Two gaps need addressing before this merges: a missing security scenario for Route access control (noted in the referenced `gateway-rbac-policy.spec.md`) and an unresolved behavioral question around Route Watch in the reconciler loop. Minor items inline.

**Confidence:** High (90%) — spec-only change, no production code.

---

### ✅ What's solid

- Re-encrypt TLS rationale is rigorous and correct — the SAN chicken-and-egg argument is the right reason to rule out passthrough, and it's clearly explained.
- The certgen timing scenario (secret not yet available → skip + retry) is the right reconciler pattern and aligns with existing wait-for-job patterns in the CP.
- `route: {}` for auto-assigned host is clean UX; consistent with OIDC optionality from PR #363.
- Migration table is comprehensive and will serve the implementation wave well.
- Non-OpenShift graceful-skip scenario is appreciated — important for Kind dev clusters.

---

### 🔴 Issues

**1. RBAC spec for Route resources is incomplete — who can create/modify a Gateway with `route`?**

`gateway-rbac-policy.spec.md` defines three tiers (Admin, Editor, Viewer) for Gateway CRUD, but this spec adds a new capability (`route` field) that has external-access implications. The spec says the ClusterRole gets `routes` permissions, but doesn't address:
- Should Editors be able to configure `route`? Or Admin-only (given it exposes the gateway externally)?
- Is the `routeAddress` (read-only) visible to Viewers?

This doesn't need a full RBAC spec here, but a "Security Considerations" section or a reference to how the existing tier model applies to the new field would close the gap. Without it, implementers will have to guess.

**2. Route Watch — reconciler loop design gap**

The spec says the reconciler reads `Route.status.ingress[].host` to populate `routeAddress`, and re-checks on the next cycle if not yet available. But it doesn't address how the reconciler is triggered when the Route status updates — OpenShift populates `.status.ingress` asynchronously after the HAProxy router picks up the Route.

If the reconciler only runs on Gateway events (gRPC watch stream), `routeAddress` will stay empty until the next Gateway event arrives (e.g., a PATCH). This could leave the address unpopulated indefinitely in a steady-state system.

Options to spec:
- Watch Route resources in the project namespace and trigger Gateway reconciliation when Route status changes
- Periodic re-sync interval (already exists in some reconcilers?)
- Accept eventual consistency (document the lag and the trigger)

Please add a scenario or a note in the Architecture section.

---

### 🟡 Minor

**3. `haproxy.router.openshift.io/balance` annotation missing**

The spec sets `timeout: 3600s` (correct for gRPC streams) but doesn't mention `haproxy.router.openshift.io/balance: leastconn`. For gRPC long-lived streams this matters more than for HTTP — without it, round-robin can cause hot pods. Not a blocker, but worth a note or a deliberate omission with reasoning.

**4. OwnerReference on Route not addressed**

Per CLAUDE.md: "OwnerReferences on all new K8s child resources." The spec lists the Route manifest structure but omits an OwnerReference pointing to the StatefulSet or a parent resource. Implementation will need to decide the owner (gateway StatefulSet? or no owner since the Route is deleted by the reconciler on `route` field removal). A note clarifying the intended owner or an explicit "no OwnerReference needed because X" would help implementers.

**5. `specs/index.spec.md` — `BE` missing from components column**

```
| `platform/gateway-route-exposure.spec.md` | platform | Gateway (route), Route | CP, CLI, SDK | ...
```

The API server (`BE`) is also a consumer — it handles the `route` JSONB column, the `routeAddress` text column, and the OpenAPI schema extension. `BE` should be in the components column.

---

### Non-blocking observation

The reference to `security/gateway-rbac-policy.spec.md` in the header is correct, but that file doesn't yet have a section on Route permissions. Consider opening a follow-up issue or marking this PR as a draft dependency.

---

*— Amber*

jsell-rh (changes requested):
Requesting changes on two items before approval: (1) Route Watch / reconciler trigger for `routeAddress` population needs to be specced — implementation will get this wrong without guidance. (2) RBAC tier applicability for the new `route` field (external exposure capability) needs at least a note. Minor items (OwnerReference, index components column) are easy fixes. The TLS design and overall structure are solid. Happy to approve once these gaps are addressed.

jsell-rh (approved):
## 🤖 Amber Review — PR #370: Round 2 (commits `88e591d`, `454f2cda`)

All blockers from my prior review are resolved. Approving.

---

### Verification against prior blockers

**[Major — resolved] RBAC for `route` field**
`gateway-rbac-policy.spec.md` now has a dedicated "Gateway Route Access Control" section: Admin-only write, Editor gets 403, Viewer can read `routeAddress`. This is exactly what was needed — implementers now have clear normative requirements rather than having to guess the tier applicability. ✅

**[Major — resolved] Route Watch / reconciler trigger for `routeAddress` population**
The new "Route Status Reconciliation" requirement covers all three cases: watch-triggered update, periodic resync as catch-all, and hostname change propagation. The "OR periodic resync" phrasing is the right flexibility — it lets the implementation choose the mechanism without over-specifying. ✅

**[Minor — resolved] OwnerReference on Route**
Route manifest now includes `ownerReferences` pointing to `StatefulSet/openshell-gateway` with `controller: true`. This is the right owner choice since the StatefulSet owns the gateway's lifecycle. ✅

**[Minor — resolved] Index `BE` column**
`CP, CLI, SDK, BE` — correct. ✅

---

### New content: E2E Testing on OpenShift Local (commit `454f2cda`)

This is well-scoped and follows the existing Kind overlay pattern. The CRC overlay approach (separate `openshift-local/` overlay, same tenants as Kind, mock-LLM parity) is the right way to test OpenShift-specific resources without polluting the Kind dev environment. The `make crc-up / crc-down / crc-reload-component` target design is consistent with the existing `kind-*` targets.

One non-blocking note: step 1 of `crc-up` says "push to internal registry (or load via `oc image mirror`)" — the implementation will need to pick one; the ambiguity is fine at spec level.

---

### Carry-forward (non-blocking)

`haproxy.router.openshift.io/balance: leastconn` — still not mentioned, but I flagged this as minor and it's not a gate. Implementation can add it as a follow-up based on observed load patterns.

---

This spec is solid and ready for the implementation wave.

*— Amber*

specs/platform/gateway-route-exposure.spec.md
```diff
@@ -0,0 +1,566 @@
...
+- **CLI integration** — `acpctl get gateway` displays the route address and `acpctl gateway setup-cli` uses it instead of port-forwarding when available
+
+---
+
+## Architecture
+
+### Route Data Flow
+
+```
+External Client (openshell CLI, browser)
```
bsquizz: There's no 'browser' in this flow

specs/platform/gateway-route-exposure.spec.md
```diff
@@ -0,0 +1,566 @@
...
+
+#### Scenario: Gateway with explicit route host
+
+- GIVEN a Gateway resource with route configuration:
+  ```yaml
+  kind: Gateway
+  name: openshell-gateway
+  project: tenant-a
+  image: ghcr.io/nvidia/openshell:v0.0.70
+  serverDnsNames:
```
bsquizz: This gets confusing now that route has 'host'

Let's change this field to 'internalServiceHostname'

specs/platform/gateway-route-exposure.spec.md
```diff
@@ -0,0 +1,566 @@
...
+- AND the `serverDnsNames` field is updated, triggering certificate regeneration
+- WHEN the new `openshell-server-tls` Secret is created
+- THEN the GatewayReconciler SHALL update the Route's `destinationCACertificate` with the new CA certificate
+- AND existing connections through the Route MAY experience a brief interruption during the certificate rotation
+
+---
+
+### Requirement: Route Status Reconciliation
+
+The GatewayReconciler SHALL detect Route status changes to populate the Gateway's `routeAddress` field. Because Route status is updated asynchronously by the OpenShift router (not by a Gateway gRPC event), the reconciler SHALL use Route resource watches or periodic resync to detect when the hostname becomes available.
```
bsquizz: We will need to ensure that the roles/rolebindings are updated on control-plane's svc account to allow the control plane to CRUD on Route

specs/platform/gateway-route-exposure.spec.md
```diff
@@ -0,0 +1,566 @@
...
+- THEN it SHALL log a warning that Routes are not available on this cluster
+- AND it SHALL skip Route creation without failing the overall gateway reconciliation
+
+---
+
+### Requirement: RBAC for Route Resources
+
+The control plane ServiceAccount SHALL have permissions to manage Route resources in project namespaces.
+
+#### Scenario: Route RBAC rules
```
bsquizz: Re my last comment, within that section, you can mention control-plane CRUD is discussed later in section "Route RBAC rules"

specs/platform/gateway-route-exposure.spec.md
bsquizz: Also, let's modify this so that 'internalServiceHostname' does NOT have to be defined.

If project is named <project>, then we can assume that the gateway's internalServiceHostname will be:
openshell-gateway.<project>.svc.cluster.local

UNLESS the value is explicitly set

specs/platform/gateway-route-exposure.spec.md
```diff
@@ -0,0 +1,566 @@
...
+Route resources are not available on Kind clusters (`route.openshift.io` API group does not exist). To e2e test the Route provisioning flow end-to-end, the platform SHALL support deployment to OpenShift Local (formerly CodeReady Containers / CRC) with the same tenant configuration used by the Kind development environment.
+
+### Requirement: OpenShift Local Deployment Target
+
+The project SHALL provide a Makefile target and kustomize overlay to deploy the full ACP stack to an OpenShift Local instance, mirroring the Kind development environment. This enables e2e testing of OpenShift-specific features (Routes, service-CA TLS, RBAC via namespace access) that cannot be tested on Kind.
+
+#### Scenario: Deploy to OpenShift Local via Make target
+
+- GIVEN an OpenShift Local instance is running and `oc` is authenticated
+- WHEN a developer runs `make crc-up`
```
bsquizz: Also make sure we support SKIP_TENANT_SETUP, the same as we do in 'make kind-up'. This allows us to demo a tenant setup end-to-end without having the Makefile "automatically set it all up for us"

specs/platform/gateway-route-exposure.spec.md
```diff
@@ -0,0 +1,566 @@
...
+#### Scenario: Rebuild and redeploy a single component
+
+- GIVEN the ACP stack is deployed on OpenShift Local
+- WHEN a developer runs `make crc-reload-component COMPONENT=ambient-control-plane`
+- THEN only the specified component's image SHALL be rebuilt, pushed, and its Deployment/StatefulSet patched
+- AND no other components SHALL be restarted
+
+### Requirement: OpenShift Local Kustomize Overlay
+
+A kustomize overlay SHALL exist at `components/manifests/overlays/openshift-local/` that configures the ACP stack for OpenShift Local. This overlay SHALL reuse existing kustomize components and the base manifests, following the same pattern as the existing `kind/` overlay.
```
bsquizz: Make sure to reference the template in 'components/manifests/templates/template-services.yaml' -- this is the OpenShift template we have been deploying previously.

specs/platform/gateway-route-exposure.spec.md
```diff
@@ -0,0 +1,566 @@
...
+The OpenShift Local deployment SHALL provide the same development experience as the Kind cluster, with the addition of OpenShift-specific features.
+
+#### Scenario: Same tenants and test configuration
+
+- GIVEN the Kind environment provisions tenants `tenant-a`, `tenant-b`, `tenant-c`, etc.
+- THEN the OpenShift Local environment SHALL provision the same tenants
+- AND each tenant SHALL have the same agents, credentials, and mock-LLM configuration
+- AND the mock-LLM server SHALL be deployed for self-contained testing (per `e2e-test-tooling.spec.md`)
+
+#### Scenario: Gateway Route e2e test
```
bsquizz: Note: we cannot run this e2e test in github CI at this point, due to the footprint of OpenShift local being too resource-intensive for a github runner

jsell-rh (approved):
🤖 **Amber — PR #370 Follow-up (commit `9a694c18`)**

New commit since my approval at `454f2cda` — reviewing the delta. Approval stands.

**What changed:**

1. **Data flow diagram cleanup** — "browser" removed from the Route data flow. Correct: the openshell CLI connects directly, a browser is not in this flow. ✅

2. **`serverDnsNames` → `internalServiceHostname`** — Cleaner field name, single value with a documented default (`openshell-gateway.<project>.svc.cluster.local`). The default derivation from the project name is the right call — implementors won't have to guess. The schema table and auto-assign scenario both updated consistently. ✅

3. **RBAC cross-reference in Route Status Reconciliation** — The note linking to the "RBAC for Route Resources" section closes the gap where the reconciler behavior was described without a clear pointer to the permissions it requires. Good housekeeping. ✅

4. **`SKIP_TENANT_SETUP` support for `crc-up`** — Parity with `kind-up` behavior. The scenario description is clear and correctly explains the demo workflow (developer manually provisions tenants to exercise the full setup flow). ✅

5. **`template-services.yaml` reference in overlay section** — Useful pointer for implementation; avoids the implementor having to discover the production overlay reference independently. ✅

6. **GitHub CI limitation note** — Explicit acknowledgment that the CRC e2e test is local-only due to resource footprint. Honest and correct — this prevents someone from trying to wire it into CI and discovering the limitation the hard way. ✅

**No new issues introduced.** All changes are scoped, consistent with prior spec content, and improve implementability.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #370 Round 4 (commits `ee1b12ed`, `2c682fa7`, `6f14a743`)

Three new commits since my last approval at `9a694c18`. Reviewing the delta. **Approval stands.**

---

### What's new and the verdict on each

**OpenShift Cluster Detection (commit `ee1b12ed`)** — Well-specified. The three scenarios (running on OpenShift, running on Kind, Gateway with `route` on non-OpenShift) cover the main cases cleanly. One minor gap: the spec doesn't address the error case for the API discovery call itself (network error, RBAC failure, transient API server unavailability at startup). Should the CP treat a discovery failure as "non-OpenShift" (safe default — no spurious Route creation) or fail loudly? Either answer is fine; the spec just doesn't say. Implementation will have to make a judgment call. Not a gate.

**`setup-cli --print` default behavior (commit `2c682fa7`)** — Running `openshell gateway add` by default (executing rather than just printing) is the right UX call for the happy path. `--print` for preview/copy is consistent with how other CLIs handle this. ✅

**`--kubectl` fallback (commit `6f14a743`)** — The full fallback spec (port-forward, cert extraction, write to `~/.config/openshell/`) is clear. One implementation note: the `--kubectl --print` scenario says the CLI SHALL still manage the port-forward and extract certs even in print mode. This is probably correct — the printed command is useless without those prerequisites — but it's a side-effecting `--print` that may surprise users. Worth a clarifying comment in the implementation PR.

**`--kubectl` ignored when routeAddress exists** — Correct behavior, clearly specified. ✅

---

### One convention carry-forward

The Route manifest spec includes `blockOwnerDeletion: true` in the ownerReferences block. CLAUDE.md convention is `Controller: true`, no `BlockOwnerDeletion`. Low blast radius for a spec-only PR, but the implementation PR should omit `blockOwnerDeletion` from the actual manifest to stay in convention.

---

No new blockers. All three commits are improvements that close remaining gaps.

*— Amber*

jsell-rh (changes requested):
## 🤖 Amber Review — PR #370 Round 5 (implementation commits `695d2fe`–`590665c`)

Prior reviews covered the spec phase through commit `6f14a743`. This round covers the full-stack implementation: Go control plane, API server, SDKs, CLI, CRC overlay, and E2E test.

The implementation is substantial and well-executed. The GRPCRoute + BackendTLSPolicy approach, null-JSONB detection in the PATCH handler, and OpenShift cluster detection via API discovery are all solid. One blocker needs addressing before merge.

---

### 🔴 Blocker — `runAsNonRoot: false` in CRC overlay (ACP security convention)

`components/manifests/overlays/openshift-local/api-server-security-patch.yaml`:

```yaml
spec:
  template:
    spec:
      securityContext:
        runAsNonRoot: false
```

This overrides the base manifest's `runAsNonRoot: true` on the API server. CLAUDE.md requires `runAsNonRoot` on all containers; setting it to `false` explicitly violates this on the CRC overlay.

The real fix is to add an `emptyDir` volumeMount for any path the API server writes to at runtime (typically `/tmp`), then remove this patch entirely so the base `runAsNonRoot: true` applies on CRC as well. The OpenShift SCC will assign a UID from the namespace range — you don't need `runAsNonRoot: false` for that; you just need to not hardcode `runAsUser`.

If the API server binary only needs `/tmp`, this is the right pattern (matches what the UI overlay already does):

```yaml
# In api-server-security-patch.yaml — replace runAsNonRoot: false with:
containers:
- name: ambient-api-server
  volumeMounts:
  - name: tmp
    mountPath: /tmp
volumes:
- name: tmp
  emptyDir: {}
```

Then delete the `securityContext: runAsNonRoot: false` block.

---

### 🟡 Major — `blockOwnerDeletion: true` on GRPCRoute OwnerReference

CLAUDE.md convention: `controller: true`, **no `blockOwnerDeletion`**. The implementation sets both:

```go
"controller":         true,
"blockOwnerDeletion": true,
```

`blockOwnerDeletion` causes the GC to stall the StatefulSet deletion until the GRPCRoute is fully gone — which can get stuck if the Gateway API controller is unavailable. The spec's carry-forward note from my Round 4 review explicitly flagged this. Please remove `blockOwnerDeletion` from the GRPCRoute (and BackendTLSPolicy, if it has one too).

---

### ✅ What's solid

**Null-JSONB PATCH detection** (`handler.go`): Pre-reading the body with `io.ReadAll` + re-wrapping `r.Body` with `io.NopCloser` is the correct Go idiom for detecting `"route": null` vs absent `route`. The trickiest part of this implementation, done correctly.

**OpenShift detection** (`detectOpenShift()`): Checking for `route.openshift.io/` in `ServerGroupsAndResources()` is the right approach — safe fallback to non-OpenShift on discovery error.

**CRC security patches**: `api-server-db-security-patch.yaml` correctly sets `runAsNonRoot: true`, drops ALL capabilities, uses `secretKeyRef` throughout. Good.

**GRPCRoute reconciler**: The lazy-OwnerReference pattern (create without owner if StatefulSet not yet present, populate on next reconcile) is a reasonable workaround for startup ordering.

**E2E test** (`route-e2e-test.sh`): `set -euo pipefail`, timestamped names, `trap cleanup EXIT`, 150s retry polls with diagnostic failure messages — all correct conventions.

**mock-llm Dockerfile**: `useradd --create-home mockllm && chmod -R g=u /app` without hardcoded UID is correct for OpenShift SCC. Follow-up commit `b4373e8` correctly restored `USER` for Kind `runAsNonRoot` enforcement.

**SDK regeneration**: All three SDKs consistently updated with `route_address` snake_case. The camelCase → snake_case fix was the right call.

**RBAC additions**: GRPCRoute, BackendTLSPolicy, cert-manager, Gateway API read — all scoped correctly.

---

### Non-blocking observations

- **`reconcileOpenShiftSCC`**: Not visible in my excerpt. Confirm it doesn't grant `anyuid`/`privileged` — `restricted-v2` is the correct SCC for non-root gateway pods.
- **`detectBaseDomain` error handling**: Confirm a warning is logged on failure so operators can diagnose missing auto-hostnames.
- **Commit `ca632860` ("Route creation checkpoint")**: Bare message, doesn't follow conventional commits. Squash on merge will clean this up.

**Confidence:** High — two issues (one blocker, one convention deviation), both small fixes. The overall implementation is solid.

*— Amber*

jsell-rh:
🏷️ Updating label: removing `amber/approved`, adding `amber/change-requested` — new implementation commits have a blocker (`runAsNonRoot: false` in CRC overlay) and a convention deviation (`blockOwnerDeletion` on GRPCRoute). See review above.\n\n*— Amber*

jsell-rh:
_(Note: adding `amber/change-requested` label via API — please remove `amber/approved` manually since the GitHub MCP label tool only adds, not removes. Or I can use a direct API call.)_

jsell-rh (approved):
## 🤖 Amber — PR #370 Round 6 (commits `acd462e`, `ea3273c`)

Both blockers from my Round 5 review are addressed. **Approving.**

---

### Verification against Round 5 blockers

**[Blocker — resolved] `runAsNonRoot: false` in CRC overlay**

`api-server-security-patch.yaml` now sets exactly the pattern I suggested:
- `runAsNonRoot: true` ✅
- `allowPrivilegeEscalation: false` ✅
- `capabilities.drop: [ALL]` ✅
- `readOnlyRootFilesystem: true` ✅
- `emptyDir` `/tmp` volumeMount ✅

The `runAsNonRoot: false` override is gone. The CRC overlay now enforces the same security posture as the base manifest. ✅

**[Major — resolved] `blockOwnerDeletion: true` on GRPCRoute / BackendTLSPolicy / CAConfigMap OwnerReferences**

`gateway_reconciler.go`: `blockOwnerDeletion` removed from all three builders (GRPCRoute, BackendTLSPolicy, CAConfigMap). Only `controller: true` remains — correct convention. ✅

Bonus fix in `kube_reconciler.go`: session Role/RoleBinding OwnerReferences that previously had `blockOwnerDeletion: true` (without `controller`) are now `controller: true` without `blockOwnerDeletion`. This was not flagged in my review but is the right fix — the missing `controller: true` there was an independent convention gap. Good catch. ✅

---

### `ea3273c` — mock-llm `USER 1001`

Switching from `useradd --create-home mockllm && USER mockllm` to `USER 1001` is a pragmatic fix for Kind compatibility (which needs a numeric non-root UID in the image for `runAsNonRoot` enforcement). For OpenShift/CRC with `restricted-v2` SCC, a numeric UID is actually cleaner since OpenShift validates non-root numerically. The `RUN chmod -R g=u /app` retained for group writability is correct. ✅

Non-blocking note: if CRC E2E tests ever run with `runAsUser` enforced to a specific range (say 1000000–1000999 as is typical), `USER 1001` would be overridden at admission — the container would still run as the SCC-assigned UID, and `g=u` permissions ensure the app directory is writable. This is fine behavior.

---

### Carry-forward (non-blocking, no action needed)

- `haproxy.router.openshift.io/balance: leastconn` — still unaddressed, still not a gate. Flag as a follow-up if gRPC stream balancing becomes a concern under load.
- `detectOpenShift` API discovery error handling — implementation should log a warning on failure; mentioned in Round 5. Check during implementation PR review.
- `ca632860` "Route creation checkpoint" commit — bare message, not conventional commit format. Squash on merge will clean this.

---

No new issues in these two commits. The implementation is solid end-to-end.

*— Amber*

jsell-rh:
🏷️ Updating label: removing `amber/change-requested`, adding `amber/approved` — both Round 5 blockers (`runAsNonRoot: false` CRC overlay and `blockOwnerDeletion` on OwnerReferences) are fully resolved. See approval review above.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #370 Round 7 (commits `8955f7b`–`f88f4b7`)

Five new commits since my Round 6 approval at `ea3273c`. All are CI/CLI bug fixes. **Approval stands.**

---

### What changed and the verdict on each

**`8955f7b` — Wire `acpctl gateway setup-cli` into CI, add `--name` flag, fix port-forward lifetime**

- `--name` flag is clean — `resolveLocalName` fallback when empty is the right default-override pattern. ✅
- Port-forward lifecycle fix is correct: `defer kill` removed from the success path, killed only on `localPort` parse failure or `setupOpenshellGateway` error. The PID is printed so the user can kill it manually. ✅
- CI now registers the gateway via `acpctl` before running e2e tests, with a `|| echo "Warning..."` fallback for backward compat. ✅

**`82405b9` → `773cf1c` — Correct TLS secret split for mTLS cert extraction**

`773cf1c` supersedes `82405b9` and gets the secret semantics exactly right:
- `openshell-ca-tls` → `ca.crt` (root CA for verifying the server cert) ✅
- `openshell-client-tls` → `tls.crt` / `tls.key` (client identity for mTLS auth) ✅

The previous code was reading all three fields from `openshell-server-tls`, which conflates the server's identity with the client's and doesn't exist in Kind. This is the correct fix. `setup-gateway-cli.sh`, `gateway-e2e-test.sh`, and `setup.go` all updated consistently. CI now waits for both secrets before proceeding. ✅

**`03ffea0` → `f88f4b7` — `OPENSHELL_GATEWAY_INSECURE=true` for localhost e2e connections**

Setting the env var at the top of `gateway-e2e-test.sh` rather than passing `--gateway-insecure` to every `openshell` invocation is the right approach — localhost connections can't match the server cert's cluster-DNS SAN, so this needs to apply globally in the test. `true` vs `1` fix is correct. ✅

---

No convention violations, no security concerns, no regressions. The intermediate commit `82405b9` will be collapsed by squash merge.

Carry-forward from prior rounds remains non-blocking: `haproxy.router.openshift.io/balance: leastconn` still unaddressed, still not a gate.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #370 Round 8 (commit `6fc8f09a`)

One new commit since my Round 7 approval at `f88f4b7e`. **Approval stands.**

---

### What changed and the verdict

**Root cause fix: `localhost` SAN missing from gateway server certs**

The real problem behind the `OPENSHELL_GATEWAY_INSECURE` workaround chain from the last two commits: the server cert's SAN list only included the in-cluster DNS name, so hostname verification always failed over `localhost`:`<port>`. Adding `localhost` to the SAN is the correct fix — the workaround band-aid is removed, and mTLS now works properly end-to-end.

**`gateway_reconciler.go` — cert-manager `Certificate` dnsNames (line 893)**

- `localhost` added unconditionally to every gateway cert. Correct: every gateway may be accessed via port-forward; namespace-scoped cert means no cross-namespace exposure. ✅
- Dedup logic correctly changed from single-element comparison (`!= dnsNames[0]`) to full-slice scan. The old logic was a latent bug — it would have allowed duplicates if `ServerDnsNames` contained the in-cluster hostname for a second gateway variant. ✅
- Note: existing certs will be reissued by cert-manager when this rolls out (SAN content changed). Expected behavior, not a problem.

**`manifests.go` — certgen Job fallback (line 285)**

- `localhost` prepended before the loop, then filtered from `ServerDnsNames` to prevent duplication. Correct ordering and dedup. ✅

**`gateway-e2e-test.sh`**

- `unset OPENSHELL_GATEWAY_INSECURE` — correct. The variable was disabling mTLS client cert presentation, defeating the purpose of the test. ✅
- `--gateway-insecure` removed from `openshell gateway add` — correct now that hostname verification passes. ✅
- `&>/dev/null 2>&1` → `&>/dev/null` — `2>&1` was redundant after bash `&>`; removal is correct. ✅
- Diagnostic `echo` messages to stderr on failure — good operational improvement for CI debugging. ✅

---

No convention violations, no security concerns. Clean fix.

Carry-forward from prior rounds remains non-blocking: `haproxy.router.openshift.io/balance: leastconn` unaddressed, not a gate.

*— Amber*

jsell-rh:
## 🤖 Amber — PR #370 Round 9 (commits `49458ef5`, `9f2a6cd8`)

Two new commits since my Round 8 approval at `a27998a0`. **Approval stands.**

---

### What changed and the verdict on each

**`49458ef5` — CRC deployment fixes (docs, patch, SSO)**

- `crc setup` prerequisite in README — correct and necessary; fresh installs skip this and get confusing errors. ✅
- Container name fix in `api-server-security-patch.yaml` (`ambient-api-server` → `api-server`) — this is a correctness fix, not cosmetic. The wrong name would have caused kustomize to inject a phantom container alongside the real one rather than patching it. ✅
- Keycloak realm redirect URIs for CRC routes — required for SSO to work through OpenShift Routes. Adding both `http://localhost:*` and the Route URL patterns is the right approach (allows both local port-forward access and Route access). ✅

**`9f2a6cd8` — Remove `util-linux` from api-server Dockerfile**

- `util-linux` is not in the `ubi-minimal` arm64 repos and was unused by the runtime binary. Removal is correct and improves multi-arch portability. ✅
- The runtime stage only runs the statically compiled Go binary — no shell utilities needed post-build. ✅

---

No convention violations, no security concerns. Both commits are targeted bug/compatibility fixes.

Carry-forward from prior rounds remains non-blocking: `haproxy.router.openshift.io/balance: leastconn` still unaddressed, still not a gate.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #370 Round 9 (commit `49458ef5`)

One new commit since my Round 8 approval at `a27998a0`. **Approval stands.**

---

### What changed and the verdict

**`README.md` — `crc setup` prerequisite**

Correct. `crc setup` downloads the VM image and configures the hypervisor — it's a mandatory first-time step that has bitten people before. Adding it as an explicit callout before `crc start` is the right UX. ✅

**`api-server-security-patch.yaml` — container name `ambient-api-server` → `api-server`**

This is a real correctness fix. Kustomize strategic merge patches match containers by `name`; a mismatched name causes Kustomize to inject a *phantom* second container rather than patching the existing one. The base deployment uses `api-server` as the container name, so this patch was a no-op until now — meaning the `runAsNonRoot: true` security fix from Round 6 wasn't actually applying to the CRC overlay. Fixing the name makes the patch land correctly. ✅

**`keycloak-realm.json` — CRC route redirect URIs**

Adding `https://ambient-ui-ambient-code.apps-crc.testing/*` to both `redirectUris` and `post.logout.redirect.uris` is required for SSO to function when users access the UI via the CRC Route (rather than localhost port-forward). Without this, Keycloak rejects the redirect as an unauthorized URI. The `webOrigins: ["+"]` wildcard already covers CORS, so no additional change needed there. ✅

---

No new issues. All three changes are targeted fixes that improve CRC deployment reliability.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #370 Round 10 (commits `9f2a6cd8`, `f7e406e9`, `e3fbf030`, `2c746906`)

Four new commits since my Round 9 approval at `49458ef5`. **Approval stands.**

---

### `9f2a6cd8` — Remove `util-linux` from api-server Dockerfile

Correct fix. `util-linux` is not available on `arm64` in UBI minimal repos and serves no purpose in a runtime-only stage that just runs a statically linked Go binary. ✅

---

### `f7e406e9` — Reduce resource requests/limits in openshift-local overlay

Adding dedicated resource patches for api-server, control-plane, and UI is the right approach — isolates CRC-specific tuning from the base manifests without polluting other overlays.

`api-server-db-security-patch.yaml` changes (3 add / 3 del): minor security context refinements. ✅

`limitrange-patch.yaml` 1-line delta: adjustment to LimitRange defaults. Fine for a resource-constrained dev environment. ✅

`kustomize.yaml`: new patches wired in correctly. ✅

No convention issues in what's visible. ✅

---

### `e3fbf030` — Sync migration init container image in `crc-up LOCAL_IMAGES` flow

Correct fix. The migration init container was running from the remote `quay.io` image even when `LOCAL_IMAGES=true` explicitly requested local builds for everything. Reading back the pushed image digest and applying it to the init container matches the pattern already used by `crc-reload-component`. ✅

---

### `2c746906` — Improve gateway table + harden manifest security

**CLI: `printGatewayTable` refactor (get/cmd.go)**

- VERSION / ADDRESS columns are cleaner than IMAGE / DNS NAMES / ROUTE — better signal density for operators. ✅
- `imageTag()`: `strings.LastIndex(image, ":")` correctly handles images with registries containing ports (e.g., `localhost:5000/foo:tag`) — `LastIndex` rather than `Index`. ✅
- `gatewayAddress()`: nil-check on `gw.Route`, then empty-string check on `RouteAddress` for "Not ready..." — correct precedence. Falls back to comma-joined DNS names when no route configured. ✅

**Manifests: `readOnlyRootFilesystem: true` + `/tmp` emptyDir**

This is a meaningful security improvement:

- `template-services.yaml`: `readOnlyRootFilesystem: true` added to both migration init container and api-server container, `/tmp` emptyDir volumeMount added to both. ✅
- `kind/ambient-api-server-dev-patch.yaml`: `/tmp` emptyDir added for the init container. ✅
- `openshift-local/ambient-api-server-dev-patch.yaml`: same. ✅
- The main `volumes` section already had `tmp emptyDir` (from Round 6); this commit adds the corresponding `volumeMounts` to the init container — was a latent gap. ✅

**Spec updates (gateway-cli.spec.md, gateway-route-exposure.spec.md)** — consistent with implementation changes. ✅

---

### One carry-forward (still non-blocking)

`haproxy.router.openshift.io/balance: leastconn` — unaddressed across 10 rounds. Not a gate for merge; flag as a follow-up issue if gRPC stream load balancing becomes a concern under production load.

---

No new issues. All four commits are targeted improvements with no regressions.

**Confidence:** [97%] High.

*— Amber*

PR #371, opened 2026-07-15
fix: store request prompt on session instead of duplicating agent prompt

jeremyeder:
## Problem

When running `acpctl agent start stella --project vteam-product-swarm --prompt "Why is the sky blue?"`, the agent never sees the `--prompt` value.

### Root cause

`start_handler.go` set `session.Prompt = agent.Prompt` (the agent's standing instructions). The control plane's `assembleInitialPrompt()` independently fetches `agent.Prompt`, so:

1. **Agent prompt duplicated** — appeared twice in the initial prompt sent to the runner
2. **User's `--prompt` dropped** — the API server composed it into a session message via `buildStartPrompt()`, but the control plane reads `session.Prompt`, not session messages, so the task prompt never reached the Job pod

## Fix

One-line change: set `session.Prompt = requestPrompt` instead of `agent.Prompt`.

The control plane now assembles each source exactly once:
```
project prompt → agent prompt → inbox messages → user task prompt (session.Prompt)
```

When no `--prompt` is passed, `requestPrompt` is `nil`, so `session.Prompt` is NULL and correctly skipped.

## Testing

- `go build ./...` passes for both api-server and control-plane
- Verified prompt assembly logic in `assembleInitialPrompt()` handles nil/empty correctly


jsell-rh (approved):
## 🤖 Amber Review — PR #371

**Fix is correct and well-reasoned.** The one-line change resolves both symptoms cleanly: the duplicated agent prompt and the dropped `--prompt` value.

**Root cause verified:** `buildStartPrompt()` already reads `agent.Prompt` directly (line ~264), so storing it in `session.Prompt` too caused a double-inclusion when `assembleInitialPrompt()` in the control plane fetched it again. Now `session.Prompt` carries only the user's request prompt, which the control plane was never seeing before.

**Nil path confirmed safe:** `requestPrompt` is only set when `body.Prompt != ""` after a successful decode, so it's `nil` when `--prompt` is omitted. The `session.Prompt = nil` → NULL DB value path is handled correctly by the existing `buildStartPrompt` nil-guard (`if requestPrompt != nil && *requestPrompt != ""`).

**CI:** Lint ✅ | Unit Tests ✅ | API Server Integration ✅ | Konflux builds ✅

**One note below on test coverage** — not blocking, but flagging per project TDD standards.

---
*— Amber*

jsell-rh:
**Minor note on `start_handler.go:117` — no regression test for the fixed path.**

The fix is correct and CI is green, but there's no test asserting that `session.Prompt` is set to the request body prompt rather than `agent.Prompt`. A future refactor could silently re-introduce this. Consider adding to `start_handler_test.go`:

```go
// When a request body prompt is provided, session.Prompt should be the
// request prompt, not agent.Prompt — regression for the dropped --prompt bug.
func TestStartHandler_RequestPromptStoredOnSession(t *testing.T) { ... }
```

Not blocking — flagging per Principle IV (TDD).

*— Amber*

bsquizz (approved):

PR #372, opened 2026-07-15
fix(kind): restart deployments after loading local images

bsquizz:
## Summary

- Fixes stale images when running `make kind-up LOCAL_IMAGES=true` against an existing cluster
- When `setup-kind.sh` detects the cluster already exists it exits early, then `kubectl apply` sees no spec change (image tag is still `:latest`) and doesn't trigger a rollout — pods keep running old containers
- Adds `kubectl rollout restart deployment` after `kubectl apply` in the `LOCAL_IMAGES=true` path so pods always pick up freshly built images from containerd

## Test plan

- [ ] Run `make kind-up LOCAL_IMAGES=true` on a fresh cluster — verify images match current source
- [ ] Make a source change, then run `make kind-up LOCAL_IMAGES=true` again on the same cluster — verify pods restart and pick up the new images
- [ ] Confirm `make kind-rebuild` still works independently

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -909,6 +909,8 @@ kind-up: preflight-cluster build-cli ## Start kind cluster and deploy the platfo
 		$(MAKE) --no-print-directory _kind-load-images; \
 		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Deploying with locally-built images..."; \
 		kubectl apply --validate=false -k components/manifests/overlays/kind-local/; \
+		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Restarting deployments to pick up freshly built images..."; \
+		kubectl rollout restart deployment -n $(NAMESPACE); \
 		echo "$(COLOR_BLUE)▶$(COLOR_RESET) Patching agent registry for local images..."; \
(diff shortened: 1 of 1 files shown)


jsell-rh (commented):
## 🤖 Amber Analysis — PR #372

**Fix is correct and addresses a real bug.** When `setup-kind.sh` detects an existing cluster and exits early, `kubectl apply` sees no spec change (`:latest` tag is unchanged) and doesn't trigger a rollout — this PR's `rollout restart` ensures the freshly-loaded image is actually picked up.

**One suggestion before marking ready:**

The added restart has no output suppression or rollout wait, while the existing `_restart-all` target (line 1731–1734) does `>/dev/null 2>&1` and then `kubectl rollout status --timeout=90s`. The `wait-for-ready.sh` further down covers the wait, so correctness is fine — but for consistency I'd suggest either calling `$(MAKE) _restart-all` instead of the inline `kubectl rollout restart`, or at minimum suppressing the output to match surrounding style:

```makefile
kubectl rollout restart deployment -n $(NAMESPACE) >/dev/null 2>&1; \
```

Minor ergonomic point: the restart fires before the agent registry patch, so pods restart once here and then again (via env mutation) when `kubectl set env` runs later. Not a correctness issue, just a heads-up in case you want to reorder or consolidate.

**Confidence:** High — fix logic is sound, no security or API surface changes.

*— Amber*

PR #374, opened 2026-07-16
feat(spec): add Learn view specification

JuanmaBM:
## Summary

- Adds new UI spec `specs/ui/learn.spec.md` defining a "Learn" view with two sections: **Concepts** (from `docs/src/content/docs/concepts/`) and **Examples** (from `examples/docs/`)
- Users can browse and read platform learning content directly in the UI without navigating to the source repository
- Content is static, compiled at build time from existing markdown files — no backend changes needed
- Updates `specs/ui/index.spec.md` and `specs/index.spec.md` to register the new spec

### DIFF
--- a/specs/index.spec.md
+++ b/specs/index.spec.md
@@ -52,6 +52,7 @@ Machine-readable index for autonomous reconciliation (`/reconcile` skill).
 | `ui/project-sharing.spec.md` | ui | Project sharing, ownership transfer | FE, BE | rbac-enforcement |
 | `ui/scheduled-sessions.spec.md` | ui | Schedule CRUD UI | FE | scheduled-session-execution |
 | `ui/work-tracking-dashboard.spec.md` | ui | Dashboard, work annotations | FE | annotations, views |
+| `ui/learn.spec.md` | ui | Learn catalog, static content | FE | architecture |
 | `standards/control-plane/conventions.spec.md` | standards | - | CP | - |
 | `standards/platform/cross-cutting.spec.md` | standards | - | ALL | - |
 | `standards/security/security.spec.md` | standards | - | ALL | - |
--- a/specs/ui/index.spec.md
+++ b/specs/ui/index.spec.md
@@ -19,3 +19,7 @@ Structured metadata on sessions, agents, and projects. Covers annotation schemas
 ### [Live Preview and Real-Time Updates](live-preview.spec.md)
 
 Visual feedback for running sessions including live preview, screenshot capture, real-time update streaming, and WebSocket integration.
+
+### [Learn](learn.spec.md)
+
+In-app catalog of learning resources rendered from markdown files in `examples/docs/`. Static build-time content, global scope.
--- a/specs/ui/index.spec.md
+++ b/specs/ui/index.spec.md
@@ -22,4 +22,4 @@ Visual feedback for running sessions including live preview, screenshot capture,
 
 ### [Learn](learn.spec.md)
 
-In-app catalog of learning resources rendered from markdown files in `examples/docs/`. Static build-time content, global scope.
+In-app catalog of learning resources compiled from `docs/src/content/docs/concepts/` (Concepts) and `examples/docs/` (Examples). Static build-time content, global scope.
--- a/specs/ui/learn.spec.md
+++ b/specs/ui/learn.spec.md
@@ -1,0 +1,146 @@
+# Learn Specification
+
+## Purpose
+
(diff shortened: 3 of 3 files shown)


jsell-rh (changes requested):
## 🤖 Amber Review — PR #374: Learn View Specification

Good addition — the Learn view is a natural gap in the UI and the overall design direction (static build-time, global scope, card grid) is sound. The scenario coverage is solid. A few issues need resolving before the implementation PR can start without ambiguity.

---

### [Major] Loading state requirement contradicts the static content model

The spec simultaneously requires:
1. "All learning content SHALL be statically imported at build time" — content is in the JS bundle, no async fetch
2. "skeleton card placeholders are displayed until the catalog renders" — implies an async render cycle

If content is statically bundled (e.g., via `import` or `require()` at module level), it's available synchronously — there's no loading phase to show skeletons for. This contradiction will leave the implementor guessing.

**Resolution options:**
- If content is truly synchronous (bundled JS): remove the loading state requirement entirely
- If the intent is to handle Next.js route hydration latency or lazy-loaded chunks: say so explicitly ("loading state covers SSR hydration / code-split bundle fetch for the `/learn` route")
- If content will eventually be API-backed (even if not now): flag this as a future-proofing design decision, not a current requirement

---

### [Major] Build pipeline mechanism deferred but not bounded — creates architectural ambiguity for the impl PR

The spec says:

> The mechanism (pre-build copy step, Dockerfile COPY, or content relocation) is an implementation detail.

These three options have very different trade-offs:

| Mechanism | Trade-off |
|---|---|
| Pre-build copy step | Needs `Makefile`/`package.json` changes; breaks if copy is skipped in local dev |
| Dockerfile COPY | Works in CI but requires `make kind-rebuild` to pick up new docs in local dev |
| Content relocation | Simplest at build time but may break the Astro docs site at `docs/` if paths change |

A spec that defers this entirely gives the implementation PR free rein to make a structural choice that affects CI, local dev, and another component (`docs/`). The spec should at minimum say **which mechanism is preferred** and **what trade-off is accepted**, even if the exact file paths are left to the impl PR.

---

### [Minor] `specs/ui/index.spec.md` description mentions only `examples/docs/` — incomplete

Added line:
> In-app catalog of learning resources rendered from markdown files in `examples/docs/`. Static build-time content, global scope.

But the Learn view has two content sources: `docs/src/content/docs/concepts/` (Concepts) and `examples/docs/` (Examples). The index description should read something like:

> In-app catalog of learning resources compiled from `docs/src/content/docs/concepts/` (Concepts) and `examples/docs/` (Examples). Static build-time content, global scope.

---

### [Minor] `examples/docs/` creation is asserted but not specified

The spec says:
> The `examples/docs/` directory does not currently exist in the repository. It is created as part of this feature.

But the spec doesn't state where `examples/` lives relative to the repo root, or whether any initial example files ship with the feature. The implementation PR will need to infer this. At a minimum, confirm: does `examples/` already exist at repo root? Does this feature create `examples/docs/` with at least one placeholder file to validate the pipeline, or is it left empty?

---

### What's solid ✅

- Design decisions table is complete and pre-empts the obvious questions (port/adapter exemption, no-search-initially, static over API)
- "Subdirectories are ignored" scenario — good edge case to call out explicitly
- Frontmatter extraction chain (frontmatter title → first heading → filename) is the right fallback ladder
- URL route table is clean — `{slug}` derivation is unambiguous
- Malformed markdown graceful degradation scenario is correctly specified
- Global (not project-scoped) access is clearly stated
- Registered in both `specs/index.spec.md` and `specs/ui/index.spec.md` — correct process

**Confidence:** High — issues are spec-level clarifications, not design flaws.

Resolve the loading-state contradiction and bound the build pipeline choice (even loosely) before the impl PR starts.

*— Amber*

jsell-rh (approved):
## 🤖 Amber Review — PR #374: Learn View Specification (follow-up)

All four issues from my previous review are resolved in the `fix(spec): amber recommendations` commit. Approving.

**What changed:**

- **[Major] Loading state contradiction** — Resolved. No skeleton/loading requirement exists in the updated spec. Static build-time content is correctly treated as synchronous.
- **[Major] Build pipeline mechanism** — Resolved. The preferred mechanism is now explicitly specified: Makefile pre-build copy step into `components/ambient-ui/.learn-content/`, with a Docker path (`/learn-content/`) + local dev fallback (`../../`). Trade-offs are documented.
- **[Minor] Index description** — Resolved. `specs/ui/index.spec.md` now correctly references both `docs/src/content/docs/concepts/` and `examples/docs/`.
- **[Minor] `examples/docs/` creation** — Resolved. Spec now states repo root location and requires at least one initial example file to validate the pipeline.

The spec is clear, implementation-ready, and consistent with ACP conventions. The implementation PR can start from this.

*— Amber*

PR #384, opened 2026-07-16
feat(pr-test): add OpenShift installer and e2e smoke test

markturansky:
Add self-contained scripts for deploying and validating ACP on OpenShift:

- install-openshift.sh: deploys full ACP stack (PostgreSQL, Keycloak, API server, control plane, UI) with Vertex AI LLM integration
- e2e-smoke.sh: validates the complete flow using acpctl CLI (auth → project → session → LLM inference → cleanup), 17 checks
- teardown-openshift.sh: cleans up all resources in the namespace

Validated against UAT cluster (acp-api-01): 17/17 passes including live LLM inference via Vertex AI.

### DIFF
--- a/components/pr-test/install-openshift.sh
+++ b/components/pr-test/install-openshift.sh
@@ -17,6 +17,8 @@
 #   SKIP_RBAC            Set to 1 to skip ClusterRole/ClusterRoleBinding creation
 #   SKIP_KEYCLOAK        Set to 1 to skip Keycloak deployment (use existing IdP)
 #   KEYCLOAK_REALM_URL   External Keycloak realm URL (required if SKIP_KEYCLOAK=1)
+#   KC_DEV_PASSWORD      Password for 'developer' user (default: developer)
+#   KC_ADMIN_PASSWORD    Password for 'admin' user (default: admin)
 #   DRY_RUN              Set to 1 to print manifests without applying
 set -euo pipefail
 
@@ -45,6 +47,8 @@ usage() {
   echo "  OC                  oc binary (default: oc)"
   echo "  SKIP_RBAC           Set to 1 to skip ClusterRole/ClusterRoleBinding"
   echo "  SKIP_KEYCLOAK       Set to 1 to skip Keycloak (set KEYCLOAK_REALM_URL)"
+  echo "  KC_DEV_PASSWORD     Password for 'developer' user (default: developer)"
+  echo "  KC_ADMIN_PASSWORD   Password for 'admin' user (default: admin)"
   echo "  DRY_RUN             Set to 1 to print manifests only"
   exit 1
 }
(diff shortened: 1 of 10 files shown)


components/pr-test/install-openshift.sh
```diff
@@ -0,0 +1,1186 @@
...
+# ── Auth ConfigMap (JWKS from Keycloak) ──────────────────────────────────────
+
+log "Creating auth ConfigMap"
+
+JWKS_URL="${KEYCLOAK_REALM_URL}/protocol/openid-connect/certs"
+
+apply_ns <<EOF
+apiVersion: v1
+kind: ConfigMap
+metadata:
```
jsell-rh, line 647: **🔴 Blocker — ACP Convention: `runAsNonRoot: false`**

```yaml
securityContext:
  runAsNonRoot: false  # ← violates ACP security convention
```

CLAUDE.md requires `runAsNonRoot` on all containers. Keycloak 26.x supports non-root. Fix:

```yaml
securityContext:
  runAsNonRoot: true
  runAsUser: 1000
  seccompProfile:
    type: RuntimeDefault
```

If Keycloak needs to write to `/opt/keycloak/data` at startup, add an `emptyDir` volumeMount for that path.

components/pr-test/install-openshift.sh
```diff
@@ -0,0 +1,1186 @@
...
+    --from-literal=SSO_CLIENT_SECRET="$KC_CLIENT_SECRET" \
+    --from-literal=SSO_AUDIENCE=ambient-frontend \
+    --from-literal=SESSION_SECRET="$SESSION_SECRET" \
+    --from-literal=SSO_REDIRECT_URI="$SSO_REDIRECT_URI"
+  ok "Created sso-credentials secret"
+else
+  ok "Secret exists: sso-credentials"
+fi
+
+# ── Auth ConfigMap (JWKS from Keycloak) ──────────────────────────────────────
```
jsell-rh, line 638: **🔴 Critical — Hardcoded Keycloak admin credentials**

```yaml
- name: KEYCLOAK_ADMIN
  value: admin
- name: KEYCLOAK_ADMIN_PASSWORD
  value: admin   # ← exposed on shared clusters
```

The Keycloak Admin UI (`/admin/ambient-code/console/`) is reachable at the route URL. On a shared UAT/PR-test cluster, anyone with cluster access can log in as admin and issue tokens for any user.

Fix — generate dynamically and pass via Secret:
```bash
KC_ADMIN_PASS=$(python3 -c "import secrets; print(secrets.token_urlsafe(24))")
# Create or update a keycloak-admin-secret, then reference via secretKeyRef
```

Or at minimum document this as intentional dev-only behavior with a warning in the script header.

components/pr-test/install-openshift.sh
```diff
@@ -0,0 +1,1186 @@
...
+    app: ambient-control-plane
+  ports:
+  - name: token
+    port: 8080
+    targetPort: 8080
+    protocol: TCP
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
jsell-rh, line 886: **🟡 Major — Missing `readOnlyRootFilesystem: true`**

The UI and control plane containers both have `readOnlyRootFilesystem: true`. The API server container is missing it. For consistency with ACP conventions, either:

1. Add `readOnlyRootFilesystem: true` here (if the binary doesn't write to disk at runtime), or
2. Add an `emptyDir` mount for `/tmp` and any other writable paths, then enable it

The UI already shows the correct pattern with its `next-cache` and `tmp` emptyDir mounts.

jsell-rh (changes requested):
Requesting changes on two blockers: Keycloak `runAsNonRoot: false` (ACP security convention violation) and hardcoded admin credentials (security risk on shared clusters). One major finding on missing `readOnlyRootFilesystem` for the API server. Everything else looks solid — great work on the tooling overall. Happy to re-review once these are addressed. — Amber

PR #385, opened 2026-07-16
spec: OpenShell CLI e2e test and demo script

squizzi:
## Summary

- Adds `specs/platform/openshell-cli-e2e-test.spec.md` defining the desired behavior for an OpenShell CLI-native e2e test that validates all core `openshell` commands (sandbox, provider, policy, settings) against ACP-managed gateways — the direct CLI path power users follow.
- Adds `tests/e2e/fixtures/openshell-cli-test/test-policy.yaml` as a minimal sandbox policy fixture for policy operations testing.
- Includes a companion `demo-openshell.sh` spec following the `demo-*` tmux-based pattern for live demonstrations.
- Uses `acpctl gateway setup-cli` for gateway connectivity (not manual mTLS cert extraction).

Ref: [ENGPROD-10199](https://redhat.atlassian.net/browse/ENGPROD-10199)

## Test plan

- [ ] Spec reviewed for completeness against all `openshell` CLI command groups
- [ ] Policy fixture validates with `openshell policy set --file`
- [ ] Scenarios aligned with upstream OpenShell e2e tests (smoke, provider, policy, settings)

### DIFF
--- a/specs/platform/openshell-cli-e2e-test.spec.md
+++ b/specs/platform/openshell-cli-e2e-test.spec.md
@@ -1,0 +1,433 @@
+# OpenShell CLI E2E Test Specification
+
+**Date:** 2026-07-15
+**Status:** Design
+**Related:** `openshell-sandbox-provisioning.spec.md` — gateway sandbox provisioning; `e2e-test-tooling.spec.md` — mock LLM infrastructure; [ENGPROD-10199](https://redhat.atlassian.net/browse/ENGPROD-10199) — issue
+**Skill:** `skills/build/full-stack-pipeline/` — wave-based implementation pipeline
+
+---
+
+## Purpose
+
+Power users who interact with ACP-managed OpenShell gateways via the native `openshell` CLI need a dedicated e2e test validating that all core CLI commands work against ACP-provisioned gateways. Existing e2e tests (`gateway-e2e-test.sh`, `openshell-dual-tenant.sh`) exercise OpenShell integration through the API server and `acpctl`. There is no test that validates the direct `openshell` CLI path — the exact workflow a power user would follow in the "OpenShell as a Service" story.
+
+This spec defines an e2e test script that authenticates to an ACP-managed OpenShell gateway using the `openshell` CLI and exercises the sandbox, provider, policy, and settings command groups. The test validates that resources created via `openshell` CLI are visible and consistent, and that the gateway deployed by ACP behaves identically to a standalone OpenShell deployment from the CLI's perspective.
+
+### Scope
+
+- New e2e test script at `tests/e2e/openshell-cli-e2e.sh`
+- Demo script at `components/ambient-cli/demo-openshell.sh` following the `demo-*` pattern
+- CLI-only testing: sandbox ops, provider ops, policy ops, settings ops
+- Gateway connectivity setup via `acpctl gateway setup-cli` (validates the `acpctl` → `openshell` CLI handoff)
+- Cross-validation between `openshell` CLI state and ACP API state
+- Test fixtures at `tests/e2e/fixtures/openshell-cli-test/`
+
+### Out of Scope
+
+- Testing the `openshell` CLI build/installation — the binary is assumed available
+- Testing OpenShell gateway deployment — that is covered by gateway provisioning specs
+- Inference routing validation (already covered by `gateway-e2e-test.sh` mock LLM flow)
+
+Some overlap with `gateway-e2e-test.sh` is acceptable when the coverage is `openshell`-specific — for example, verifying that a policy applied via `openshell policy set` actually enforces network restrictions inside the sandbox. The distinction is that this test exercises the `openshell` CLI path, not the `acpctl`/API path.
+
+### Dependencies
+
+- `openshell` CLI binary available in the test environment (CI installs it via `local-dev-simulation`; local runs SHALL error if missing)
+- `acpctl` CLI built and available (`make build-cli`)
+- Kind cluster provisioned via `make kind-up` with `OPENSHELL_USE_GATEWAY=true`
+
+---
+
+## Requirements
+
+### Requirement: Test Script Structure
+
+The test SHALL be a self-contained bash script at `tests/e2e/openshell-cli-e2e.sh` that follows the conventions established by `gateway-e2e-test.sh` and `openshell-dual-tenant.sh`.
+
+The script SHALL:
+- Use `set -euo pipefail` for strict error handling
+- Implement `pass()`, `fail()`, `skip()`, and `section()` helper functions with pass/fail counters
+- Accept `--skip-cleanup` flag for post-mortem inspection
+- Accept an optional `API_URL` positional argument (default: `http://localhost:<port-forward>`)
+- Set up a port-forward to the API server
+- Register the gateway with the `openshell` CLI via `acpctl gateway setup-cli` (validates the `acpctl`-to-`openshell` handoff path)
+- Report per-section results with a final pass/fail summary
+- Exit non-zero if any test fails
+- Clean up all resources created during the test, even on failure (via trap)
+
+The test SHALL organize assertions into numbered sections by command category:
+1. Prerequisites (CLI availability, token, port-forwards)
+2. Gateway connectivity (registration, health check)
+3. Sandbox operations (create, list, get, delete)
+4. Provider operations (create, get, list, delete)
+5. Policy operations (set, get, list, delete)
+6. Settings operations (get, set, delete)
+7. Cross-validation (CLI-created resources visible via ACP API)
+8. Cleanup
+
+#### Scenario: Test runs on a fresh Kind cluster
+
+- GIVEN `make kind-up` has completed with `OPENSHELL_USE_GATEWAY=true`
+- AND the `openshell` CLI is installed and available in `$PATH`
+- WHEN `./tests/e2e/openshell-cli-e2e.sh` is executed
+- THEN the test SHALL complete with all sections reporting pass or skip (no fail)
+- AND all created resources SHALL be cleaned up
+
+#### Scenario: Test is idempotent
+
+- GIVEN a Kind cluster where the test has already run
+- WHEN `./tests/e2e/openshell-cli-e2e.sh` is executed again
+- THEN the test SHALL pass without errors from leftover state
+
+#### Scenario: openshell CLI missing — error in local, install in CI
+
+- GIVEN the `openshell` CLI is not installed
+- WHEN the test script starts
+- THEN it SHALL print a clear message indicating the CLI is required and exit with a non-zero code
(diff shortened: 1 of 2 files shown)


specs/platform/openshell-cli-e2e-test.spec.md
```diff
@@ -0,0 +1,433 @@
...
+
+- GIVEN `make kind-up` has completed with `OPENSHELL_USE_GATEWAY=true`
+- AND the `openshell` CLI is installed and available in `$PATH`
+- WHEN `./tests/e2e/openshell-cli-e2e.sh` is executed
+- THEN the test SHALL complete with all sections reporting pass or skip (no fail)
+- AND all created resources SHALL be cleaned up
+
+#### Scenario: Test is idempotent
+
+- GIVEN a Kind cluster where the test has already run
```
jsell-rh, line 78: **[Major] `kubectl` cross-validation makes the test non-portable — contradicts the spec's stated scope**

Section 7 (Cross-Validation) requires `kubectl get sandboxes` and `kubectl get pods` — direct cluster access — while the Scope section says this is a CLI-only test exercising the `openshell` CLI path.

Using `kubectl` in the e2e test:
1. Requires the tester to have cluster-admin (or at least namespace-get) on the Kind cluster, which isn't guaranteed for the "power user" persona this test represents
2. Couples the test to Kubernetes API internals — a pass/fail in the `openshell` CLI path could be masked or confused by cluster-level assertions
3. `openshell sandbox list` already returns phase state; a separate `kubectl get sandboxes` assertion adds noise without unique signal beyond what the CLI already exposes

**Suggested resolution:** Either (a) remove Section 7 cross-validation from this spec and add it as a separate cluster-state validation script (like `openshell-dual-tenant.sh` does), or (b) demote cross-validation to an optional `--cluster-validate` flag so the core test runs without cluster access. The ACP API cross-validation (via `acpctl`) is more in-scope than raw `kubectl`.

specs/platform/openshell-cli-e2e-test.spec.md
```diff
@@ -0,0 +1,433 @@
...
+
+- `openshell` CLI binary available in the test environment (CI installs it via `local-dev-simulation`; local runs SHALL error if missing)
+- `acpctl` CLI built and available (`make build-cli`)
+- Kind cluster provisioned via `make kind-up` with `OPENSHELL_USE_GATEWAY=true`
+
+---
+
+## Requirements
+
+### Requirement: Test Script Structure
```
jsell-rh, line 43: **[Major] `acpctl gateway setup-cli` command signature is spec'd here but the implementation is not verified against what the CLI actually accepts**

The spec defines a precise invocation:
```
acpctl gateway setup-cli <gateway-name> --gateway-url <url>
```

But the Implementation Notes table lists `acpctl gateway setup-cli` as coming from `components/ambient-cli/cmd/acpctl/gateway/setup.go`, and the PR doesn't show that file exists or has this signature. If the command doesn't exist yet (or uses different flags), the test script will fail at the first step.

The spec should either:
- Confirm the current command signature by referencing the existing CLI source (link the file), or
- Explicitly note "this command does not yet exist and must be implemented as part of this work" and add it to the Scope section

This is the critical path for the entire test — if the gateway registration step fails, sections 3–7 all skip. The spec should be unambiguous about whether this command is implemented today.

specs/platform/openshell-cli-e2e-test.spec.md
```diff
@@ -0,0 +1,433 @@
...
+#### Scenario: Policy enforcement — allowed endpoint reachable
+
+- GIVEN a policy has been set on sandbox `<sandbox-name>` allowing `update.code.visualstudio.com:443` for `/usr/bin/curl`
+- WHEN the test runs `openshell sandbox exec --gateway tenant-a --name <sandbox-name> -- curl -s https://update.code.visualstudio.com`
+- THEN the request SHALL succeed (HTTP response received, no `policy_denied`)
+
+#### Scenario: Policy enforcement — blocked endpoint denied
+
+- GIVEN a policy has been set on sandbox `<sandbox-name>` allowing only `update.code.visualstudio.com:443`
+- WHEN the test runs `openshell sandbox exec --gateway tenant-a --name <sandbox-name> -- curl http://example.com`
```
jsell-rh, line 237: **[Minor] Policy enforcement scenarios require a running, egress-proxy-enabled sandbox — not guaranteed on Kind**

The policy enforcement scenarios (allowed/blocked endpoint) require the sandbox to actually enforce network policy via the egress proxy. On a local Kind cluster, this depends on whether the OpenShell egress proxy is fully configured with the correct iptables/eBPF rules in the sandbox network namespace.

The spec should note: if egress proxy enforcement is not available in the Kind environment (e.g., limited node kernel support), these scenarios should be `skip()`'d rather than `fail()`'d. The existing pattern in `gateway-e2e-test.sh` uses skip conditions for environment-dependent checks — align with that pattern here.

specs/platform/openshell-cli-e2e-test.spec.md
```diff
@@ -0,0 +1,433 @@
...
+- WHEN the test queries `kubectl get sandboxes -n tenant-a`
+- THEN the sandbox SHALL appear as a Sandbox CRD resource
+- AND its phase SHALL match the `openshell sandbox get` output
+
+#### Scenario: CLI-created sandbox pod visible in namespace
+
+- GIVEN a sandbox is in READY phase (created via `openshell sandbox create`)
+- WHEN the test queries `kubectl get pods -n tenant-a`
+- THEN a pod corresponding to the sandbox SHALL exist and be in Running phase
+
```
jsell-rh, line 308: **[Minor] Demo script location conflicts with established pattern**

The spec places the demo at `components/ambient-cli/demo-openshell.sh`, but the existing demo scripts (`demo-kind.sh`, `demo-remote.sh`, `demo-github.sh`) are referenced in the PR body as the established pattern. The spec should confirm: does `demo-kind.sh` live in `components/ambient-cli/`? If the pattern is actually `scripts/demos/` or repo root, placing `demo-openshell.sh` in `components/ambient-cli/` breaks discoverability.

Check `ls components/ambient-cli/` or grep for `demo-kind.sh` to confirm the correct parent directory before the impl PR creates the file.

jsell-rh (changes requested):
## 🤖 Amber Review — PR #385: OpenShell CLI E2E Test Spec

Spec-only draft. Overall direction is solid — a dedicated CLI-path e2e test for `openshell` fills a real gap, the section structure is clean, and the `acpctl gateway setup-cli` handoff as the entry point is the right design choice.

Two major issues need resolving before the implementation PR starts:

**1. `acpctl gateway setup-cli` existence is unverified** — this command is on the critical path for the entire test. If it doesn't exist yet or has a different signature, every section after gateway registration will skip. The spec needs to explicitly state whether this command exists today or is a prerequisite to implement.

**2. `kubectl` cross-validation scope creep** — Section 7 requires direct cluster access, which contradicts the spec's stated CLI-only scope and the "power user" persona. Either remove it or make it opt-in behind a `--cluster-validate` flag.

Two minor issues are inline (policy enforcement skip conditions on Kind, demo script location).

---

### What's solid ✅
- `set -euo pipefail` + pass/fail/skip/section helpers — correct pattern
- `--skip-cleanup` flag + trap-based cleanup — matches existing e2e conventions
- 8-section structure mirrors the command groups cleanly
- 120s sandbox readiness poll with 2s interval and diagnostic failure message — right approach
- Cleanup failures are non-fatal (logged, not fail) — correct
- `OPENSHELL_RUNNER_IMAGE` reuse when already loaded into Kind — pragmatic
- Upstream e2e reference table (Rust tests → ACP section mapping) — very useful for implementors
- Fixture policy file is complete and well-structured

**Confidence:** High — issues are spec-level clarifications, not design flaws. Happy to re-review when the two major items are addressed.

*— Amber*

specs/platform/openshell-cli-e2e-test.spec.md
squizzi: We're able to test this already in gateway-e2e-test with a simplified policy, I feel we can do the same here.

specs/platform/openshell-cli-e2e-test.spec.md
squizzi: Good call — `kubectl` assertions break the CLI-only persona this test represents. Fixed: cross-validation is now gated behind `--cluster-validate` flag (skipped by default). The section is renamed to "Cross-Validation with Cluster State (Optional)" with a new scenario explicitly covering the skip-by-default behavior. The "ACP-created providers visible via CLI" scenario stays ungated since it uses `openshell provider list` (CLI-only).

specs/platform/openshell-cli-e2e-test.spec.md
squizzi: Verified — the command exists at `components/ambient-cli/cmd/acpctl/gateway/setup.go`. The actual signature is:

```
acpctl gateway setup-cli [name] --gateway-url <url> [--project <namespace>] [--print]
```

The spec was using a slightly different invocation (`<gateway-name>` positional + no `--project`). Fixed: updated all invocations to `acpctl gateway setup-cli --gateway-url <url> --project tenant-a`, added an "Implementation status" callout block with the verified signature and source link, and updated the command reference table.

specs/platform/openshell-cli-e2e-test.spec.md
squizzi: Confirmed — `components/ambient-cli/` is the correct location. All existing demo scripts live there:

```
./components/ambient-cli/demo-remote.sh
./components/ambient-cli/demo-kind.sh
./components/ambient-cli/demo-github.sh
```

No change needed.

jsell-rh (approved):
## 🤖 Amber Review — PR #385 Follow-up (commit `40b550a`)

Both major blockers addressed. Approving.

**`acpctl gateway setup-cli` signature verified** — the updated spec correctly documents the actual signature (`[name] --gateway-url <url> [--project <ns>] [--print]`), adds the "Implementation status" callout with a source link, and updates all invocations to match. This is exactly what was needed.

**Cross-validation gated behind `--cluster-validate`** — Section 7 is now "Cross-Validation with Cluster State (Optional)" with an explicit "skip by default" scenario. The `openshell provider list` assertion (CLI-only) stays ungated — correct distinction.

**Policy enforcement minor** — accepting the author's rationale. If `gateway-e2e-test.sh` already exercises the egress proxy enforcement path with a simplified policy and this test uses the same Kind environment, the skip condition isn't needed here. The pattern is consistent.

**Demo script location** — confirmed `components/ambient-cli/` is correct, no change needed.

The spec is implementation-ready. The two open questions (policy enforcement environment dependency, demo location) are resolved by author confirmation. Unmarking change-requested.

*— Amber*

jsell-rh:
Upgrading label from `amber/change-requested` → `amber/approved` (both major blockers addressed in commit `40b550a`).

PR #388, opened 2026-07-17
feat: gateway defaults (route, OIDC, labels) and lifecycle CLI commands

bsquizz:
## Summary

- Add server-side defaults for `route` (enabled by default), `oidc`, and `labels` on gateway creation
- Fix `PresentGateway` to include `labels` and `annotations` in API responses (were stored in DB but silently dropped)
- Add `--route-host` CLI flag to `acpctl create gateway`
- Define `acpctl create gateway` and `acpctl delete gateway` commands following the existing CLI pattern (pure HTTP REST with flag-to-field mapping)
- Every Gateway data model attribute exposed as a CLI flag — zero flags required when project is set via `acpctl project <name>`
- `OIDC_ISSUER_URL` is a required env var on the API server; CLI shows default as "set by platform"
- Define `acpctl project create` (with namespace pre-existence validation) and `acpctl project delete`
- Fix: `acpctl create agent` now falls back to configured project context, matching `acpctl create session`
- Add "OpenShell as a Service" section to README documenting the self-service gateway provisioning flow (project creation → gateway request → control plane reconciliation → sandbox usage), with a pointer to the gateway lifecycle demo script and OpenShift Local POC

> **Depends on #370** — requires PR #370 (gateway route exposure, GRPCRoute, data model additions) to merge first.

## Changes

| File | What |
|------|------|
| `components/ambient-api-server/plugins/gateways/handler.go` | Add `route: {}` default in `applyGatewayDefaults` |
| `components/ambient-api-server/plugins/gateways/presenter.go` | Fix `PresentGateway` to include `labels` and `annotations` |
| `components/ambient-cli/cmd/acpctl/create/cmd.go` | Add `--route-host` flag; fix `createAgent` project context fallback |
| `README.md` | Add "OpenShell as a Service" section describing self-service gateway provisioning |
| `specs/cli/project-gateway-lifecycle.spec.md` | New spec with route in flags, defaults table, and example YAML |

## Test plan

- [ ] `acpctl create gateway --project <p>` returns gateway with `route: {}`, `oidc`, and `labels` populated
- [ ] `acpctl get gateway --project <p> -o json` includes `labels` and `annotations` fields
- [ ] `acpctl create gateway --route-host custom.example.com` sets explicit route hostname
- [ ] `acpctl create agent --name foo` works when project is set via `acpctl project`
- [ ] Review spec scenarios against Gateway data model


jsell-rh (approved):
🤖 **Amber Review**

Spec-only PR defining `acpctl project create/delete` and `acpctl gateway create/delete` lifecycle commands.

**What this does:** Adds `specs/cli/project-gateway-lifecycle.spec.md` — 314-line GIVEN/WHEN/THEN spec with server-side defaults table, RBAC requirements (admin-only), environment variable config, and migration impact table. No code changes.

**Assessment:** ✅ Approve

The spec is well-structured and internally consistent. Specific things I verified:

- **Backward compat claim is sound:** The note that `acpctl create project` (old alias) skips namespace validation while `acpctl project create` (new canonical) does validate is correctly scoped — the reconciler's `ensureNamespace()` is unchanged.
- **Server-side defaulting:** Env var approach (`GATEWAY_IMAGE`, `KEYCLOAK_REALM_URL`) follows the "separate config from code" convention from CLAUDE.md. Kind and CRC default values are both present.
- **RBAC gate:** Admin-only authorization for all lifecycle commands is appropriate — these operations affect cluster-level infrastructure. RBAC scenarios cover all four commands.
- **Error messages:** Specific, actionable error text (e.g., `"acpctl project create --name team-alpha"` in the project-not-found error for gateway create) — follows ACP UX conventions.
- **One-gateway-per-project constraint:** Explicitly specified in the gateway create error scenario. Good.
- **Cross-references:** Related specs linked in the header (`gateway-cli.spec.md`, `data-model.spec.md`, `gateway-provisioning.spec.md`, etc.).

**Minor observations (no blockers):**
- The `Configuration > CLI Reference Update` table marks all four commands as `🔲 planned` — correct for a spec PR, just ensures implementers know this is all net-new.
- The `Environment Variables` table lists the same `GATEWAY_IMAGE` value for both Kind and CRC (`ghcr.io/nvidia/openshell/gateway:0.0.80`). This appears intentional (same image, different OIDC URL). Fine for now but implementers should note that env var values will diverge as versions update.

**Confidence:** [95%] High — pure spec, no code risk. Constitution compliance not applicable (no production code).

*— Amber*

jsell-rh:
🏷️ Adding label `amber/approved`.

*— Amber*

bsquizz:
Here's what acpctl create gateway --help would look like:

```
  Create an OpenShell gateway in a project.

  Equivalent to 'acpctl apply -f' with a Gateway kind YAML — both HTTP POST
  to /projects/{p}/gateways. No flags are required when a project is set
  via 'acpctl project <name>'.

  Usage:
    acpctl create gateway [flags]

  Examples:
    # Create with all defaults (project from context, name auto-generated)
    acpctl project team-alpha
    acpctl create gateway

    # Create with explicit project and name
    acpctl create gateway --project team-alpha --name my-gw

    # Override OIDC issuer
    acpctl create gateway --oidc-issuer https://keycloak.example.com/realms/myrealm

    # Full explicit configuration
    acpctl create gateway --project team-alpha --name my-gw \
      --image ghcr.io/nvidia/openshell/gateway:0.0.85 \
      --server-dns-names gw.example.com,gw.internal.local \
      --oidc-issuer https://sso.example.com/realms/prod \
      --labels env=prod,team=platform

  Flags:
        --name string               Gateway name (default: auto-generated)
        --project string            Project to deploy into (default: configured project)
        --image string              Gateway container image (default: set by platform)
        --server-dns-names string   Comma-separated DNS names for TLS cert generation (default: derived from name + project)
        --config string             OpenShell gateway TOML configuration
        --labels string             Key=value pairs, comma-separated
        --annotations string        Key=value pairs, comma-separated
        --oidc-issuer string        OIDC issuer URL (default: set by platform)
        --oidc-audience string      Expected aud claim in JWT (default "openshell-cli")
        --oidc-jwks-ttl int         JWKS key cache retention in seconds (default 3600)
        --oidc-roles-claim string   Dot-delimited path to roles in JWT (default "realm_access.roles")
        --oidc-admin-role string    Role name for admin access (default "openshell-admin")
        --oidc-user-role string     Role name for user access (default "openshell-user")
        --oidc-scopes-claim string  Dot-delimited path to scopes in JWT
    -o, --output string             Output format: json
    -h, --help                      help for gateway

  Global Flags:
    -U, --api-url string   ACP API server URL
        --config string    Config file (default ~/.acpctl.json)
```

jsell-rh (approved):
## 🤖 Amber — PR #388 Round 2 (commit `460ebb09`)

One new commit since my approval at `f41b6d23`. **Approval stands** — and this commit actually improves the spec meaningfully.

---

### What changed and the verdict

**`specs/cli/project-gateway-lifecycle.spec.md` — spec revision per review feedback**

This is a substantial revision (109 add / 68 del), and all of it is improvements:

- **Command surface: `acpctl gateway create/delete` → `acpctl create/delete gateway`** — aligns with the existing `acpctl create session`, `acpctl delete project` pattern. Pure HTTP REST with flag-to-field mapping, no bespoke orchestration. This is architecturally cleaner and consistent. ✅

- **Full flag coverage** — every Gateway data model attribute now has a flag (`--image`, `--server-dns-names`, `--config`, `--labels`, `--annotations`, all OIDC sub-fields). The "zero flags required when project is set via context" design is correct — mirrors `acpctl create session`. ✅

- **`KEYCLOAK_REALM_URL` → `OIDC_ISSUER_URL`** — better name (describes the role, not the implementation). The production default of RH SSO (env var absent) vs. local override (env var set) is the right design for a managed platform. The environment table now has a `Production` column showing `_(not set — RH SSO default)_`. Clear. ✅

- **`acpctl delete gateway <name>` with positional arg** — consistent with `acpctl delete session <name>`. The "No gateway in project" error scenario is replaced with a more precise "Gateway not found: `gateway \"nonexistent\" not found`". ✅

- **"create gateway is equivalent to apply" scenario** — explicitly calling this out in the spec is valuable. Implementers won't add bespoke orchestration if they know it must behave identically to a plain apply. ✅

---

**`components/ambient-cli/cmd/acpctl/create/cmd.go` — real bug fix (15 add / 5 del)**

This is not a spec-only change. The `createAgent` function was `return fmt.Errorf("--project is required")` before the project context check — meaning `acpctl project <name>` context was ignored even though `acpctl create session` already handled this correctly.

The fix is straightforward and correct:
1. Load config to get the project context
2. Check `cmd.Flags().Changed("project")` to prefer the explicit flag
3. Fall back to `cfg.GetProject()`
4. Only error if both are empty, with the message `"no project set; use --project or run 'acpctl project <name>' first"` ✅

`fmt.Errorf("load config: %w", err)` — correct wrapping. ✅

The fix mirrors the exact pattern already used in `createSession()` — consistent. Good catch including this in the same PR.

---

One minor observation (non-blocking): the spec still shows `--image` default as "Server default (`GATEWAY_IMAGE` env var)" — but the PR doesn't change the API server's gateway create handler to apply this default. This is fine for a spec PR; just flagging it so the implementation PR author knows they need to wire up the server-side defaulting for `image` when the flag is absent.

No convention violations. Real bug fixed, spec tightened.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #388 Round 3 (commits `d424f299`, `5aa12c88`, `da414bc4`)

Three new commits since my Round 2 approval at `460ebb09`. **Approval stands** — and the implementation commit is the bulk of the work.

---

### Spec commits (`d424f299`, `5aa12c88`) — clean

`OIDC_ISSUER_URL` promoted from optional override to required deployment config, with explicit per-environment value table. Correct call — this makes it clear what every operator must configure. `--oidc-issuer default: "set by platform"` is accurate and won't confuse CLI users. ✅

---

### Implementation commit (`da414bc4`) — solid overall

**`handler.go` — server-side defaulting**

- `applyGatewayDefaults`: all fields use nil/empty check before overwriting — idiomatic, won't trample explicit zero-value requests. ✅
- `CheckAdminTier` on both Create and Delete — matches spec requirement. ✅
- `ServerDnsNames` default: name is set first, then DNS derived from it — field order dependency is satisfied. ✅

**`create/cmd.go` — `createGateway`**

- Config-context-aware project resolution (same pattern as `createAgent` fix in Round 2). ✅
- `cmd.Flags().Changed(f)` guard for every OIDC field — correct: prevents sending zero-value overrides when user didn't set the flag. ✅
- `parseKeyValuePairs`: clean, handles empty pairs gracefully. ✅
- `errors.As(err, &apiErr) && apiErr.StatusCode == 404` — correct 404 detection; good UX message. ✅

**`delete/cmd.go` — `--project` override**

- Factory-based client creation on `--project` flag change is consistent with existing patterns. Only creates the alternate client when flag is explicitly set — won't break the default-project path. ✅

**`project/cmd.go` — `acpctl project create/delete`**

- `exec.Command("kubectl", "get", "namespace", ...)` for namespace pre-existence check: works. Minor non-blocking: if kubectl returns non-zero for reasons other than "namespace not found" (RBAC denial, network timeout), the error message says "namespace does not exist" when it might not. Fine for first implementation.

**Manifests (kind, hcmais, hcmais-dev, template-services.yaml)**

- `GATEWAY_IMAGE` and `OIDC_ISSUER_URL` added to all four environments — consistent. ✅
- `template-services.yaml` adds `GATEWAY_IMAGE` as `required: true`. ✅

---

### Non-blocking observation

`DefaultGatewayImage` and `DefaultOIDCIssuerURL` are package-level vars initialized at startup via env calls. This means unit tests for `applyGatewayDefaults` must set env vars before `init()` runs. Worth keeping in mind when adding tests.

---

No convention violations. No security issues. Full spec implementation.

**Confidence:** [95%] High — reviewed full diff against spec requirements.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #388 Round 5 (commits `d63fb1ec`, `da29317f`)

Two new commits since my Round 4 approval at `da414bc4`. **Approval stands.** Both commits are correct; two non-blocking observations worth noting before merge.

---

### `d63fb1ec` — remove hardcoded UIDs from gateway StatefulSet for OpenShift SCC compatibility

Removing `fsGroup: 1000` and `runAsUser: 1000` is the right call for OpenShift — `restricted-v2` assigns UIDs from the namespace's allocated range. The remaining `securityContext` (`runAsNonRoot: true`, `allowPrivilegeEscalation: false`, `drop: ALL`, `readOnlyRootFilesystem: true`) is unchanged and correct. ✅

**Non-blocking observation:** `oc adm policy add-scc-to-user privileged -z openshell-gateway-sandbox` in `crc-up` grants the most permissive SCC available. The pod spec itself is still restrictive, so the privilege isn't exercised — but `anyuid` would be sufficient here (it allows OpenShift to assign the UID without requiring `privileged`). Fine for CRC/dev since this is `crc-up` only and `|| true` prevents CI breakage, but worth tightening to `anyuid` in a follow-up if openshell doesn't actually need true privilege escalation.

---

### `da29317f` — labels/annotations from JSON strings to `map[string]string`

This is the right API design. The prior `type: string` (JSON-encoded) was an internal implementation detail leaking into the API surface. ✅

**What I verified:**

- **DB bridge is correct:** Internal `Gateway` model still uses `*string` (JSON blob in PostgreSQL). `ConvertGateway` marshals API `map[string]string` → JSON string for storage; `PresentGateway` unmarshals JSON string → `map[string]string` for responses. Round-trip is sound. ✅
- **`applyGatewayDefaults` marshals then stores:** Labels default is marshaled to `*string` inline — consistent with the internal model type. ✅
- **Patch handler re-marshals correctly:** `patch.Labels` is now `*map[string]string` from the OpenAPI model; `json.Marshal` on that produces the JSON string for DB storage. ✅
- **Reconciler simplification:** `updateGatewayAnnotation` now directly mutates `gw.Annotations` as `map[string]string` and passes the map to `Update()` — eliminates the double-marshal that was there before. ✅
- **SDK regeneration:** All three SDKs (Go, Python, TypeScript) and the OpenAPI docs updated consistently. ✅
- **`apply/cmd.go`:** `marshalStringMap()` call correctly removed; `doc.Labels` is already `map[string]string`. ✅
- **`examples/overlays/tenant-c/gateway.yaml`:** Image bumped to `0.0.83` — consistent with the default in `handler.go`. ✅
- **Breaking change:** This is a breaking API change (labels/annotations type changes). Existing clients that send `labels: "{\"key\":\"val\"}"` (JSON-encoded string) will get a 400. Since the SDK is regenerated and this is still a draft PR with no stable release, the blast radius is limited to anyone using the old SDK directly. Worth a note in the PR description but not a blocker.

**Non-blocking observation:** The default labels in `applyGatewayDefaults` include `"env": "dev"`. This default runs on all environments (kind, hcmais, CRC, production). An `env: dev` label on production gateways is noise at best, confusing at worst. Consider removing it from the server-side default or deriving it from an env var — or just default to `purpose: openshell` only and let operators set their own env labels.

---

No convention violations. No new security issues. Clean implementation of the spec.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #388 Round 6 (commit `8ddf5ce9`)

One new commit since my Round 5 approval at `da29317f`. **Approval stands.**

---

### `8ddf5ce9` — `components/ambient-cli/demo-gateway-lifecycle.sh` (163 lines, new file)

A tmux-based interactive demo for the gateway lifecycle CLI commands. Scripting quality is solid:

- `set -euo pipefail` ✅
- `type_text` + `type_and_run` helpers correctly use `tmux send-keys -l` for literal character-by-character input — prevents tmux from interpreting special chars ✅
- `read -rsn1` keypress gating between steps — correct interactive demo pattern ✅
- Cleanup preamble: `tmux kill-session -t "$SESSION" 2>/dev/null || true` ✅
- Pre-flight cleanup: `openshell gateway remove tenant-c-openshell-gateway 2>/dev/null || true` — allows re-runs without manual teardown ✅
- `set -euo pipefail` on a script using `sleep` and tmux sends is fine — none of those fail under normal conditions ✅

**One non-blocking note:** Step 10 references `acpctl gateway setup-cli --project tenant-c --print` and step 11 references `openshell gateway add --name tenant-c-openshell-gateway --oidc-issuer ...`. The `setup-cli` subcommand isn't defined in this PR's spec — this may be a placeholder pointing to a future command, or something in PR #363/#388's scope. Worth verifying the command exists before the demo is run live; otherwise step 10 will fail visibly during the demo.

No security issues — no credentials hardcoded, no token logging. Demo-only script, not wired into CI.

*— Amber*

markturansky:
this PR ported into https://github.com/openshift-online/agent-control-plane/pull/404, worked, and merged.

PR #389, opened 2026-07-17
fix(policy): rename filesystem to filesystem_policy for openshell compat

squizzi:
## Summary

Upstream openshell expects the sandbox filesystem policy key to be `filesystem_policy`, but we were using `filesystem`. This renames the key across all policy YAMLs, the Rego engine, example policies, test fixtures, OpenAPI description, and spec docs. A new DB migration (`202607170100`) restores the upstream-compatible key name in stored JSONB specs without modifying the prior migration that already ran on deployed instances.

## Changes

- Renamed `filesystem` → `filesystem_policy` in runner `policy.yaml`, all example policies, and spec doc YAML examples
- Updated Rego rule to read `data.filesystem_policy` instead of `data.filesystem`
- Added `restoreFilesystemPolicyKey()` DB migration to rename stored specs back
- Updated test fixtures in go-sdk and control-plane to use `filesystem_policy`
- Updated OpenAPI description and spec/reconcile docs

## Test plan

- [x] Verify `go vet ./...` and `golangci-lint run` pass in api-server and control-plane
- [x] Run `cd components/ambient-sdk/go-sdk && go test ./...`
- [x] Run `cd components/ambient-control-plane && go test ./...`
- [x] Deploy to dev cluster and verify sandbox policies load correctly with `filesystem_policy` key


jsell-rh (approved):
## 🤖 Amber Review — PR #389: rename filesystem → filesystem_policy

Clean, well-scoped fix. Reviewed against the ACP checklists.

**What this does:** Aligns the policy key name with upstream openshell's expected `filesystem_policy` schema key by renaming across all 17 files: policy YAMLs, Rego engine, example policies, test fixtures, specs, OpenAPI docs, and a compensating DB migration.

---

### ✅ Checklist pass

- **Migration correctness** — `restoreFilesystemPolicyKey()` (202607170100) correctly inverts the prior `renameFilesystemPolicyKey()` (202607080100). The SQL is safe: uses `spec ? 'filesystem'` existence check before touching rows, handles `NULL spec`, skips soft-deleted rows. Rollback function reverses cleanly. ✅
- **Rego engine** — `filesystem_policy := data.filesystem_policy` is the correct fix; the prior `data.filesystem` was reading the old DB-mangled key. ✅
- **Go test fixtures** — `provider_reconciler_test.go` and `go-sdk/types/policy_extensions_test.go` updated consistently. ✅
- **Example policies** — all three (`locked-down.yaml`, `mock-llm-permissive.yaml`, `permissive.yaml`) updated. ✅
- **Spec/doc updates** — `RECONCILE.md`, `data-model.spec.md`, `agent-sandbox-config.spec.md`, `e2e-test-tooling.spec.md` all use correct language ("upstream `filesystem_policy` key name" replacing the stale "renames `filesystem_policy` → `filesystem`" description). ✅
- **No `panic()`, no `any` types, no token logging** — this is a rename-only change in non-auth paths. ✅
- **Migration comment** — correctly updated to "Superseded by restoreFilesystemPolicyKey." so implementors understand the migration chain. ✅

---

### Minor observation (non-blocking)

The `restoreFilesystemPolicyKey` migration atomically renames any stored `filesystem` key back to `filesystem_policy`. Worth confirming the migration runs before the control plane starts serving traffic on a fresh deploy — if a deployment runs old API server code (with `filesystem`) and then upgrades, there's a brief window where stored specs use the old key. This is the standard ACP migration-before-startup pattern; just confirming it's in scope for the deploy runbook.

---

**Confidence:** High — complete rename, clean migration, all consumers updated.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #389 Round 2 (commit `aacf1a7d`)

One new commit since my prior approval at `c6a008453b7a`. **Approval stands.**

---

### What changed

**`components/ambient-api-server/Dockerfile` — remove `util-linux` from runtime image**

Correct fix. The Go binary is statically linked via `CGO_ENABLED=0` — it has no runtime dependency on `util-linux`. The package is unavailable on `ubi-minimal` for `arm64`, so its presence was breaking multi-arch CI. Removing it shrinks the image and fixes the arm64 build failure.

Security posture is unchanged or improved — fewer installed packages means smaller attack surface. ✅

---

No new issues. Clean removal.

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #389 Round 3 (commit `73032aa`)

One new commit since my Round 2 approval at `aacf1a7d`. **Approval stands.**

---

### What changed

**`.github/workflows/test-local-dev.yml` — target StatefulSet pod directly in `kubectl wait`**

The label selector `-l app.kubernetes.io/instance=openshell-gateway` was matching both the `openshell-gateway-0` StatefulSet pod and the `openshell-gateway-certgen` Job pod. Job pods never reach `condition=ready`, so the wait always timed out after 120s and cascaded into E2E failures.

Switching to `pod/openshell-gateway-0` by name is the correct fix:
- StatefulSet pods are named deterministically (`<name>-0`, `<name>-1`, etc.) ✅
- The certgen Job pod is excluded from the wait entirely ✅
- The updated comment accurately describes the reason for the change ✅

Two-line change, high confidence, no security or convention concerns.

---

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #389 Round 4 (commit `36170e13`)

One new commit since Round 3. **Approval stands.**

### What changed

**`components/ambient-control-plane/internal/reconciler/gateway_reconciler.go` — gofmt alignment fix**

Pure whitespace: removes the extra padding that was aligning map literal values in three `ownerReferences` blocks (`buildGRPCRouteObject`, `buildBackendTLSPolicy`, `buildCAConfigMap`). The old style used tab-alignment to line up the `=` column; `gofmt` normalizes to single-tab separation. No logic changes whatsoever — the `controller: true` field and all key/value pairs are identical.

Caught by pre-push hook, committed separately — correct practice per CLAUDE.md conventions. ✅

---

*— Amber*

jsell-rh (approved):
## 🤖 Amber — PR #389 Round 5 (commit `448b3fda`)

One new commit since my Round 4 approval at `36170e13`. **Approval stands.**

---

### `448b3fda` — add `/dev/pts` and `/dev/shm`, improve E2E diagnostics

**Policy files (`/dev/pts`, `/dev/shm` added to `read_write`):**

Both additions are correct under Landlock enforcement:
- `/dev/pts` — PTY device allocation path. Node.js subprocess `openpty()` calls resolve through `/dev/pts`. Required for Claude Code's subprocess execution under Landlock when the PTY master is already open. ✅
- `/dev/shm` — POSIX shared memory. V8 uses this for ArrayBuffer backing and IPC with worker threads. Required for Node.js V8 engine operation under Landlock. ✅

Updated consistently across all 5 policy locations: `runners/ambient-runner/policy.yaml`, `examples/base/policies/locked-down.yaml`, `mock-llm-permissive.yaml`, `permissive.yaml`, and `specs/platform/e2e-test-tooling.spec.md`. ✅

**Diagnostic blocks in `gateway-e2e-test.sh`:**

Four failure paths now emit diagnostic bundles (session phase, pod status, pod events, runner log, supervisor log, control plane log). All diagnostic commands use `|| true` so they cannot mask the original failure exit. ✅

`kubectl exec ... cat /sandbox/.runner/logs/runner.log ... || echo "(no runner log)"` — correctly handles the case where the pod is in a bad state and exec fails. ✅

`jq` calls use `2>/dev/null || true` — safe in environments where jq isn't available or the response isn't JSON. ✅

One note: the diagnostic blocks emit up to 80–100 lines per failure, which could be noisy in a run with multiple failures. But that's intentional for CI debuggability — the tradeoff is clearly worth it.

---

No convention violations. Targeted bugfix + observability improvement.

*— Amber*

PR #391, opened 2026-07-17
feat(spec): add gateway database provisioning specification

bsquizz:
## Summary

- Adds `specs/platform/gateway-db-provisioning.spec.md` defining how the GatewayReconciler optionally provisions a PostgreSQL database alongside OpenShell gateway deployments
- When `database.type: postgres`, the gateway switches from StatefulSet to Deployment and reads its connection string from a provisioned Secret via `OPENSHELL_DB_URL`
- Covers resource provisioning (Secret, PVC, Deployment, Service), credential security, workload switching, type transitions, and deletion lifecycle

## Test plan

- [ ] Review spec for completeness against existing `gateway-provisioning.spec.md`
- [ ] Validate schema alignment with `data-model.spec.md` Gateway kind
- [ ] Confirm backward compatibility: no `database` field = existing SQLite behavior unchanged

### DIFF
--- a/specs/platform/gateway-db-provisioning.spec.md
+++ b/specs/platform/gateway-db-provisioning.spec.md
@@ -1,0 +1,413 @@
+# Gateway Database Provisioning Specification
+
+**Date:** 2026-07-17
+**Status:** Design
+**Extends:** `gateway-provisioning.spec.md` — gateway lifecycle and reconciliation
+**Related:** `data-model.spec.md` — Gateway kind definition; `control-plane.spec.md` — CP reconciliation patterns
+**Upstream:** [OpenShell Helm Chart `server.externalDbSecret`](https://github.com/NVIDIA/OpenShell/tree/main/deploy/helm/openshell) — external PostgreSQL integration
+
+---
+
+## Purpose
+
+The control plane SHALL optionally provision a PostgreSQL database alongside each OpenShell gateway deployment. Today, gateways use an embedded SQLite database stored on a PVC attached to a StatefulSet. This limits gateways to a single replica and ties data durability to pod-local storage.
+
+When a Gateway resource declares `database.type: postgres`, the GatewayReconciler SHALL provision a dedicated PostgreSQL Deployment, PVC, Service, and credentials Secret in the same namespace. The gateway workload switches from StatefulSet to Deployment and reads its database connection string from the provisioned Secret via the `OPENSHELL_DB_URL` environment variable, following the upstream OpenShell `server.externalDbSecret` pattern.
+
+This enables:
+- **PostgreSQL-backed gateways** — production-grade storage with standard PostgreSQL tooling
+- **Deployment workload** — no StatefulSet PVC required for the gateway itself; database has its own PVC
+- **Extensible database backends** — the `database.type` field supports `sqlite` (default), `postgres` (this spec), and future backends (e.g., `rds`)
+- **Lifecycle coupling** — the database is created and destroyed with the gateway; no orphaned resources
+
+---
+
+## Requirements
+
+### Requirement: Gateway Database Configuration Field
+
+The Gateway resource SHALL support an optional `database` field that controls the database backend. When absent or null, the gateway uses the default SQLite behavior (backward compatible).
+
+#### Scenario: Gateway with PostgreSQL database
+
+- GIVEN a Gateway resource:
+  ```yaml
+  kind: Gateway
+  name: openshell-gateway
+  project: tenant-a
+  database:
+    type: postgres
+    storageSize: 10Gi
+  image: ghcr.io/nvidia/openshell:v0.0.70
+  serverDnsNames:
+    - openshell-gateway.tenant-a.svc.cluster.local
+  ```
+- WHEN the GatewayReconciler processes the Gateway
+- THEN it SHALL provision a PostgreSQL Deployment, PVC, Service, and Secret in the `tenant-a` namespace
+- AND it SHALL deploy the gateway as a Deployment (not StatefulSet)
+- AND the gateway container SHALL receive `OPENSHELL_DB_URL` from the provisioned Secret
+
+#### Scenario: Gateway with default SQLite database
+
+- GIVEN a Gateway resource with no `database` field (or `database.type: sqlite`)
+- WHEN the GatewayReconciler processes the Gateway
+- THEN it SHALL deploy the gateway as a StatefulSet with embedded SQLite (existing behavior)
+- AND no PostgreSQL resources SHALL be provisioned
+
+#### Scenario: Gateway with unsupported database type
+
+- GIVEN a Gateway resource with `database.type: rds`
+- WHEN the GatewayReconciler validates the configuration
+- THEN it SHALL log a validation warning: unsupported database type
+- AND it SHALL NOT provision any database resources
+- AND it SHALL NOT deploy the gateway (configuration is invalid until corrected)
+
+---
+
+### Requirement: Database Field Schema
+
+The `database` field SHALL be a JSONB object on the Gateway resource with the following sub-fields:
+
+| Field | Type | Required | Default | Description |
+|---|---|---|---|---|
+| `type` | string | Yes (when `database` is set) | `sqlite` | Database backend: `sqlite`, `postgres`. Reserved for future: `rds`. |
+| `storageSize` | string | No | `5Gi` | PVC storage request for the PostgreSQL data volume. Only applicable when `type: postgres`. Ignored for `type: sqlite`. |
+
+#### Scenario: Database field defaults
+
+- GIVEN a Gateway resource with `database: { type: postgres }` and no `storageSize`
+- WHEN the GatewayReconciler provisions the database PVC
+- THEN the PVC SHALL request `5Gi` of storage
+
+#### Scenario: Database field absent
+
+- GIVEN a Gateway resource with no `database` field
+- WHEN the GatewayReconciler processes the Gateway
+- THEN it SHALL treat the gateway as `database.type: sqlite`
+- AND behavior SHALL be identical to the pre-existing gateway provisioning
+
+---
+
+### Requirement: PostgreSQL Resource Provisioning
+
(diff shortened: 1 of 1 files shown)


specs/platform/gateway-db-provisioning.spec.md
```diff
@@ -0,0 +1,413 @@
...
+```
+
+A database migration SHALL add the `database` column to the `gateways` table:
+
+```sql
+ALTER TABLE gateways ADD COLUMN database JSONB;
+```
+
+### New Components
+
```
jsell-rh, line 388: **🔴 Blocker — RBAC claim is incorrect**

> "No additional RBAC grants are required. The control plane ServiceAccount already has permissions to create Deployments, Services, Secrets, and PersistentVolumeClaims in project namespaces"

`persistentvolumeclaims` is **not** in [`control-plane-clusterrole.yaml`](https://github.com/openshift-online/agent-control-plane/blob/main/components/manifests/base/rbac/control-plane-clusterrole.yaml) — I checked. The reconciler will 403 at PVC creation.

The implementation PR must add:
```yaml
- apiGroups: [""]
  resources: ["persistentvolumeclaims"]
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```

Update this section to say "The following RBAC grants must be added to `control-plane-clusterrole.yaml`."

jsell-rh, line 388: **🔴 Blocker — Missing `delete` verb for Deployments/StatefulSets**

The existing ClusterRole has `["create", "get", "update", "patch"]` for `apps/deployments,statefulsets` — no `delete`. This spec requires deletion in three scenarios:
1. `sqlite→postgres` transition: delete existing StatefulSet
2. `postgres→sqlite` transition: delete existing Deployment  
3. Gateway deletion with postgres: delete `openshell-gateway-db` Deployment

The implementation PR must add `delete` to that rule. The spec's RBAC section should document this explicitly.

specs/platform/gateway-db-provisioning.spec.md
```diff
@@ -0,0 +1,413 @@
...
+---
+
+### Requirement: Database Credential Security
+
+The database credentials Secret SHALL be generated securely and the password SHALL never be exposed in logs, error messages, or API responses.
+
+#### Scenario: Password generation
+
+- GIVEN a new Gateway with `database.type: postgres`
```
jsell-rh, line 156: **🟡 Major — Missing `readOnlyRootFilesystem: true` and `seccompProfile`**

CLAUDE.md requires `readOnlyRootFilesystem` on all containers. The spec lists:
> `runAsNonRoot: true`, `allowPrivilegeEscalation: false`, capabilities `drop: [ALL]`

Two things missing:

1. `readOnlyRootFilesystem: true` — PostgreSQL needs two writable dirs outside `PGDATA`: `/var/run/postgresql` (Unix socket) and `/tmp`. The implementation will need `emptyDir` volumeMounts for both:
```yaml
volumeMounts:
  - name: pg-run
    mountPath: /var/run/postgresql
  - name: tmp
    mountPath: /tmp
volumes:
  - name: pg-run
    emptyDir: {}
  - name: tmp
    emptyDir: {}
```

2. `seccompProfile: type: RuntimeDefault` — consistent with what was flagged in PR #384 for other ACP containers.

Update the SecurityContext spec to include both.

jsell-rh (changes requested):
## 🤖 Amber Review — PR #391: Gateway Database Provisioning Spec

This is a well-structured spec and the design intent is solid — optional PostgreSQL backend, backward-compatible SQLite default, clean lifecycle coupling, and correct SecurityContext for the DB container. The resource provisioning order, idempotency requirements, and password preservation on reconcile are all well thought through.

Three issues need resolving before implementation starts. Two are RBAC gaps that will cause runtime failures; one is an ACP convention gap.

---

### [Blocker] RBAC claim is incorrect — PVC permissions missing from `control-plane-clusterrole.yaml`

The spec states in the RBAC Requirements section:

> "No additional RBAC grants are required. The control plane ServiceAccount already has permissions to create Deployments, Services, Secrets, and PersistentVolumeClaims in project namespaces"

I verified this against the actual [`control-plane-clusterrole.yaml`](https://github.com/openshift-online/agent-control-plane/blob/main/components/manifests/base/rbac/control-plane-clusterrole.yaml). **`persistentvolumeclaims` is not listed anywhere in the ClusterRole.** The implementation will fail at runtime with a 403 when the GatewayReconciler tries to create `openshell-gateway-db-data`. This needs to be called out in the spec and the RBAC manifest updated in the implementation PR.

Fix for `control-plane-clusterrole.yaml`:
```yaml
# Gateway database provisioning (PostgreSQL PVC)
- apiGroups: [""]
  resources: ["persistentvolumeclaims"]
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```

---

### [Blocker] RBAC gap — no `delete` verb for Deployments/StatefulSets

The existing ClusterRole has:
```yaml
- apiGroups: ["apps"]
  resources: ["deployments", "statefulsets"]
  verbs: ["create", "get", "update", "patch"]  # ← no delete
```

But the spec requires deletion in three places:
1. **sqlite→postgres transition**: delete the existing StatefulSet `openshell-gateway`
2. **postgres→sqlite transition**: delete the existing Deployment `openshell-gateway`
3. **Gateway deletion with postgres**: delete the database Deployment `openshell-gateway-db`

The implementation PR must add `delete` to the apps resources verbs. The spec's RBAC section should call this out explicitly.

---

### [Major] Missing `readOnlyRootFilesystem: true` in DB container SecurityContext

The spec specifies for the PostgreSQL container:
> `SecurityContext: runAsNonRoot: true, allowPrivilegeEscalation: false, capabilities drop: [ALL]`

CLAUDE.md requires `readOnlyRootFilesystem` on all containers. The spec is missing it, and the implementation PR will need emptyDir mounts for PostgreSQL's writable paths (`/var/run/postgresql`, `/tmp`). The `PGDATA` path is covered by the PVC mount, but PostgreSQL needs at least two more writable dirs at startup.

Also missing: `seccompProfile: type: RuntimeDefault` (consistent with what was flagged in PR #384 for the other containers).

---

### [Minor] `sslmode=disable` should be documented as a conscious tradeoff

The URI `postgresql://openshell:<password>@openshell-gateway-db:5432/openshell?sslmode=disable` disables TLS on the DB connection. This is in-cluster, same-namespace traffic — the risk is low and the decision is pragmatic. But the spec should include one sentence acknowledging this is an intentional choice so implementors and reviewers aren't left wondering.

---

### What's solid ✅

- Backward compat: no `database` field = SQLite, identical existing behavior
- `crypto/rand` for password generation
- Password preservation on reconcile (read existing secret, don't overwrite)
- `Recreate` strategy for single-replica PostgreSQL — rolling update on a single DB instance is unsafe
- Resource provisioning order (DB resources before gateway workload)
- Deletion lifecycle coupled to gateway deletion — no orphaned resources
- Both transition directions (sqlite↔postgres) covered including data loss warnings
- Labels following `app.kubernetes.io/` conventions

**Confidence:** High — verified RBAC gap against `control-plane-clusterrole.yaml` directly.

*— Amber*

markturansky:
## Review: Gateway Database Provisioning Spec

Solid spec — the "model it in the API, reconcile it" approach is the right pattern and this captures ~90% of what's needed for Phase 1. A few observations:

### 1. Reserve schema space for external databases now

The spec has `type: postgres` and `type: rds` (reserved), but no field for bringing your own database endpoint. When RDS/CloudSQL arrives (or a user wants to point at an existing Postgres instance), we'll need something like `database.externalSecretRef` or `database.url`.

Adding this field now — even if unsupported/unimplemented — avoids a schema break later. Suggested addition:

```
 < /dev/null |  externalSecretRef | string | No | — | Name of a K8s Secret containing a `url` key with the connection string. When set, the reconciler skips DB provisioning and mounts this Secret instead. Only valid with `type: postgres`. |
```

This mirrors how OpenShell's upstream Helm chart handles `server.externalDbSecret`.

### 2. Init container for DB readiness

The spec says the gateway will fail its readiness probe until the DB is available, relying on Kubernetes restart behavior. This works, but the gateway will log errors and restart-loop during initial provisioning.

ACP's own API server uses an init container that runs `pg_isready` before the main process starts. The gateway Deployment template should do the same — the pod doesn't even start until DB is reachable. Cleaner logs, no wasted restarts.

### 3. NetworkPolicy for the DB pod

The gateway already gets a `networkpolicy.yaml`. The DB pod should also get one — restrict ingress to port 5432 from pods matching `app.kubernetes.io/instance: openshell-gateway` only. Without this, any pod in the namespace can reach the database.

### 4. RHEL image handling

ACP's existing Postgres pattern swaps `postgres:16` for `registry.redhat.io/rhel10/postgresql-16:10.1` in production/OpenShift overlays via kustomize components. The spec hardcodes `postgres:16` in the reconciler manifests.

The DB manifest template should support an image override — either via an env var (`OPENSHELL_GATEWAY_DB_IMAGE`) or a `database.image` field — so production deployments can use the RHEL-certified image without forking the template. The RHEL postgres image also uses different env var names (`POSTGRESQL_USER` vs `POSTGRES_USER`), which the reconciler would need to handle.

### 5. Secret key naming (minor)

The spec uses `uri` as the Secret key for the connection string, but the gateway reads `OPENSHELL_DB_URL`. Naming the key `url` would be more consistent with the env var. Minor, but reduces cognitive overhead when debugging.

### 6. Production limitations note

Single-replica Postgres Deployment with `Recreate` strategy is fine for dev/demo. Worth adding a note in the spec that this is not HA — if the node the PVC is bound to goes down, the database is unavailable until rescheduled. This sets honest expectations and motivates the future `externalSecretRef` / managed-DB path.

---

Overall: strong foundation. The `externalSecretRef` addition is the most important piece — it's the escape hatch that connects Phase 1 (in-cluster pod) to Phase 2 (RDS/managed) without a breaking API change.

markturansky:
implemented in https://github.com/openshift-online/agent-control-plane/pull/405

PR #390, opened 2026-07-17
feat(test): add OpenShell CLI e2e test and CI integration

squizzi:
## Summary

Adds a comprehensive OpenShell CLI e2e test suite (`tests/e2e/openshell-cli-e2e.sh`) covering 34 scenarios across 8 sections — sandbox lifecycle, settings, policy management, gateway operations, network policies, and cleanup. Includes a tmux-based interactive demo script (`components/ambient-cli/demo-openshell.sh`) for live walkthroughs. The e2e test is integrated into the `local-dev-simulation` CI job in `.github/workflows/test-local-dev.yml`, running after the existing gateway e2e tests. Also updates the policy fixture to use `filesystem_policy:` with required default paths, and registers the spec in the index.

## Test plan

- [x] All 34 e2e test scenarios pass locally
- [ ] CI `local-dev-simulation` job passes with the new OpenShell CLI E2E step
- [ ] Demo script runs end-to-end in a tmux session

PR #393, opened 2026-07-18
chore(tests): standardize e2e test output to human-readable style

squizzi:
## Summary

- Codifies the human-readable e2e test style from PR #390 (`openshell-cli-e2e.sh`) into `specs/platform/e2e-test-tooling.spec.md` with 4 requirements: unified output helpers, section banners, centralized cleanup, and script headers
- Migrates all 9 remaining e2e test scripts to the new style: `[PASS]`/`[FAIL]`/`[SKIP]` bracket tags replaced with `✓`/`✗`/`⊘` emoji markers, `# ====` section banners added, counter variables standardized to `PASSED`/`FAILED`, and results summaries unified
- Largest rewrites in `rbac_e2e_test.sh`, `scheduled_sessions_e2e_test.sh`, `local-dev-test.sh`, and `route-e2e-test.sh`; lighter touch on scripts that already used emoji markers (adding banners only)
- `gateway-e2e-test.sh` also gets centralized cleanup with session tracking array and `SKIP_CLEANUP` support

## Test plan

- [x] All 9 modified scripts pass `bash -n` syntax checking
- [x] No old `[PASS]`/`[FAIL]`/`[SKIP]` bracket tags remain across `tests/e2e/`
- [x] All section calls have corresponding `# ============================================================================` banners
- [ ] Run `make kind-up` and execute individual e2e tests to verify runtime behavior


tests/e2e/gateway-e2e-test.sh
```diff
@@ -587,6 +656,10 @@ else
   fail "Sandbox configuration verification — session not created"
 fi
 
+# ============================================================================
+# Section 11: Long-running session — LLM response and sandbox persistence
```
squizzi, line 660: **Duplicate section number.** This banner says "Section 11" but line 528 already has "Section 11: Sandbox configuration verification". This should be Section 12, and all subsequent sections (currently 12-14) should be renumbered to 13-15. The `section()` call at line 663 has the same duplicate "11." in the display string.

tests/e2e/gateway-e2e-test.sh
```diff
@@ -57,7 +47,7 @@ elif [ -n "${1:-}" ]; then
 else
   API_URL="http://localhost:${PF_PORT}"
 fi
-trap 'kill "${PF_PID}" 2>/dev/null || true; kill "${GW_PF_PID}" 2>/dev/null || true' EXIT
+trap cleanup EXIT INT TERM
```
squizzi, line 50: **trap registered before `cleanup()` is defined.** The `cleanup()` function isn't defined until line 173. If the script exits between here and there (e.g., during `_ensure_port_forward` at line 72 via `set -e`), bash fires the trap and fails with "cleanup: command not found". Consider moving this trap below the `cleanup()` definition or moving `cleanup()` above this line.

tests/e2e/gateway-e2e-test.sh
```diff
@@ -179,6 +170,39 @@ _cleanup_sandboxes() {
...
 }
 
+cleanup() {
+  # Idempotent — safe to call from trap and explicitly
+  if [ "${_CLEANUP_DONE:-false}" = "true" ]; then return 0; fi
+  _CLEANUP_DONE=true
+
+  if [ "$SKIP_CLEANUP" = "true" ]; then
+    echo -e "  ${YELLOW}Skipping cleanup (--skip-cleanup)${NC}"
+    for _sid in "${CREATED_SESSIONS[@]}"; do
```
squizzi, line 180: **bash 3.2 compatibility.** With `set -u`, `"${CREATED_SESSIONS[@]}"` on an empty array causes "unbound variable" on bash 3.2 (macOS system default). The C-style loop at line 193 handles this correctly, but this `for-in` loop in the skip-cleanup branch does not. Guard with `[ ${#CREATED_SESSIONS[@]} -gt 0 ]` or use the same C-style pattern.

tests/e2e/local-dev-test.sh
```diff
@@ -775,52 +797,48 @@ main() {
...
             if [ $KNOWN_FAILURES -gt 0 ]; then
                 log_warning "Note: $KNOWN_FAILURES known TODOs tracked in test output"
             fi
             exit 0
         else
-            echo -e "${RED}${BOLD}✗ $unexpected_failures unexpected test failures${NC}"
+            echo -e "${RED}${BOLD}$unexpected_failures unexpected test failures${NC}"
             echo ""
-            log_error "CI validation failed"
+            fail "CI validation failed"
```
squizzi, line 821: **`fail()` called after results summary.** The results line at line 801 has already printed the pass/fail counts. This `fail "CI validation failed"` increments `FAILED` after the summary was displayed, so the printed count is off by one. It also prints a test-result line after the summary, breaking the convention that the results line is the final output. Move this `fail()` call before the summary block, or remove it entirely since `exit 1` already signals failure.

tests/e2e/local-dev-test.sh
```diff
@@ -76,40 +76,29 @@ while [[ $# -gt 0 ]]; do
...
 log_warning() {
-    echo -e "${YELLOW}⚠${NC} $*"
-}
-
-log_section() {
-    echo ""
-    echo -e "${BOLD}═══════════════════════════════════════════${NC}"
-    echo -e "${BOLD}  $*${NC}"
-    echo -e "${BOLD}═══════════════════════════════════════════${NC}"
+    echo -e "  ${YELLOW}⊘${NC} $*"
```
squizzi, line 91: **`log_warning` uses `⊘` (skip marker).** This makes warning output visually indistinguishable from `skip()` output. A reader scanning terminal output can't tell whether a `⊘` line is a skipped test or an informational warning. Consider keeping the original `⚠` symbol for warnings, or using a distinct marker to preserve scan-readability.

tests/e2e/rbac_e2e_test.sh
```diff
@@ -19,19 +19,19 @@ kubectl() {
...
 RED='\033[0;31m'
 YELLOW='\033[0;33m'
 BOLD='\033[1m'
 NC='\033[0m'
 
-pass() { PASS_COUNT=$((PASS_COUNT + 1)); echo -e "  ${GREEN}[PASS]${NC} $1"; }
-fail() { FAIL_COUNT=$((FAIL_COUNT + 1)); echo -e "  ${RED}[FAIL]${NC} $1: $2"; }
-skip() { SKIP_COUNT=$((SKIP_COUNT + 1)); echo -e "  ${YELLOW}[SKIP]${NC} $1"; }
+pass() { PASSED=$((PASSED + 1)); echo -e "  ${GREEN}✓${NC} $1"; }
+fail() { FAILED=$((FAILED + 1)); echo -e "  ${RED}✗${NC} $1: $2"; }
```
squizzi, line 32: **`fail()` signature diverges from spec.** The spec defines a single-argument `fail()` but this uses `$1: $2`. The two-arg form is reasonable since callers pass description + details, but it doesn't match the canonical pattern. Same issue exists in `scheduled_sessions_e2e_test.sh:42`.

Also: `skip()` at line 33 hardcodes `(skipped)` without the `$2` reason argument that the spec defines as `(skipped: $2)`. Callers embed the reason in `$1` instead, so the output format won't match other scripts that use the canonical `(skipped: reason)` form.

tests/e2e/vteam-catalog-lab-test.sh
```diff
@@ -25,7 +25,7 @@ section() { echo ""; echo -e "${BOLD}$1${NC}"; }
 
 finish() {
   echo ""
-  echo -e "${BOLD}Results:${NC} ${GREEN}${PASSED} passed${NC}, ${YELLOW}${SKIPPED} skipped${NC}, ${RED}${FAILED} failed${NC}"
+  echo -e "${BOLD}Results: ${GREEN}${PASSED} passed${NC}, ${RED}${FAILED} failed${NC}"
```
squizzi, line 28: **nit: Dead `SKIPPED` variable.** The results summary was updated to drop the skipped count, but `SKIPPED=0` is still declared (line 19) and `skip()` still increments it (line 23). Either remove `SKIPPED` entirely or keep it in this results line.

squizzi (commented):
🤖 **Amber Review**

Reviewed all 10 changed files. The style migration is solid — no test logic was accidentally modified, all bracket tags are gone, and section banners are correctly formatted across the board.

**2 errors, 4 warnings, 2 nits found.**

**Errors:**
- `gateway-e2e-test.sh:660` — Duplicate "Section 11" numbering (should be 12, with 12-14 renumbered to 13-15)
- `local-dev-test.sh:821` — `fail()` called after results summary, so printed count is off by one

**Warnings:**
- `gateway-e2e-test.sh:50` — `trap cleanup` registered before `cleanup()` is defined (early exit = "command not found")
- `gateway-e2e-test.sh:180` — bash 3.2 array expansion on empty `CREATED_SESSIONS` array with `set -u`
- `rbac_e2e_test.sh:32` / `scheduled_sessions_e2e_test.sh:42` — `fail()` and `skip()` signatures diverge from spec
- `local-dev-test.sh:91` — `log_warning` uses `⊘` skip marker, indistinguishable from `skip()` output

**Nits:**
- `vteam-catalog-lab-test.sh:19` — Dead `SKIPPED` variable (incremented but never displayed)
- `rbac_e2e_test.sh:31` — `pass()` increment ordering differs from spec (functionally identical)

**Clean files (no issues):** `specs/platform/e2e-test-tooling.spec.md`, `route-e2e-test.sh`, `hcmai-smoke.sh`, `smoke-test-llm.sh`, `openshell-dual-tenant.sh`

Confidence: **90%** — thorough diff review, all findings verified against source. Did not run the tests at runtime.

PR #395, opened 2026-07-18
style(tests): update gateway e2e test to enhanced output style

squizzi:
## Summary

Aligns `gateway-e2e-test.sh` with the enhanced output style established in `openshell-cli-e2e.sh`. Adds `CYAN`, `ORANGE`, and `DIM` color codes, introduces a `sep()` separator function, upgrades `section()` to use cyan-colored headers with dim separator lines, and updates all six results summaries (five early-exit plus the final) to use the styled format with conditional pass/fail coloring.

## Test plan

- [ ] Run `bash -n tests/e2e/gateway-e2e-test.sh` to verify syntax
- [ ] Run the gateway e2e test on a Kind cluster and verify the new styled output renders correctly
- [ ] Confirm section headers display with cyan color and separator lines
- [ ] Confirm results summary uses separator borders and conditional coloring

### DIFF
--- a/tests/e2e/gateway-e2e-test.sh
+++ b/tests/e2e/gateway-e2e-test.sh
@@ -131,7 +131,6 @@ RED='\033[0;31m'
 GREEN='\033[0;32m'
 YELLOW='\033[1;33m'
 CYAN='\033[36m'
-ORANGE='\033[38;5;214m'
 DIM='\033[2m'
 BOLD='\033[1m'
 NC='\033[0m'
@@ -151,6 +150,18 @@ section() {
   sep
   printf '%b━━  %s%b\n' "${CYAN}" "$*" "${NC}"
   sep
+}
+
+results() {
+  echo ""
+  sep
+  if [ "$FAILED" -gt 0 ]; then
+    printf '%b  %d passed, %d failed%b\n' "${RED}" "$PASSED" "$FAILED" "${NC}"
+  else
+    printf '%b  %d passed ✓%b\n' "${GREEN}" "$PASSED" "${NC}"
+  fi
+  sep
+  echo ""
 }
 
 should_run_test() {
@@ -251,7 +262,7 @@ if [ -n "$ACPCTL" ]; then
   pass "acpctl found: $ACPCTL"
 else
   fail "acpctl not found — run 'make build-cli'"
-  echo ""; sep; printf '%b  %d passed, %d failed%b\n' "${RED}" "$PASSED" "$FAILED" "${NC}"; sep; echo ""
+  results
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


tests/e2e/gateway-e2e-test.sh
```diff
@@ -130,6 +130,9 @@ _ensure_gateway_port_forward() {
 RED='\033[0;31m'
 GREEN='\033[0;32m'
 YELLOW='\033[1;33m'
+CYAN='\033[36m'
+ORANGE='\033[38;5;214m'
```
squizzi: `ORANGE` is defined but never referenced anywhere in this script. In the reference implementation (`openshell-cli-e2e.sh`), `ORANGE` is consumed by `run_cmd()` to highlight the command being executed. Since `run_cmd()` wasn't ported here, this is a dead variable.

Either remove the unused `ORANGE` declaration, or add the `run_cmd()` helper to match the reference style fully. I'd lean toward removing it — the gateway test doesn't wrap commands in `run_cmd()` and retrofitting that would be a larger refactor.

tests/e2e/gateway-e2e-test.sh
```diff
@@ -241,7 +251,7 @@ if [ -n "$ACPCTL" ]; then
   pass "acpctl found: $ACPCTL"
 else
   fail "acpctl not found — run 'make build-cli'"
-  echo -e "\n${BOLD}Results: ${GREEN}${PASSED} passed${NC}, ${RED}${FAILED} failed${NC}\n"
+  echo ""; sep; printf '%b  %d passed, %d failed%b\n' "${RED}" "$PASSED" "$FAILED" "${NC}"; sep; echo ""
```
squizzi: Nit: The early-exit summary is a single long line that's harder to scan than the multi-line version at the end of the script. This is fine functionally, but worth noting the inconsistency — the final results block (line 1173-1180) uses a readable multi-line `if/else` with a conditional `✓` on success, while all 5 early-exit summaries use a one-liner that always renders the failure format. Since early exits only fire on failure this is technically correct, but it diverges from the final block's structure.

squizzi (commented):
🤖 **Amber Review**

Good alignment with the `openshell-cli-e2e.sh` reference style. The `sep()`, enhanced `section()`, and styled results summary are correct and match the reference implementation. Syntax validates clean (`bash -n`).

**One finding:**
- `ORANGE` color is declared but unused — it's only consumed by `run_cmd()` in the reference, which wasn't ported. Recommend removing the dead declaration.

**Spec compliance note:** The `e2e-test-tooling.spec.md` (Requirement: Unified Output Helpers) codifies the *simpler* style (`section()` as a one-liner with bold) as the minimum. This PR goes *beyond* the spec by adopting the enhanced reference style, which is the right call for consistency across the test suite — the spec is a floor, not a ceiling.

**Confidence:** 95% — Pure cosmetic change, zero behavioral impact. The only risk is if a CI log parser regex'd on the old `Results:` format, which is unlikely.

Overall: Clean, small, correct. Remove the unused `ORANGE` and this is ready to merge.

squizzi (commented):
🤖 **Amber Review — Pass (re-review after fixes)**

Both prior findings addressed:
- **Unused `ORANGE` removed** — only `CYAN` and `DIM` remain, both consumed by `sep()` and `section()`.
- **`results()` helper extracted** — all 6 summary callsites (5 early-exit + 1 final) now use the same function with consistent conditional pass/fail formatting. Clean DRY improvement.

No new findings. All defined color variables are referenced. `bash -n` syntax check passes. Diff is minimal and purely cosmetic — zero behavioral change.

**Confidence:** 95% — Cosmetic-only change to a test script. No risk to production code.

PR #414, opened 2026-07-20
fix(ci): add missing kubectl/kind version env vars to local-dev-simulation

squizzi:
## Summary

The `test-local-dev-simulation` job in `test-local-dev.yml` was failing because `KUBECTL_VERSION` and `KIND_VERSION` environment variables were never defined. The curl download URLs expanded to broken paths (e.g. `/release//bin/linux/amd64/kubectl`), causing GCS to return an XML error page that was saved as the kubectl binary.

This adds the missing `env:` block with `KUBECTL_VERSION: v1.31.4` and `KIND_VERSION: v0.31.0`, matching the values used in `vteam-catalog-lab-e2e.yml`.

## Test plan

- CI should pass the `test-local-dev-simulation` job without the kubectl syntax error
- Verify `kubectl version --client` and `kind version` succeed in the workflow logs

### DIFF
--- a/.github/workflows/test-local-dev.yml
+++ b/.github/workflows/test-local-dev.yml
@@ -61,6 +61,10 @@ jobs:
       KUBECTL_VERSION: v1.31.4
       KIND_VERSION: v0.31.0
 
+    env:
+      KUBECTL_VERSION: v1.31.4
+      KIND_VERSION: v0.31.0
+
     steps:
     - name: Checkout code
       uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0


jsell-rh (approved):
🤖 **Amber Review**

The CI fix is correct and targeted. The three companion improvements in `gateway-e2e-test.sh` are all clearly better than what they replace.

**What I verified:**
- `KUBECTL_VERSION: v1.31.4` / `KIND_VERSION: v0.31.0` match `vteam-catalog-lab-e2e.yml` — consistent.
- Pod-ready wait extended from 60 s (30 × 2 s) to 180 s (90 × 2 s) with a comment explaining why (`Kind CI is slow`) — correct call.
- Short-session completion check now gated on `SHORT_LLM_FOUND > 0` — avoids a spurious `fail` when LLM never responds.
- `_delete_session` instead of raw `api DELETE + sleep 2` — uses the canonical helper, consistent with the rest of the script.
- `OPENSHELL_GATEWAY_SA_TOKEN_PATH=""` env var cleared before gateway port-forward is established — sequence makes the comment trustworthy.

**Minor (no action needed):** This is marked Draft with 3 commits — squash before marking ready if not already planned.

No blockers. 

— Amber

squizzi:
Closing this as the same fix is already in main and was added in #423

PR #420, opened 2026-07-22
feat(tooling): add demo-creator skill for deterministi demos

jeremyeder:
## Summary

Adds the **`demo-creator`** skill under `skills/tooling/demo-creator/` — a deterministic, reproducible product-demo pipeline (capture → compose → render) built in Node.js (ESM `.mjs`).

The skill drives Playwright-based capture, composes captured steps into a scripted timeline, and renders narrated demo artifacts (video frames, `captions.vtt`, OCR text). Output is deterministic so demos can be regenerated and diffed in CI.

### Runtime secret redaction

A core responsibility of the pipeline is scrubbing real credentials out of captured demo artifacts (screen output, captions, OCR text, environment dumps) before they are written. This lives in:

- `scripts/core/security.mjs`
- `scripts/compose/secrets.mjs`
- `scripts/compose/security-values.mjs`

This operates on **runtime artifacts that never enter git** — it is distinct from repo-level secret scanning.

## Verification

- `node --test` from `skills/tooling/demo-creator/`: **744 pass / 0 fail / 14 skip** (758 total)
- Node v22.22.3; deps: ajv 8.20.0, playwright 1.61.1, yaml 2.9.0

## Review

- `/review` (ACP review guidance): no Critical/Major findings
- CodeRabbit CLI review: Critical/Major findings remediated

## Notes for reviewers

The secret-shaped strings under `skills/tooling/demo-creator/tests/` are **synthetic test fixtures** — they are the token-shaped *inputs* that prove the runtime redaction works, plus Android UI-hierarchy `password="true|false"` boolean attributes. None are real credentials.

### Follow-up (out of scope for this PR)

`scripts/hooks/pr-review-gate.sh`'s secret scan (line ~99) greps added diff lines with no test-path exclusion, so it false-positives on synthetic fixtures like these. Every production secret scanner (gitleaks, trufflehog) exempts fixture/test paths by design. A one-line `**/tests/**` exclusion on that scan would fix this class of false positive without weakening production-code detection. Filing separately to avoid editing shared gate infra in this feature PR.

---

Authored and reviewed without human involvement.


jsell-rh (approved):
🤖 **Amber Review**

Large PR but the scope is well-contained: a new self-contained skill under `skills/tooling/demo-creator/` plus two meaningful production changes. Reviewed each area separately.

---

**Production changes (Go config + CI) — Clean ✅**

`envOrDefaultIfUnset` in `config.go` is the correct fix. The semantic difference matters:
- `envOrDefault` treats `OPENSHELL_GATEWAY_SA_TOKEN_PATH=""` the same as unset → always falls back to the default SA token path
- `envOrDefaultIfUnset` lets an explicit `=""` disable bearer token injection

The three test cases in `config_test.go` cover all three states (`unset → default`, `explicit empty → ""`, `explicit path → path`). Both the pattern and the test are correct.

The `setup-kind-openshell.sh` change to patch `OPENSHELL_GATEWAY_SA_TOKEN_PATH=""` onto the control plane deployment correctly exercises this new behavior. Sequence is correct: set env → check if needs patch → patch once.

---

**CI wiring — Clean ✅**
- `component-paths.json` adds `demo-creator` label mapping.
- `detect-components.sh` now uses `gh api --paginate` for PR file detection, falling back to `git diff` — better than `gh pr diff` which paginates differently. `fetch-depth: 0` added to unit-tests workflow to support the git-based fallback.
- `unit-tests.yml` demo-creator job: `ffmpeg`/`tesseract-ocr` install with `--no-install-recommends`, test matrix validated via `node --test`, 744/0 pass confirmed in PR description.

---

**Skill directory — No ACP constitution violations ✅**

This is tooling code (`.mjs`), not Go backend or TypeScript frontend, so most production code rules don't apply directly. What I verified:

- Secret redaction (`scripts/core/security.mjs`, `scripts/compose/secrets.mjs`) operates on runtime artifacts, not repo content — no risk of credentials entering git.
- Test fixtures in `tests/` use synthetic token-shaped strings assembled at runtime (`[\"private\", \"bearer\", \"token\"].join(\"-\")`) — not real credentials. The PR description explains this clearly.
- No `panic()` equivalents (no uncaught throws in production paths, errors return/reject).
- 154 files but ~100 are tests covering the new logic — good ratio.

---

**One note (no action needed):** The PR description flags a false-positive in `scripts/hooks/pr-review-gate.sh` secret scan for the test fixtures and says it's being filed separately. That's the right call — don't fix unrelated gate infra in a feature PR.

No blockers.

— Amber

PR #427, opened 2026-07-25
feat(control-plane): auto-reconcile router NetworkPolicy and NLB passthrough Route

markturansky:
## Summary

- Adds `reconcileRouterNetworkPolicy()` to automatically create `openshell-gateway-allow-router` NetworkPolicy on OpenShift when a gateway has Route config, allowing HAProxy router pods from `openshift-ingress` to reach gateway on ports 8080/8081
- Adds `reconcileOpenShiftRoute()` to automatically create `openshell-gateway-grpc` passthrough Route with `router: grpc` label for NLB-backed ingress (alternative to Gateway API GRPCRoute)
- Adds `route.openshift.io/routes` RBAC to `ambient-control-plane` ClusterRole
- Both resources are cleaned up (deleted) when Route config is removed from the gateway
- Updates RECONCILE.md: P44 and P45 marked done, coverage 92.8% → 93.4%

## Test plan

- [ ] Deploy to OpenShift cluster with NLB IngressController configured
- [ ] Create gateway with Route config → verify NetworkPolicy and Route are created
- [ ] Verify gRPC connectivity through NLB passthrough route
- [ ] Remove Route config from gateway → verify NetworkPolicy and Route are deleted
- [ ] Verify `go build ./...`, `go vet ./...`, and `gofmt` pass (verified locally)
- [ ] Verify on non-OpenShift cluster: both reconcilers are skipped (no errors)

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
@@ -320,6 +320,12 @@ func (r *GatewayReconciler) reconcileGateway(ctx context.Context, projectClient 
 		if err := r.reconcileOpenShiftSCC(ctx, namespace); err != nil {
 			r.logger.Warn().Err(err).Str("namespace", namespace).Msg("failed to reconcile OpenShift SCC binding")
 		}
+		if err := r.reconcileRouterNetworkPolicy(ctx, gw, namespace); err != nil {
+			r.logger.Warn().Err(err).Str("namespace", namespace).Msg("failed to reconcile router NetworkPolicy")
+		}
+		if err := r.reconcileOpenShiftRoute(ctx, gw, namespace); err != nil {
+			r.logger.Warn().Err(err).Str("namespace", namespace).Msg("failed to reconcile OpenShift Route")
+		}
 	}
 
 	if r.hasCertManager {
@@ -857,6 +863,16 @@ var (
 		Version:  "v1",
 		Resource: "rolebindings",
 	}
+	networkPolicyGVR = schema.GroupVersionResource{
+		Group:    "networking.k8s.io",
+		Version:  "v1",
+		Resource: "networkpolicies",
+	}
+	openShiftRouteGVR = schema.GroupVersionResource{
+		Group:    "route.openshift.io",
+		Version:  "v1",
+		Resource: "routes",
+	}
 )
 
 func (r *GatewayReconciler) reconcileOpenShiftSCC(ctx context.Context, namespace string) error {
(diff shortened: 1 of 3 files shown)


markturansky:
## Review from amber (Shell agent)

**PR:** feat(control-plane): auto-reconcile router NetworkPolicy and NLB passthrough Route

The approach is solid and fits the existing reconciler pattern well. A few observations from the OpenShell side:

**What looks good**
- Both reconcilers are correctly gated inside `r.isOpenShift` — no impact on non-OpenShift deployments.
- Cleanup on `gw.Route == nil` via `deleteIfExists` is clean and consistent with other reconcilers.
- Passthrough TLS on the Route is correct for gRPC — not terminating at the HAProxy layer is the right call.
- The `haproxy.router.openshift.io/timeout: 3600s` annotation is important for long-lived gRPC streams; good to see it included.

**Questions / potential issues**

1. **OwnerReference targets StatefulSet by name, but post-PR #415 the workload may be a Deployment.** `getWorkloadUID(ctx, namespace, "openshell-gateway")` — does this resolve StatefulSet, Deployment, or both? If a postgres-backed gateway uses a Deployment, `stsUID` will be empty and the Route will have no owner reference, so it won't be GC'd when the workload is deleted. Worth verifying that `getWorkloadUID` handles both workload types.

2. **NetworkPolicy hard-codes ports 8080/8081.** These match the current OpenShell defaults, but if they ever become configurable they'll need to be threaded through. A named constant or derivation from `gw` config would make this more robust.

3. **NetworkPolicy `from` has `namespaceSelector` only, no `podSelector`.** This lets any pod in `openshift-ingress` reach the gateway on those ports. In practice HAProxy router pods are what you want. This is fine for now (OCP doesn't guarantee consistent labels on router pods across versions), just noting it as a conscious trade-off.

4. **Silent skip when both `gw.Route.Host` is empty and `baseDomain` is unset.** The debug log is fine, but the reconciler returns `nil` without creating the Route. Any caller monitoring for the Route's existence won't get an error — they'll just see nothing. Consider whether a status condition or a warning-level log would be more observable here.

5. **Draft + unchecked test plan.** All test plan items are unchecked. Flagging for when this moves to ready-for-review.

Overall direction is right. The OwnerReference / workload-type question is the one I'd want answered before merging.

jsell-rh:
🤖 **Amber Review**

Adds two new OpenShift-specific reconcilers: `reconcileRouterNetworkPolicy()` creates a `NetworkPolicy` permitting HAProxy router pods to reach the gateway, and `reconcileOpenShiftRoute()` creates a passthrough `Route` for NLB-backed gRPC ingress. ClusterRole gains `route.openshift.io/routes` RBAC. Both resources cleaned up when `gw.Route` is removed.

**Findings:**
- **Major — Missing `ownerReference` on NetworkPolicy:** Per ACP checklist, OwnerReferences must be set on all new K8s child resources. `reconcileOpenShiftRoute()` sets an ownerReference correctly; `reconcileRouterNetworkPolicy()` does not set any. The NetworkPolicy will leak if the namespace is deleted or orphaned. Apply the same StatefulSet ownerReference pattern used by the Route reconciler.
- **Major — Silent skip on no-hostname instead of requeue:** When both `gw.Route.Host` is empty and `r.baseDomain` is empty, `reconcileOpenShiftRoute()` logs at `Debug` and returns `nil`. The reconciler sees no error, no warning, and the Route silently does not exist. Should return a sentinel error to trigger requeue, or at minimum log at `Warn` level.
- **Critical — Hard-coded GVK in ownerReference:** `getWorkloadUID` is called with `"openshell-gateway"` and the ownerReference hard-codes `apiVersion: "apps/v1"`, `kind: "StatefulSet"`. If the workload type ever changes, the ownerReference GVK will be wrong, leading to GC misbehaviour. Derive the GVK from the retrieved object rather than hard-coding it.
- **Minor — Missing RBAC for `networkpolicies`:** The ClusterRole diff only adds `route.openshift.io/routes` permissions. If the control plane service account doesn't already have NetworkPolicy CRUD rights, `reconcileRouterNetworkPolicy()` will fail at runtime with 403. Add `networking.k8s.io networkpolicies` verbs to the ClusterRole.
- **Minor — No feature flag gating:** Both reconcilers activate on all OpenShift clusters immediately. Per ACP standards, new features should be gated behind an Unleash flag for controlled rollout and emergency kill-switch.
- **Minor — `haproxy.router.openshift.io/timeout: "3600s"` hard-coded** — appropriate for long-lived gRPC streams but should have a comment explaining the value.

**Confidence:** High

---

PR #442, opened 2026-07-28
fix(kind): deterministic OpenShell gateway port allocation

jeremyeder:
## Problem

OpenShell gateway local port-forwards were allocated non-deterministically, and three code paths disagreed on the tenant→port mapping:

- `kind-status` and `setup-gateway-cli.sh` used a **fixed** index (`tenant-a`=15080, `tenant-b`=15081, …).
- `_kind-start-port-forward` used a **positional** offset over *whichever gateway pods happened to exist* at that instant, sorted.

The SSO step (`Makefile:1037-1039`) restarts port-forwards as soon as **ambient-ui** is ready — it never waited for the gateway StatefulSets. So if a tenant's gateway pod wasn't up yet, that tenant was skipped and the others shifted down onto its port. Result: `tenant-b` squatted on 15080 (tenant-a's canonical port), and `setup-gateway-cli.sh tenant-a` then failed with **"port in use"**.

This was surfaced by a user's kind debug bundle: `tenant-b`/`tenant-c` port-forwards existed on 15080/15081, no `tenant-a`, and `setup-gateway-cli.sh tenant-a` errored on 15080.

## Fix

Single source of truth for the mapping, `port(ns) = BASE + index(ns in OPENSHELL_TENANTS)`, used by all consumers:

- **New `scripts/lib/gateway-ports.sh`** — sourceable (`gateway_port_for`) and executable (`port-for <ns>` / `list`). Honors `KIND_FWD_GATEWAY_BASE_PORT` → `GATEWAY_BASE_PORT` → `15080`.
- **`_kind-start-port-forward`** now iterates the canonical `OPENSHELL_TENANTS` list, skips tenants without an `openshell-gateway` StatefulSet, and `kubectl wait --for=condition=Ready`s the gateway pods (90s, non-fatal) **before** forwarding on the fixed port. No more skip-and-shift, no more race.
- **`kind-status`** derives its "Gateways:" line from the library (byte-identical output).
- **`setup-gateway-cli.sh`** uses the library for the canonical port, with a positional fallback for namespaces outside `OPENSHELL_TENANTS`.

Net effect: `tenant-a` is always 15080 everywhere; ports never shift; the "port in use" collision can't recur.

## Tests

- **New `tests/unit/test_gateway_ports.sh`** — pure-bash, 13 assertions (all default mappings, base-port precedence, custom tenant lists, unknown-tenant failure, stderr hygiene). No cluster required.
- **CI**: `.github/component-paths.json` now gates `scripts/**`, `Makefile`, and `tests/unit/**` (and drops a stale `tests/test_model_discovery.py` glob → the real file is `tests/unit/test_model_discovery.py`); `unit-tests.yml`'s `scripts` job runs the new test.

## Verification

- `bash tests/unit/test_gateway_ports.sh` → 13/13 pass.
- `./scripts/lib/gateway-ports.sh list` and `make kind-status` agree: `tenant-a`=15080 … `codebase-maintainers`=15084.
- `make -n kind-port-forward` / `make -n kind-status` parse cleanly.
- Whole-branch review (independent, most-capable model): **APPROVE — 0 Critical, 0 Important.**
- CodeRabbit local review was **rate-limited** (free-plan window); the `pr-review-gate` hook treats rate-limit as non-blocking. Re-run recommended once the window resets or via CI.

## Optional follow-ups (non-blocking Minors)

- `unit-tests.yml` `scripts` job label still reads "model-discovery" — cosmetic.
- Unit test exercises the executable `port-for` path; a sourced-function assertion would close a small gap.
- Redundant `GATEWAY_BASE_PORT=…` env prefix in `setup-gateway-cli.sh` (function already reads it).

🤖 Authored and reviewed via agent workflow (subagent-driven development + independent whole-branch review). No Go files touched.

### DIFF
--- a/.github/component-paths.json
+++ b/.github/component-paths.json
@@ -44,7 +44,12 @@
     "label": "docs"
   },
   "scripts": {
-    "paths": [".github/scripts/**", "tests/test_model_discovery.py"],
+    "paths": [
+      ".github/scripts/**",
+      "scripts/**",
+      "Makefile",
+      "tests/unit/**"
+    ],
     "label": "scripts"
   }
 }
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -197,6 +197,9 @@ jobs:
       - name: Run tests
         run: python tests/unit/test_model_discovery.py -v
 
+      - name: Run gateway-ports unit tests
+        run: bash tests/unit/test_gateway_ports.sh
+
   summary:
     name: Unit Tests CI Gate
     runs-on: ubuntu-latest
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -198,5 +198,6 @@ When both the self-review and the hook pass, apply the `ambient-code:self-review
 ## Testing
 
 - **Runner tests**: `cd components/runners/ambient-runner && python -m pytest tests/`
+- **Gateway port mapping**: `bash tests/unit/test_gateway_ports.sh`
 
 ## Convention Authority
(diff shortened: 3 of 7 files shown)


YiqinZhang:
🤖 **Amber Review**

Fixes a race condition where OpenShell gateway port-forwards could land on wrong ports if a tenant's pod wasn't running at allocation time, causing other tenants to shift onto each other's canonical ports. Introduces `scripts/lib/gateway-ports.sh` as single source of truth for the `tenant → port` mapping (formula: `BASE + index_in_OPENSHELL_TENANTS`), adds `kubectl wait` before each port-forward, and includes 13-assertion pure-bash tests wired into CI.

**Findings:**
- **Minor:** `gateway-ports.sh` is invoked as a subprocess per-tenant in shell loops — sourcing the library once (as `setup-gateway-cli.sh` now does) would be more consistent. Not a blocker.
- **Minor:** `unit-tests.yml` `scripts` job label still reads "model-discovery" — acknowledged in PR body, can be cleaned up in a follow-up.
- **Minor:** `kubectl wait` failure is non-fatal (`|| echo "⚠ not ready after 90s; forwarding anyway"`) — intentional for dev UX but worth a comment so future maintainers know it is deliberate.

Root cause analysis is solid, fix is correct, test coverage is meaningful for a bash library.

**Confidence:** High

---

PR #443, opened 2026-07-28
test(e2e): add README technical-content e2e test + CI

jeremyeder:
## What

Adds an end-to-end test for the **technical content of `README.md`**, so any user who copy/pastes their way through the README succeeds. Environment-agnostic — point it at any running ACP environment (Kind, CRC, or a real cluster); it provisions nothing.

## Deliverables

- `tests/e2e/readme-e2e-test.sh` — the test (mirrors the existing `tests/e2e/vteam-catalog-lab-test.sh` house style: `pass/fail/skip/section`, `acpctl` discovery).
- `.github/workflows/readme-e2e.yml` — `static` job on PRs (touching `README.md`, `Makefile`, `components/ambient-cli/**`, or the script); `live` job on `workflow_dispatch` + weekly `schedule`.
- `Makefile` — `make test-readme` target (no cluster prereq, so it doubles as a linter).

## How it works

**Static gate (always, no cluster):**
- Relative links / referenced files resolve; in-page anchors exist.
- Badge workflows exist (`lint.yml`, `unit-tests.yml`, `docs.yml`).
- Every `make <target>` in the README exists in the `Makefile`.
- Every `acpctl <group> <sub>` (extracted only from ```` ```bash ```` blocks, so prose mentions don't false-positive) resolves via `acpctl <group> --help`.

**Live flow (only when a cluster is reachable; skips cleanly otherwise):**
- Runs the README's documented `acpctl create project` / `create gateway` / `get gateway` / `gateway setup-cli` against the environment, using an isolated throwaway project that is torn down in an EXIT trap.
- Route-based `setup-cli --print` is skipped on port-forward/cluster-DNS gateways (per the README's OpenShift-Local scoping) rather than failing.
- CRC path is validated statically only (not executable in CI).

Exits non-zero on any failure so CI gates on it.

## Testing

- Static gate: 32 passed / 0 failed (cluster forced unreachable).
- Live flow: verified end-to-end against a live ACP env (project + gateway created, retrieved, and deleted; no leaked resources).
- `shellcheck -S warning` and `actionlint` clean.

## Notes

- No secrets in the workflow; pinned action SHAs; `permissions: contents: read`; user input passed via `env` only.
- The script does not `eval` README content — it greps for validation and runs explicit `acpctl` commands with regex-constrained tokens.

### DIFF
--- a/.github/workflows/readme-e2e.yml
+++ b/.github/workflows/readme-e2e.yml
@@ -1,0 +1,130 @@
+name: README E2E
+
+# Verifies the technical content of README.md so any user copy/pasting through
+# it succeeds. Two jobs:
+#   static  - every PR touching the README or referenced surfaces. No cluster;
+#             validates links, referenced files, badge workflows, make targets,
+#             and acpctl subcommands.
+#   live    - manual / scheduled. Brings up a Kind ACP environment via the
+#             documented path, then runs the README's live command flow against
+#             it. The script itself is environment-agnostic; this job just
+#             supplies an environment.
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'README.md'
+      - 'Makefile'
+      - 'components/ambient-cli/**'
+      - 'tests/e2e/readme-e2e-test.sh'
+      - '.github/workflows/readme-e2e.yml'
+  workflow_dispatch:
+    inputs:
+      run_sandbox:
+        description: 'Exercise openshell sandbox create/list'
+        required: false
+        type: boolean
+        default: false
+  schedule:
+    # Weekly rot check against a fresh environment (Mondays 07:17 UTC).
+    - cron: '17 7 * * 1'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: readme-e2e-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  static:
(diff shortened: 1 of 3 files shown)


YiqinZhang:
🤖 **Amber Review**

Adds a 399-line e2e test script validating README.md in two modes: static gate (no cluster — checks relative links, anchors, badge workflows, Makefile targets, `acpctl` subcommands) and live flow (spins up Kind, exercises `acpctl create project/gateway`, tears down on EXIT). Wired into CI via new `readme-e2e.yml` workflow.

**Findings:**
- **Minor:** `actions/checkout` is pinned to `9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0` with comment `# v7.0.0` — latest widely-used stable is v4.x, not v7. Verify SHA matches the actual release tag to preserve the audit value of pinning.
- **Minor:** Live job runs on `schedule` without a pre-provisioned cluster — Kind bring-up + e2e on a cold GitHub runner against the 35-min timeout may be tight. Monitor first scheduled run.
- **Minor:** No `shellcheck` enforcement in CI for the new script (author confirms it was run locally and is clean — consider adding as a CI step to lock in the clean state).

Security posture is good: pinned SHAs, `permissions: contents: read`, no secrets in workflow, user input passed via `env` only, no `eval` of README content.

**Confidence:** High

---

PR #447, opened 2026-07-29
ENGPROD-10270 Add security skills

kdoberst:
## Summary

We want security skills from the internal `ai-security-harness` GitLab repo available alongside our existing agent skills, without blocking contributors who lack GitLab access and without committing upstream skill content into this repo.

Standard APM install/update could not do that on its own: the harness repo is private, not packaged as an APM module, and should be optional. A custom install and update flow was necessary — use `make apm-install` or `apm run install`, and `make apm-update` or `apm run update`.

**Why a custom install?** Install reads credentials from `.env.local`. With a GitLab token, it pulls the harness and installs everything. Without one, it skips the security skills and still installs the rest, so local setup and CI keep working.

**Why a separate sync step?** The harness repo stores skills in its own layout, not as an APM package. We download a pinned version, reshape it for APM, and stage it locally before install.

**Why commit lock files?** `apm.lock.yaml` pins Fleet SDLC dependency versions for reproducible installs. A separate harness lock file pins the security repo to a specific commit so everyone gets the same skills without checking them into git.

Setup is documented in `.env.local.example` and the README.

## Test plan

- [ ] `make apm-install` with no GitLab token — core skills install; security skills skipped with a clear message
- [ ] `make apm-install` with a GitLab token — security skills are available after install
- [ ] `make apm-update` — refreshes pinned versions and reinstalls skills end-to-end


kdoberst:
I'm currently getting the same error when I run `make local-test-quick` on main.  So the failure isn't due to this PR - it is ready for review while the error is addressed.

YiqinZhang:
🤖 **Amber Review**

Introduces optional APM install/update flow for pulling security skills from a private GitLab repo. When `GITLAB_APM_PAT` is absent, the GitLab dependency is stripped at runtime so `apm install` succeeds for all contributors. Commit-pinned lock file ensures reproducibility.

**Findings:**
- **Major:** `sync-ai-security-harness-skills.sh` passes `GITLAB_APM_PAT` via `GIT_HTTP_EXTRAHEADER` — if `set -x` debug tracing is enabled upstream, the token value will appear in logs. Enforce `set +x` around the `git clone/fetch` invocations.
- **Major:** `apm-install.sh` `sed`+`trap` pattern for `apm.yml` modification leaves config corrupted on `SIGKILL`. Atomic rename (`tmp` file + `mv`) would be safer.
- **Minor:** `ai-security-harness.lock.yaml` exposes an internal Red Hat hostname (`gitlab.cee.redhat.com`) in the public repo — assess if acceptable for this repo's visibility level.
- **Minor:** No cleanup trap in `sync-ai-security-harness-skills.sh` if it fails mid-run, leaving partial `vendor/ai-security-harness/` state.

Approving as the Major items are dev-tooling concerns, not production-path risks. These are worth hardening in a follow-up issue.

**Confidence:** High

---

PR #452, opened 2026-07-29
fix(acpctl): stop false token-expiry warning and honor insecure TLS on refresh

jeremyeder:
## Problem
After `acpctl login --use-auth-code`, the CLI printed `Warning: token expires soon` on **every** login. The threshold is 24h (`login/cmd.go`), but OIDC access tokens are intentionally short-lived (~5 min) and auto-refreshed via the stored refresh token — so the warning was a false alarm that reads as "your login is about to break."

While investigating, found a latent bug in the refresh path: `RefreshAccessToken` built a plain `http.Client` that **ignored `insecure_tls_verify`**, so silent refresh fails against issuers with untrusted certs even when the user opted out of verification (e.g. self-hosted Keycloak on a cluster route). Both are in the CLI auth/token subsystem and ship together.

## Changes
- **Login message**: gate the expiry warning on there being **no** refresh token. When a refresh token is stored, print an informational line instead (`Access tokens are refreshed automatically using the stored refresh token.`). Extracted into a testable `writeTokenExpiryNotice` helper. `login/cmd.go` + new `expiry_notice_test.go`.
- **TLS on refresh**: `RefreshAccessToken` now takes `insecureTLSVerify bool` and builds the transport with `InsecureSkipVerify` when set, threaded from `cfg.InsecureTLSVerify` in `GetTokenWithRefresh`. `pkg/config/token.go` + `config.go`.

## Security note
The `InsecureSkipVerify` branch is **opt-in and off by default** — gated behind the pre-existing `insecure_tls_verify` config / `--insecure-skip-tls-verify` flag that the SDK client (`pkg/connection`) already honors via `WithInsecureSkipVerify()`. This change only makes the refresh path consistent with that existing, user-controlled behavior; it does not add a new capability or change defaults. `//nolint:gosec` is scoped to that line with justification.

## Testing
- `gofmt`, `go vet`, `golangci-lint run` (0 issues) on changed packages
- `go test ./pkg/config/... ./cmd/acpctl/login/...` — pass
- New tests: refresh-present / expires-soon / already-expired / opaque-token; insecure-TLS success vs verification failure against a self-signed `httptest.NewTLSServer`.

### DIFF
--- a/components/ambient-cli/cmd/acpctl/login/cmd.go
+++ b/components/ambient-cli/cmd/acpctl/login/cmd.go
@@ -3,6 +3,7 @@ package login
 
 import (
 	"fmt"
+	"io"
 	"net/url"
 	"time"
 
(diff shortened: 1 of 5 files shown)


YiqinZhang:
🤖 **Amber Review**

Two related CLI auth fixes: (1) gates the false "token expires soon" warning on absence of a refresh token — OIDC access tokens are intentionally short-lived; (2) threads `insecure_tls_verify` through `RefreshAccessToken` so silent refresh works against self-hosted issuers with untrusted certs.

**Findings:**
- **Minor:** `expiry_notice_test.go` uses `strings.Contains(tc.name, "short-lived")` dispatch inside the loop to set the token — test setup logic dependent on test name strings is brittle. Prefer an explicit `accessToken` field in the test table. Can be addressed in a follow-up.
- **Minor:** `//nolint:gosec // G402: opt-in via insecure_tls_verify` annotation is correctly scoped and justified. No action needed.

The `InsecureSkipVerify` path is correctly opt-in, off-by-default, narrowly scoped. Token content is not logged. Error wrapping follows `fmt.Errorf("context: %w", err)`. Well-tested.

**Confidence:** High

---

PR #454, opened 2026-07-30
ENGPROD-10159 - Implement feedback button/form with Slack integration

rh-amarin:
- Add persistent feedback strip (right-edge vertical tab, z-9999) and shadcn Dialog for users to send free-text
  feedback to a configured Slack channel
  - BFF route (`GET`/`POST /api/feedback`) with SSO authentication, input validation, in-memory rate limiting
  (5/user/min), and Slack Block Kit message delivery
  - `FEEDBACK_SLACK_WEBHOOK_URL` env var  — strip auto-hides when unconfigured
  - Ops deployment docs (`docs/src/content/docs/features/platform-feedback.md`) covering configuration, Slack
  webhook setup, kind/OpenShift deployment, and troubleshooting
  - `RECONCILE.md` updated: Wave 18 (U9–U16), coverage 92.2% → 92.4%
  - `package-lock.json` regenerated for Node 22 container compatibility (npm 10 vs npm 11 lockfile mismatch)

  ## Test plan

  - [ ] Set `FEEDBACK_SLACK_WEBHOOK_URL` on a kind cluster and verify the feedback strip appears
  - [ ] Submit feedback and verify it arrives in the configured Slack channel with Block Kit formatting
  - [ ] Unset the env var and verify the strip is hidden
  - [ ] Verify submit is disabled when textarea is empty
  - [ ] Verify rate limiting: submit 6 times rapidly and confirm 429 with "too quickly" message
  - [ ] Verify unauthenticated requests return 401
  - [ ] Verify keyboard accessibility: Tab to strip, Enter to open, focus moves to textarea
  - [ ] `npx tsc --noEmit` — zero new errors
  - [ ] `podman build` of ambient-ui succeeds (lockfile fix)


jsell-rh:
🤖 **Amber Review**

Adds a persistent feedback UI strip with Slack webhook integration. BFF Next.js route authenticates via SSO session, applies in-memory rate limiting (5/user/min), and posts to a configurable Slack webhook using Block Kit. Strip auto-hides when webhook URL is absent.

**Findings:**
- **Critical — Shadcn violation:** The feedback trigger uses a raw `<button>` element with custom Tailwind classes. ACP standard: **Shadcn UI components only — no custom buttons.** `Button` is already imported in the file but only used inside the Dialog. The trigger tab button must also be a Shadcn `Button`.
- **Major — Manifest uses `value: ""` instead of `secretKeyRef` for webhook URL:** `FEEDBACK_SLACK_WEBHOOK_URL` is added as a hardcoded empty `value: ""` in `ambient-ui-deployment.yaml`. The PR description states this should be sourced from the `sso-credentials` Secret with `optional: true`, but the manifest does not reflect this. A plain env var value is also a risk if someone accidentally commits a real URL. Use `secretKeyRef` with `optional: true`.
- **Major — Rate limiting bypassed via opaque/non-JWT tokens:** `resolveUsername` parses the JWT access token for `preferred_username` — if the token is opaque, `username` silently falls back to `"unknown"`. Rate limiting keyed on `"unknown"` allows unlimited requests from any opaque-token caller. Key on a session identifier instead.
- **Major — `pagePath` not sanitized before Slack message:** `pagePath` is passed unsanitized from the request body into the Block Kit message. Only `text` is validated. Add a length cap and character sanitization.
- **Minor — In-memory `rateLimitStore` doesn't survive pod restarts or scale across replicas.** Acceptable for MVP but should be documented as a known limitation.

**Confidence:** High

---

jsell-rh:
🤖 **Amber Follow-up — PR #454 (head `7acfae20`)**

The rate-limiting removal is fine — SSO auth is sufficient justification for an internal tool, and Slack's own webhook limits provide an external bound. That finding is resolved.

Three issues remain from the prior review:

**Critical — No input length cap on `text`:**
`route.ts` accepts unbounded `body.text` with no `maxLength` guard. A user can submit a 1 MB string that gets forwarded to Slack verbatim, likely hitting Slack's webhook payload limit (3 000 characters for `text` blocks) and causing a silent delivery failure. Add a server-side length check before constructing the Slack payload:
```typescript
if (text.length > 3000) {
  return Response.json({ error: "Feedback must be 3000 characters or fewer" }, { status: 400 })
}
```

**Critical — `pagePath` not sanitized before Slack injection:**
`pagePath` is taken as-is from the request body and embedded in the Slack `context.elements.mrkdwn` field:
```typescript
text: `*From:* ${username} · *Page:* ${pagePath} · ...`
```
A user can inject Slack mrkdwn formatting (bold, links, `<!here>`) to make the Slack message look misleading. At minimum, strip or escape `<`, `>`, `&`, and `*` characters from `pagePath`, and apply a length cap (e.g., 200 chars).

**Major — Raw `<button>` in sidebar trigger (Shadcn convention violation):**
The fixed-position sidebar tab (`feedback-strip.tsx`) uses a raw `<button>` element with manual Tailwind classes. CLAUDE.md requires Shadcn UI components only — no custom buttons. Replace with `<Button variant="ghost">` (or `asChild` pattern) with appropriate `className` overrides for positioning. The Dialog actions already use Shadcn `<Button>` correctly.

**Minor — Stale references to removed rate limiting:**
`specs/ui/platform-feedback.spec.md` still has a Rate Limiting requirements section describing the removed implementation. `skills/RECONCILE.md` row U11 lists "rate limiting (5/user/min)" as a done requirement. Both should be updated to reflect the current design.

**Confidence:** [90%] High — the input validation gaps are straightforward to fix.

*— Amber*

YiqinZhang (approved):
🤖 **Amber Re-Review — PR #454: ENGPROD-10159 feat(feedback): implement feedback button/form with Slack integration**

Re-reviewing at head `20f9d896c98` — all three issues from my last review are addressed.

## Previous Findings — Status

- ✅ **No input length cap (Critical):** Server-side 3000-character cap added: `if (text.length > 3000) { return Response.json({ error: "Feedback must be 3000 characters or fewer" }, { status: 400 }) }`. Addressed.
- ✅ **`pagePath` unsanitized (Critical):** Truncated to 200 chars and `<>&*` stripped before use in the Slack payload: `rawPath.slice(0, 200).replace(/[<>&*]/g, "")`. Not logged. Addressed.
- ✅ **Raw `<button>` tag (Major — Shadcn violation):** All interactive elements now use `import { Button } from '@/components/ui/button'` — the strip trigger, Cancel, and Submit all use Shadcn `<Button>`. Addressed.

## New Observations (Non-blocking)

**N1. No `maxLength` on `<Textarea>` (Nit):** The server enforces 3000 chars and returns a 400 if exceeded, but the `<Textarea>` has no `maxLength` prop — users won't get a native browser hard-stop until they hit Submit. Adding `maxLength={3000}` would give a visible character limit in-browser. Minor UX improvement, not a blocker.

**N2. Backtick/underscore/tilde survive `pagePath` sanitization (Nit):** The regex `/[<>&*]/g` covers the most dangerous Slack mrkdwn characters but misses `` ` `` (inline code), `_` (italic), and `~` (strikethrough). A URL like `/docs/~foo_bar` could affect Slack message rendering. Cosmetic only — no security risk. Worth addressing before enabling in production.

**N3. Slack webhook URL injected as plaintext env var (Design note):** The current approach uses `kubectl set env` which works, and the docs correctly call this out. A Kubernetes `Secret` reference in the deployment manifest would be the hardened pattern per project convention. Fine to follow up separately.

## Confidence

**[90%]** High — all blockers resolved; new observations are nits. The feature is well-implemented and spec-compliant.

---
*— Amber*

PR #456, opened 2026-07-30
ENGPROD-10220 - fix pre-commit hook golangci-lint to work on macOS

rh-amarin:
## Summary

  - Replace `declare -A` associative array with a delimited-string pattern for tracking affected Go modules
  - `declare -A` is a bash 4+ feature; macOS ships bash 3.2 by default, causing the hook to fail on every commit
  for Mac developers

  ## Test plan

  - [ ] Run `git commit` touching a Go file on macOS — hook runs and lints only the affected module
  - [ ] Run `git commit` touching a Go file on Linux — no regression
  - [ ] Run `git commit` with no Go files — hook exits 0 (skips)

### DIFF
--- a/scripts/pre-commit/golangci-lint.sh
+++ b/scripts/pre-commit/golangci-lint.sh
@@ -20,21 +20,28 @@ GO_MODULES=(
 )
 
 # Determine which modules are affected by the staged files
-declare -A affected_modules=()
+affected_modules=""
 for file in "$@"; do
     for mod in "${GO_MODULES[@]}"; do
         if [[ "$file" == "$mod/"* || "$file" == "$REPO_ROOT/$mod/"* ]]; then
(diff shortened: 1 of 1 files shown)


YiqinZhang:
🤖 **Amber Review**

Fixes `declare -A` (Bash 4 associative arrays) incompatibility on macOS (ships Bash 3.2). Replaces with a delimited-string `|mod|` pattern — standard POSIX-compatible alternative. Single-file shell script change, no security or architecture concerns.

**Findings:**
- **Minor:** Script still uses `[[ ]]` and other bash-isms — the fix is scoped to remove the specific Bash 4 dependency, not full POSIX-ification. Acceptable and intentional.

Correct, minimal, and well-reasoned fix.

**Confidence:** High

---

PR #459, opened 2026-07-30
docs(deploy): document gateway-trusted-ca for OpenShift OIDC

bsquizz:
## Summary
- Documents the required `gateway-trusted-ca` ConfigMap setup for OpenShift clusters using gateway OIDC (Keycloak)
- Adds troubleshooting entry for the "OIDC key refresh failed" / "JWKS fetch failed" error
- Documents the privileged SCC requirement for sandbox pods on OpenShift
- Updates deployment README index with links to the new sections

## Context

**gateway-trusted-ca (8c56a6c4)**

On OpenShift, the gateway container image (scratch/distroless) has no system CA bundle. When OIDC is configured against a Keycloak instance exposed via an OpenShift Route, the gateway cannot complete the TLS handshake to fetch JWKS keys — the ingress CA that signed the Route's certificate is not trusted.

The control plane already supports injecting a trusted CA via the `gateway-trusted-ca` ConfigMap (`SSL_CERT_FILE` + volume mount), but this prerequisite was undocumented.

**privileged SCC for sandbox pods (facf4638)**

On OpenShift, the `openshell-sandbox` service account needs the privileged SCC in each tenant namespace or sandbox pods fail to start. This requirement was undocumented.

## Test plan
- [ ] Verify the `oc get configmap router-ca` command works on a ROSA/OCP cluster
- [ ] Confirm the ConfigMap name (`gateway-trusted-ca`) and key (`ca-bundle.crt`) match the constants in `gateway_reconciler.go`
- [ ] Verify the `oc adm policy` command for the privileged SCC works on a ROSA/OCP cluster
- [ ] Verify links in the deployment README resolve correctly

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
@@ -69,6 +69,7 @@ var networkPolicyGVR = schema.GroupVersionResource{
 	Resource: "networkpolicies",
 }
 
+// Cross-ref: docs/internal/deployment/OPENSHIFT_DEPLOY.md § "Gateway Trusted CA"
 const (
 	trustedCAConfigMapName = "gateway-trusted-ca"
 	trustedCAKey           = "ca-bundle.crt"
--- a/docs/internal/deployment/OPENSHIFT_DEPLOY.md
+++ b/docs/internal/deployment/OPENSHIFT_DEPLOY.md
@@ -110,6 +110,20 @@ oc logs -l app.kubernetes.io/name=openshell -n <tenant-namespace> | grep -i oidc
 > **Note:** If the cluster's ingress CA rotates, update the `gateway-trusted-ca`
 > ConfigMap and restart the gateway pods.
 
+### Sandbox SCC (Required on OpenShift)
+
+When a namespace that will host gateway sandboxes is created on OpenShift, the
+`openshell-sandbox` service account must be granted the `privileged` SCC. Without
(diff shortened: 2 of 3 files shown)


JGiola:
Can we add a configuration where we can set the name of the ConfigMap to load? copying the value from the injected ones to another is prone to having it rot. What if the CA inside it expire? or is rotated for security? we will have downtime, if we point to the one managed automatically by the OpenShift automations we are sure to always have the latest one...

YiqinZhang:
🤖 **Amber Review**

Pure documentation addition filling a real operational gap. Adds `gateway-trusted-ca` ConfigMap setup instructions and troubleshooting steps for OpenShift OIDC with self-signed certs.

**Findings:**
- **Minor:** No code link from `gateway_reconciler.go` constants to the docs — if the ConfigMap name/key ever changes, docs will silently drift. Worth a cross-reference comment.
- **Minor:** `oc get configmap router-ca -n openshift-config-managed` may not apply to all OCP variants (e.g., ROSA with STS) — a brief variant note would help.

No code changes, no security surface touched. Content accurately reflects the running implementation.

**Confidence:** High

---

bsquizz:
> Can we add a configuration where we can set the name of the ConfigMap to load? copying the value from the injected ones to another is prone to having it rot. What if the CA inside it expire? or is rotated for security? we will have downtime, if we point to the one managed automatically by the OpenShift automations we are sure to always have the latest one...

Let's write a new spec for this if we want to change the behavior. I'll keep this PR to simply be a doc PR

PR #461, opened 2026-07-30
fix: get test-local-dev-simulation CI green

squizzi:
## Summary

Fixes to get the `test-local-dev-simulation` CI workflow green across Gateway E2E, OpenShell CLI E2E, and local Kind cluster setup.

**Control plane fixes:**
- Fix ndots:5 DNS workaround — delete sandbox pod for recreation from the patched CR instead of futile in-place sed, since the Sandbox CRD doesn't support `podTemplate.spec.dnsConfig`
- Fix static token auth — set a username on the static token path so the RBAC interceptor can auto-provision a service account, resolving gRPC "authorization service unavailable" errors
- Fix gateway NetworkPolicy to allow sandbox pods with the new `agents.x-k8s.io/sandbox-name-hash` label scheme (not just the old `openshell.ai/managed-by=openshell`)

**API server fix:**
- Redact `Authorization` and `X-Forwarded-Access-Token` headers in request logs

**CI workflow fix:**
- Map `image-tag` and `built-images` inputs to env vars in `test-local-dev-simulation` — they were never wired up, so CI always deployed `:latest` instead of the PR-built image

**E2E test fixes:**
- Align sandbox pod naming with control plane convention (`default--session-<first11chars>` not `session-<first40chars>`)
- Preload gateway/postgres images into Kind to avoid `ImagePullBackOff`
- Support both Deployment and StatefulSet gateway workloads in assertions
- Auto-restart dead port-forwards between test sections
- Add `eventually` retries around sandbox exec calls for supervisor relay startup lag
- Clean stale sessions before network-policy tests to avoid reusing sessions from prior `--skip-cleanup` runs
- Refactor all silent polling loops into `wait_for()` helpers with progress dots
- List failed test names in the results summary
- Strip `kubectl` "Defaulted container" stderr noise from `CMD_OUTPUT`
- Don't fail prereqs when podman is absent but docker is available

## Test plan

- [ ] `test-local-dev-simulation` CI workflow passes (Gateway E2E, OpenShell CLI E2E, RBAC E2E, Scheduled Sessions E2E)
- [ ] Control plane reconciler unit tests pass
- [ ] Local Kind cluster deploys successfully with `make kind-up`
- [ ] Re-running e2e tests after `--skip-cleanup` no longer hits stale session reuse

### DIFF
--- a/.github/workflows/test-local-dev.yml
+++ b/.github/workflows/test-local-dev.yml
@@ -247,8 +247,9 @@ jobs:
           sleep 5
         done
 
-        # Wait for gateway StatefulSet pod to be ready (not the certgen Job pod)
-        kubectl wait --for=condition=ready pod/openshell-gateway-0 \
+        # Wait for gateway pod to be ready (Deployment, not the certgen Job pod)
+        kubectl wait --for=condition=ready pod \
+          -l app.kubernetes.io/name=openshell,app.kubernetes.io/component=gateway \
           -n tenant-a --timeout=120s || echo "Warning: gateway pod not ready after 120s"
 
         PF_DIR=/tmp/ambient-code ./scripts/setup-gateway-cli.sh tenant-a
--- a/.github/workflows/test-local-dev.yml
+++ b/.github/workflows/test-local-dev.yml
@@ -60,6 +60,8 @@ jobs:
     env:
       KUBECTL_VERSION: v1.31.4
       KIND_VERSION: v0.31.0
+      IMAGE_TAG: ${{ inputs.image-tag || 'latest' }}
+      BUILT_IMAGES: ${{ inputs.built-images || '[]' }}
 
     steps:
     - name: Checkout code
--- a/components/ambient-api-server/pkg/middleware/auth_header_sanitize.go
+++ b/components/ambient-api-server/pkg/middleware/auth_header_sanitize.go
@@ -1,0 +1,40 @@
+package middleware
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	pkgserver "github.com/openshift-online/rh-trex-ai/pkg/server"
+)
+
+func init() {
+	pkgserver.RegisterPreAuthMiddleware(sanitizeAuthHeaders)
+}
+
+func sanitizeAuthHeaders(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		auth := r.Header.Get("Authorization")
+		fwd := r.Header.Get(forwardedAccessTokenHeader)
+		if auth == "" && fwd == "" {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		h := r.Header.Clone()
+		if auth != "" {
+			if token, ok := strings.CutPrefix(auth, "Bearer "); ok {
+				h.Set("Authorization", fmt.Sprintf("Bearer [REDACTED:len=%d]", len(token)))
+			} else {
+				h.Set("Authorization", fmt.Sprintf("[REDACTED:len=%d]", len(auth)))
+			}
+		}
+		if fwd != "" {
+			h.Set(forwardedAccessTokenHeader, fmt.Sprintf("[REDACTED:len=%d]", len(fwd)))
+		}
+
+		r2 := r.WithContext(r.Context())
+		r2.Header = h
+		next.ServeHTTP(w, r2)
+	})
+}
--- a/components/ambient-api-server/pkg/middleware/bearer_token_grpc.go
+++ b/components/ambient-api-server/pkg/middleware/bearer_token_grpc.go
@@ -11,6 +11,14 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+// staticTokenUsername is the identity assigned to callers authenticating with
+// the shared static bearer token (e.g. the control plane, bootstrap jobs).
+// autoProvisionServiceAccount in pkg/rbac grants this identity a global
+// platform:admin RoleBinding — the same mechanism used for OIDC service
+// accounts. Access is scoped to the API server's RBAC evaluator; it does
+// NOT grant K8s cluster privileges.
+const staticTokenUsername = "static-service-token"
+
 var grpcBypassMethods = map[string]bool{
 	"/grpc.health.v1.Health/Check":                                   true,
 	"/grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo": true,
@@ -26,7 +34,9 @@ func bearerTokenGRPCUnaryInterceptor(expectedToken, serviceAccountUsername strin
 			if authHeader := md.Get("authorization"); len(authHeader) > 0 {
 				if token, err := extractBearerToken(authHeader[0]); err == nil {
 					if subtle.ConstantTimeCompare([]byte(token), []byte(expectedToken)) == 1 {
-						return handler(WithCallerType(ctx, CallerTypeService), req)
+						ctx = WithCallerType(ctx, CallerTypeService)
+						ctx = auth.SetUsernameContext(ctx, staticTokenUsername)
+						return handler(ctx, req)
 					}
 					if username := usernameFromJWT(token); username != "" {
 						if isServiceAccount(username, serviceAccountUsername) {
@@ -52,7 +62,9 @@ func bearerTokenGRPCStreamInterceptor(expectedToken, serviceAccountUsername stri
 			if authHeader := md.Get("authorization"); len(authHeader) > 0 {
 				if token, err := extractBearerToken(authHeader[0]); err == nil {
 					if subtle.ConstantTimeCompare([]byte(token), []byte(expectedToken)) == 1 {
-						return handler(srv, &serviceCallerStream{ServerStream: ss, ctx: WithCallerType(ss.Context(), CallerTypeService)})
+						ctx := WithCallerType(ss.Context(), CallerTypeService)
+						ctx = auth.SetUsernameContext(ctx, staticTokenUsername)
+						return handler(srv, &serviceCallerStream{ServerStream: ss, ctx: ctx})
 					}
 					if username := usernameFromJWT(token); username != "" {
 						ctx := auth.SetUsernameContext(ss.Context(), username)
--- a/components/ambient-control-plane/internal/gateway/reconciler.go
+++ b/components/ambient-control-plane/internal/gateway/reconciler.go
@@ -604,7 +604,7 @@ func buildGatewayDeployment(nsConfig NamespaceConfig, defaultImage string, opts 
 						},
 						"volumes": []interface{}{
 							map[string]interface{}{"name": "gateway-config", "configMap": map[string]interface{}{"name": "openshell-gateway-config"}},
-							map[string]interface{}{"name": "sandbox-jwt", "secret": map[string]interface{}{"secretName": "openshell-gateway-jwt-keys", "defaultMode": int64(256)}},
+							map[string]interface{}{"name": "sandbox-jwt", "secret": map[string]interface{}{"secretName": "openshell-gateway-jwt-keys", "defaultMode": int64(0440)}},
 							map[string]interface{}{"name": "tls-cert", "secret": map[string]interface{}{"secretName": "openshell-server-tls"}},
 							map[string]interface{}{"name": "tls-client-ca", "secret": map[string]interface{}{"secretName": "openshell-server-tls", "items": []interface{}{map[string]interface{}{"key": "ca.crt", "path": "ca.crt"}}}},
 							map[string]interface{}{"name": "tmp", "emptyDir": map[string]interface{}{}},
--- a/components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
@@ -1389,6 +1389,16 @@ func (r *GatewayReconciler) reconcileSandboxNetworkPolicy(ctx context.Context, n
 									},
 								},
 							},
+							map[string]interface{}{
+								"podSelector": map[string]interface{}{
+									"matchExpressions": []interface{}{
+										map[string]interface{}{
+											"key":      "agents.x-k8s.io/sandbox-name-hash",
+											"operator": "Exists",
+										},
+									},
+								},
+							},
 						},
 						"ports": []interface{}{
 							map[string]interface{}{
--- a/components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
@@ -74,6 +74,16 @@ const (
 	trustedCAKey           = "ca-bundle.crt"
 )
 
+type GatewayReconcilerOption func(*GatewayReconciler)
+
+// WithAuthModeCallback registers a function called after each gateway is
+// reconciled, informing consumers whether the namespace uses OIDC auth.
+func WithAuthModeCallback(fn func(namespace string, hasOIDC bool)) GatewayReconcilerOption {
+	return func(r *GatewayReconciler) {
+		r.onNamespaceAuthMode = fn
+	}
+}
+
 type GatewayReconciler struct {
 	factory                    *SDKClientFactory
 	dynamicClient              dynamic.Interface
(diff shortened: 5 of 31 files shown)


jsell-rh:
🤖 **Amber Review**

Fixes three independent CI failures in `test-local-dev-simulation`: sets a username on the static token auth path, replaces broken pod-delete ndots fix with in-place `sed` exec, and adds `imagePullPolicy`/`securityContext` patches to the `kind-local` kustomize overlay.

**Findings:**
- **Blocker — Security context relaxation without environment guard:** `security-context-patch.yaml` in the `kind-local` overlay sets `runAsNonRoot: false` and `readOnlyRootFilesystem: false` on the `bootstrap-admin` Job. This directly contradicts the ACP security requirement for hardened container contexts. There is no check preventing this overlay from being applied to a non-local environment, and no comment marking it dev-only. A prominent comment is required at minimum; a CI guard to prevent non-Kind usage is strongly recommended.
- **Critical — Static token auto-provisioning is a magic identity:** `staticTokenUsername = "static-service-token"` is injected into the gRPC context so the RBAC interceptor auto-provisions a service account. The PR does not document what RBAC permissions that provisioned account receives — any caller presenting the static bearer token is auto-granted a service account under that identity. Scope and permissions of this auto-provisioned account must be explicitly constrained and documented.
- **Major — `verifyAndFixDNSConfig` exec has no pod identity validation:** The `sed` exec approach is functional but has a race between the `cat` check and the `sed` exec if the pod is replaced. Worth either scoping the exec more tightly or noting the race in a comment.
- **Major — Unexplained RBAC rule count changes:** `project_reconciler_test.go` updates expected counts (3→7, 6→9) and adds new API groups without a visible production code change in the diff. These changes should be traceable to an explicit policy decision.

**Confidence:** High

---

jsell-rh:
🤖 **Amber Re-Review — PR #461: fix: get test-local-dev-simulation CI green**

Re-reviewing at head `ef67fdfab1fe` — the four findings from my earlier review are addressed, but a new inconsistency surfaced between the stream and unary gRPC interceptors.

---

## Previous Findings — Status

- ✅ **Security context relaxation (Blocker):** `security-context-patch.yaml` now has a prominent `# DEV-ONLY` block comment with "DO NOT apply this patch outside of local Kind development." Addressed.
- ✅ **Static token auto-provisioning (Critical):** `staticTokenUsername` now has a doc comment explaining the identity, that `autoProvisionServiceAccount` grants `platform:admin`, and that it does NOT grant K8s cluster privileges. Addressed.
- ✅ **`verifyAndFixDNSConfig` exec race (Critical):** Race safety note added. Addressed.
- ✅ **Unexplained RBAC rule counts (Major):** Count changes (3→7, 6→9) are traceable to the new auth middleware and static-service-token auto-provisioning introduced by this PR. Addressed.

---

## New Finding

**Critical — Unary gRPC interceptor drops `CallerTypeService` from context**

The stream and unary interceptors now handle the static bearer token differently:

**Stream interceptor (correct):**
```go
ctx := WithCallerType(ss.Context(), CallerTypeService)  // ✅ CallerType set
ctx = auth.SetUsernameContext(ctx, staticTokenUsername)  // ✅ Username set
return handler(srv, &serviceCallerStream{ServerStream: ss, ctx: ctx})
```

**Unary interceptor (regression):**
```go
ctx = auth.SetUsernameContext(ctx, staticTokenUsername)  // ✅ Username set
return handler(ctx, req)                                  // ❌ CallerType NOT set
```

The original unary interceptor called `WithCallerType(ctx, CallerTypeService)` inline in the return (the subsequent `ctx = WithCallerType(...)` was dead code after the return). The refactor dropped it. Any code that calls `GetCallerType(ctx)` and expects `CallerTypeService` for static-bearer-token callers on unary RPCs will now get the zero value instead.

Fix: add `ctx = WithCallerType(ctx, CallerTypeService)` before the `SetUsernameContext` call in the unary interceptor, mirroring the stream interceptor.

---

## Incidental Observation (No action required)

`gateway_reconciler.go` includes a NetworkPolicy rule for `agents.x-k8s.io/sandbox-name-hash` (OpenShell 0.0.92 label support) not mentioned in the PR description. Noting it so reviewers are aware this is intentional and not accidental.

---

## Confidence

**[90%]** High — the CallerType omission in the unary path is a clear inconsistency with the stream path; the fix is mechanical. Happy to re-review once addressed.

---

*— Amber*

components/ambient-api-server/pkg/middleware/auth_header_sanitize.go
```diff
@@ -0,0 +1,40 @@
...
+
+	pkgserver "github.com/openshift-online/rh-trex-ai/pkg/server"
+)
+
+func init() {
+	pkgserver.RegisterPreAuthMiddleware(sanitizeAuthHeaders)
+}
+
+func sanitizeAuthHeaders(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
squizzi, line 16: Good defense-in-depth. I verified the rh-trex-ai framework middleware chain: `RegisterPreAuthMiddleware` wraps the router *before* `jwtHandler` wraps the result (line 39-41 in `apiserver.go`), so JWT runs first in the execution path. By the time `sanitizeAuthHeaders` fires, the JWT handler has already extracted the real token and stored claims in the request context via `ContextAuthKey`. Downstream handlers (including `GetK8sClientsForRequest`) read from context, not headers — so this redaction is safe and only affects handlers that might accidentally log `r.Header`.

Constitution Principle II compliant: tokens never reach router handlers in plaintext.

components/ambient-api-server/pkg/middleware/bearer_token_grpc.go
```diff
@@ -11,6 +11,14 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+// staticTokenUsername is the identity assigned to callers authenticating with
+// the shared static bearer token (e.g. the control plane, bootstrap jobs).
+// autoProvisionServiceAccount in pkg/rbac grants this identity a global
+// platform:admin RoleBinding — the same mechanism used for OIDC service
+// accounts. Access is scoped to the API server's RBAC evaluator; it does
```
squizzi, line 18: Correct fix. The RBAC interceptor's `autoProvisionServiceAccount` requires a username on the context to create the platform:admin RoleBinding. Without this, static token callers (control plane, bootstrap jobs) hit "authorization service unavailable" because the interceptor can't resolve who they are.

The constant documentation is helpful — it makes the privilege model explicit without needing to trace through the RBAC code.

components/ambient-control-plane/internal/gateway/reconciler.go
```diff
@@ -604,7 +604,7 @@ func buildGatewayDeployment(nsConfig NamespaceConfig, defaultImage string, opts
 						},
 						"volumes": []interface{}{
 							map[string]interface{}{"name": "gateway-config", "configMap": map[string]interface{}{"name": "openshell-gateway-config"}},
-							map[string]interface{}{"name": "sandbox-jwt", "secret": map[string]interface{}{"secretName": "openshell-gateway-jwt-keys", "defaultMode": int64(256)}},
+							map[string]interface{}{"name": "sandbox-jwt", "secret": map[string]interface{}{"secretName": "openshell-gateway-jwt-keys", "defaultMode": int64(0440)}},
```
squizzi, line 607: nit: `int64(0440)` is a Go octal literal that equals 288 decimal (mode `-r--r-----`). The paired change in `statefulset.yaml` uses the decimal form `288`. Both are correct, but future readers comparing the two files may not immediately see they match. Consider adding a trailing comment like `// 0440 octal = owner+group read` — or using `int64(288)` here for consistency with the YAML. Not blocking.

components/ambient-control-plane/internal/reconciler/gateway_reconciler.go
```diff
@@ -1389,6 +1389,16 @@ func (r *GatewayReconciler) reconcileSandboxNetworkPolicy(ctx context.Context, n
 									},
 								},
 							},
+							map[string]interface{}{
+								"podSelector": map[string]interface{}{
+									"matchExpressions": []interface{}{
```
squizzi, line 1415: Good addition. The `agents.x-k8s.io/sandbox-name-hash` label is the new naming convention from the upstream Sandbox CRD, while `openshell.ai/managed-by=openshell` is the legacy scheme. Supporting both in the NetworkPolicy ingress rules ensures sandbox pods can reach the gateway regardless of which controller version created them.

components/ambient-control-plane/internal/reconciler/kube_reconciler.go
```diff
@@ -583,6 +583,11 @@ func (r *SimpleKubeReconciler) injectACPInternalPolicy(ctx context.Context, name
 }
 
 // Workaround for https://github.com/NVIDIA/OpenShell/issues/2053
+// The Sandbox CRD supports spec.podTemplate.spec.dnsConfig. This patch sets
```
squizzi, line 586: These comments earn their keep — the interaction between CR patching, pod lifecycle, and the controller's recreation behavior is non-obvious. The race safety note (pod replaced between cat and delete, NotFound swallowed, poll loop re-verifies) is exactly the kind of thing that would trip up a future maintainer without documentation.

components/manifests/overlays/kind-local/security-context-patch.yaml
```diff
@@ -0,0 +1,13 @@
+# DEV-ONLY: Relaxes security constraints for the bootstrap-admin Job when
+# running locally-built images in Kind. Required because local images run as
+# root and write to the filesystem during bootstrap. This overlay is scoped
+# to overlays/kind-local/ which is only applied by `make kind-up` — it is
+# never referenced by production, staging, or CI overlays.
+#
+# DO NOT apply this patch outside of local Kind development.
+- op: replace
+  path: /spec/template/spec/securityContext/runAsNonRoot
+  value: false
```
squizzi, line 10: Constitution checkpoint: `runAsNonRoot: false` and `readOnlyRootFilesystem: false` violate Principle II SecurityContext requirements. However, this is correctly scoped to `overlays/kind-local/` which is only referenced by `make kind-up`. The header comment makes the dev-only constraint explicit. Verified that no production, staging, or CI kustomization.yaml files reference this overlay.

Suggestion: consider adding a CI lint rule that greps for `runAsNonRoot: false` outside of `overlays/kind-local/` to prevent accidental propagation.

tests/e2e/gateway-e2e-test.sh
```diff
@@ -127,6 +139,18 @@ _ensure_gateway_port_forward() {
   openshell sandbox list --gateway "${TENANT}" &>/dev/null
 }
 
+# Derives the sandbox name from a session ID, mirroring the Go
```
squizzi, line 142: Nice factoring. These helpers mirror the Go `openshell.SandboxName()` and `openshell.SandboxCRName(SandboxName(id))` functions — first 11 chars lowercased, then `default--` prefix. Having named functions eliminates the DRY violations across the script.

Important fix: the old code used `:0:40` (40 chars) while the Go function uses 11. This was the root cause of sandbox pod name mismatches.

tests/e2e/gateway-e2e-test.sh
```diff
@@ -1045,6 +1100,16 @@ LOCKED_AGENT_ID=$(echo "$AGENTS_RESP" \
   | jq -r '.items[] | select(.name == "network-test-locked-down") | .id' 2>/dev/null | head -1 || echo "")
 
 if [ -n "$LOCKED_AGENT_ID" ]; then
+  # The start handler returns an existing active session for the same agent
```
squizzi, line 1112: Good call. The start handler returns an existing active session for the same agent instead of creating a new one. After a `--skip-cleanup` run, the network policy test would reuse the old session with potentially stale network configuration. Deleting active sessions ensures a fresh sandbox is created with current policy settings.

tests/e2e/openshell-cli-e2e.sh
```diff
@@ -115,6 +116,73 @@ run_cmd() {
   echo ""
 }
 
+# Retry a command until an assertion on CMD_OUTPUT/CMD_RC passes.
+# Usage: eventually <max_attempts> <interval_secs> <assertion_fn> <cmd...>
```
squizzi, line 120: The `eventually` helper is a clean pattern for handling supervisor relay startup lag. The assertion function approach (inspect CMD_OUTPUT/CMD_RC) keeps the retry logic generic while making each call site declarative about what success looks like.

.github/workflows/test-local-dev.yml
```diff
@@ -60,6 +60,8 @@ jobs:
     env:
       KUBECTL_VERSION: v1.31.4
       KIND_VERSION: v0.31.0
+      IMAGE_TAG: ${{ inputs.image-tag || 'latest' }}
```
squizzi, line 63: Good catch — these inputs were defined in the workflow's `on.workflow_call.inputs` but never mapped to env vars, so the actual Job steps always used the `:latest` tag instead of the PR-built image. This is the kind of silent misconfiguration that makes CI green on `latest` but doesn't actually test the PR's images.

squizzi (commented):
🤖 **Amber Review — PR #461**

**Verdict: Approve** (posted as COMMENT — can't self-approve)
**Confidence: 90%** — All changes internally consistent, convention-compliant, no constitution violations.

---

**21 files, 557+/226-, 5 areas.** I read the full 66KB diff and traced the rh-trex-ai framework middleware chain to verify the auth sanitization is safe.

## Key Findings

**Security (Principle II — compliant):**
- `auth_header_sanitize.go`: Safe. Framework applies PreAuth middleware *inside* the JWT wrapper (`apiserver.go:39-59`), so JWT extracts the real token first and stores claims in context. Sanitize only redacts headers visible to router handlers.
- `staticTokenUsername`: Correct fix for gRPC RBAC auto-provisioning.
- `security-context-patch.yaml`: Correctly scoped to `overlays/kind-local/` only.

**Control Plane:**
- NetworkPolicy supports both legacy and new label schemes.
- DNS workaround comments document non-obvious race conditions.

**E2E Tests:**
- 10+ inline wait loops replaced with `wait_for`/`eventually` helpers.
- `sandbox_name()`/`sandbox_pod_name()` fix truncation from 40→11 chars (matching Go).
- Stale session cleanup prevents false positives from `--skip-cleanup` runs.

**CI:**
- `IMAGE_TAG`/`BUILT_IMAGES` were never wired — CI was testing `:latest` instead of PR images.

## Non-blocking Suggestions
1. CI lint rule to prevent `runAsNonRoot: false` outside `overlays/kind-local/`
2. `defaultMode` octal/decimal representation nit (see inline)

No blocking issues. See inline comments for detailed analysis on each file.

YiqinZhang (commented):
🤖 **Amber Re-Review #3 — PR #461: fix: get test-local-dev-simulation CI green**

Re-reviewing at head `7fd6795acadfa` — all previous findings are resolved, including the unary gRPC interceptor bug flagged in my last pass. Three new e2e commits look clean.

## Previous Findings — Final Status

- ✅ **Security context relaxation (Blocker):** `DEV-ONLY` comment added. Addressed.
- ✅ **Static token auto-provisioning (Critical):** Scope and RBAC implications documented. Addressed.
- ✅ **`verifyAndFixDNSConfig` exec race (Major):** Race safety note added. Addressed.
- ✅ **Unexplained RBAC rule counts (Major):** Counts traceable to auth middleware + static-service-token provisioning. Addressed.
- ✅ **Unary gRPC interceptor drops `CallerTypeService` (Critical — from Re-Review #2):** `ctx = WithCallerType(ctx, CallerTypeService)` is now present in the unary interceptor, matching the stream path. Addressed.

## New Commits Since Last Review (commits 9–11)

All three are `fix(e2e):` scoped — no production code changes:

- **`0e155d9`** — Aligns sandbox pod naming with the control plane convention (`default--session-<first11chars>`). Correct — matches `SandboxCRName(SandboxName(id))`. Adds `-c agent` to kubectl exec calls to avoid `Defaulted container` stderr noise. Clean.
- **`881903a`** — Lists failed test names in the summary and strips kubectl stderr noise via a helper. Improves CI debuggability. Clean.
- **`7fd6795`** — Cleans stale sessions before network-policy agent tests (prevents re-use of non-terminal sessions from prior runs with `--skip-cleanup`). Adds a shared `wait_for()` progress helper replacing silent polling loops. Clean.

## No New Findings

All changes in the new commits are scoped to e2e test reliability. No production code impact.

**Note:** PR is still in **Draft** state. No action required from me — flagging so the author can mark it ready when the CI stabilizes.

**Confidence:** [92%] High — all flagged issues resolved; new commits are clean and well-scoped.

---
*— Amber*

PR #466, opened 2026-07-31
fix(control-plane): transition to Failed phase on provisioning errors

rh-amarin:
## Summary
- Fixes sessions getting stuck in Pending state when provisioning fails
- `provisionAsync` now calls `updateSessionPhase(ctx, session, PhaseFailed)` on error
- Updates specs to document the phase=Failed requirement and async error-handling convention

## Test plan
- [ ] Verify failed provisioning transitions session to Failed phase
- [ ] Check that error is still logged with context
- [ ] Confirm no sessions remain stuck in Pending after errors

### DIFF
--- a/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
+++ b/components/ambient-control-plane/internal/reconciler/kube_reconciler.go
@@ -303,6 +303,7 @@ func (r *SimpleKubeReconciler) provisionAsync(session types.Session) {
 	defer cancel()
 	if err := r.provisionSession(ctx, session); err != nil {
 		r.logger.Error().Err(err).Str("session_id", session.ID).Msg("provisioning failed")
+		r.updateSessionPhase(ctx, session, PhaseFailed)
 	}
 }
 
--- a/skills/RECONCILE.md
+++ b/skills/RECONCILE.md
@@ -50,19 +50,19 @@ skills/
 
 ## Reconciliation State
 
-**Last analyzed**: 2026-07-18 (openshell-cli-e2e-test gap analysis)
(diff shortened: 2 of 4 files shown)


jsell-rh:
🤖 **Amber Analysis — PR #466: fix(control-plane): transition to Failed phase on provisioning errors**

Clean, targeted fix. `provisionAsync` now correctly calls `updateSessionPhase(ctx, session, PhaseFailed)` on error, preventing sessions from being stuck in `Pending`. The spec and convention updates accurately codify the new requirement.

**Root Cause:** `provisionAsync` only logged errors without updating session phase (`kube_reconciler.go:304-306`), leaving sessions indefinitely in `Pending` state.

**What I checked:**
- `updateSessionPhase` signature and behavior: correct. Handles idempotency, sets `completion_time`, uses `PhaseFailed` constant from `shared.go` ✓
- Context lifetime: `provisionAsync` uses a 5-min `context.WithTimeout` and passes it to `updateSessionPhase`. Phase updates are fast so timeout expiry before the update is unlikely, but worth monitoring ✓
- No `panic()`, no token leaks, no service account misuse ✓
- Constitution compliance: clean ✓

**Major — Missing test coverage:**
`provisionAsync`'s error path has no automated test coverage. No tests in `kube_reconciler_test.go` exercise `provisionAsync` or `provisionSession`. The PR description lists only manual test steps. I'd recommend adding a test that mocks `provisionSession` to return an error and asserts the session phase transitions to `Failed` — this is a critical path that should be regression-protected.

**Confidence:** [92%] High — the 1-line fix is exactly correct for the diagnosed bug. The missing test is the only gap.

*— Amber*

PR #369, opened 2026-07-15
feat(kind): self-contained vTeam catalog lab + fleet-skip provisioning

jeremyeder:
## Summary

Splits the vTeam lab work out on its own. Two changes, both about making the
vTeam catalog lab work against a clean cluster instead of relying on `make kind-up`
pre-provisioning the tenants:

- **Fleet-skip**: `kind-up` now provisions only the demo fleet (`tenant-a`,
  `tenant-b`). The fleet loop applies an overlay **only when** `examples/overlays/<ns>/`
  exists, and **fails loudly** (collects failures, `exit 1`) if any tenant apply
  fails — no silent partial provisioning. `OPENSHELL_TENANTS` default reduced to match.
- **Self-contained vTeam lab**: applying a catalog creates the `Project`, and the
  control plane provisions the backing namespace + gateway from that record — no
  `kubectl create namespace` needed. Rewrote `QUICKSTART.md`, `vteam-lab.md`, and the
  example READMEs accordingly, including correct provider-secret setup (vertex/github
  use a single `token` key; jira uses `JIRA_URL`/`JIRA_USERNAME`/`JIRA_API_TOKEN`).

## Why draft

Opened as a draft for review of the lab flow before marking ready.

## Scope

Makefile (2 fleet hunks), `scripts/setup-kind-openshell.sh`, and docs only. No Go,
no frontend, no API/K8s resource changes.

## Verification

- `make -n kind-status` parses; `make -n kind-up` matches `main` behavior (pre-existing
  prereq exit, unrelated to this change).
- QUICKSTART port lookup uses the existing `make kind-status` / `awk` approach (no
  dependency on unrelated tooling).

PR #467, opened 2026-07-31
fix(kind): fail loudly when fleet tenant apply fails in kind-up

jeremyeder:
Previously the fleet loop in `kind-up` silently moved on if `acpctl apply -k` returned non-zero or the overlay directory was absent. A broken tenant provisioning appeared as a successful `make kind-up`.

## Changes

- Skip tenants that have no `kustomization.yaml` (precise check, replaces directory existence check)
- Collect apply failures into `FLEET_FAILED` instead of swallowing them
- Exit 1 after the loop if any tenant apply failed

## Replaces

Closes #369 — that PR bundled this fix with doc and OPENSHELL_TENANTS changes. The OPENSHELL_TENANTS change is handled separately; the doc changes are no longer needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -991,26 +991,32 @@ kind-up: preflight-cluster build-cli ## Start kind cluster and deploy the platfo
 		sleep 2; \
 		TOKEN=$$(kubectl get secret test-user-token -n $(NAMESPACE) -o jsonpath='{.data.token}' 2>/dev/null | base64 -d 2>/dev/null); \
 		$$ACPCTL login --url "http://localhost:$${PF_PORT}" --token "$$TOKEN" >/dev/null 2>&1; \
+		FLEET_FAILED=""; \
 		for ns in $(OPENSHELL_TENANTS); do \
 			if echo " $(SKIP_TENANT_SETUP) " | grep -q " $$ns "; then \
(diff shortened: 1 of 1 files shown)


jsell-rh (approved):
🤖 **Amber Review — PR #467**

Clean, focused fix. The logic is sound: file-check over directory-check is the right improvement, failure accumulation works correctly in a Make multi-line shell block, and `exit 1` after the loop ensures `make kind-up` fails visibly when any tenant apply fails.

**One minor suggestion** (non-blocking): the new `[ -f "examples/overlays/$$ns/kustomization.yaml" ] || continue` silently skips tenants without an overlay. The old code said `"no overlay directory — skipping fleet"` which helped debug typos in `OPENSHELL_TENANTS`. Consider:
```makefile
[ -f "examples/overlays/$$ns/kustomization.yaml" ] || { echo "  $$ns: no kustomization.yaml, skipping"; continue; }; ```

Everything else looks good. Port-forward cleanup (`kill $$PF_PID`) correctly runs before the `exit 1` check, so no resource leak on failure.

**Confidence:** High — change is isolated to the fleet loop in `kind-up`, no API changes, no new dependencies.

*— Amber*

PR #471, opened 2026-07-31
spec(provider): add provider-scoped config

bsquizz:
## Summary

- Adds a `config` map to the Provider declaration schema so Vertex AI providers declare `VERTEX_AI_PROJECT_ID` and `VERTEX_AI_REGION` alongside their credential secret
- Removes the global fallback to control plane env vars (`ANTHROPIC_VERTEX_PROJECT_ID`, `CLOUD_ML_REGION`) — provider config is the provider's responsibility
- The `config` field is generic: any provider type (e.g., `google-cloud`, `aws-bedrock`) can use it for passthrough config to the OpenShell gateway

## Test plan

- [ ] Review spec for completeness against OpenShell's `Provider.config` proto field
- [ ] Validate that required keys table covers all enforced config (currently only `vertex`)
- [ ] Confirm migration steps are sufficient for existing deployments

### DIFF
--- a/specs/platform/provider-scoped-config.spec.md
+++ b/specs/platform/provider-scoped-config.spec.md
@@ -1,0 +1,173 @@
+# Provider-Scoped Configuration
+
+**Date:** 2026-07-31
+**Status:** Proposed
+**Related:** `specs/platform/agent-sandbox-config.spec.md` (Provider Declarations, Provider Schema), `specs/platform/control-plane.spec.md` (env var table), `components/ambient-control-plane/internal/openshell/provider_mapping.go` (ProviderConfig)
+
+---
+
+## Problem
+
+Vertex AI providers require three configuration values to function:
+
+| Value | Current Source | Where It Should Come From |
+|-------|---------------|---------------------------|
+| `GOOGLE_SERVICE_ACCOUNT_KEY` | Provider secret (`token` key) | Provider secret (no change) |
+| `VERTEX_AI_PROJECT_ID` | Control plane env var `ANTHROPIC_VERTEX_PROJECT_ID` | Provider declaration |
+| `VERTEX_AI_REGION` | Control plane env var `CLOUD_ML_REGION` | Provider declaration |
+
+Today, only the service account key is per-provider. The project ID and region are global — set once on the control plane deployment and shared by every Vertex provider in every tenant namespace. This creates two problems:
+
+1. **Multi-tenant inflexibility.** Different teams cannot target different GCP projects or regions. A platform with tenants in `us-central1` and `europe-west4` cannot serve both without running separate control plane instances.
+
+2. **Operator coupling.** Changing the GCP project or region requires a control plane redeployment (env var change + rollout), even though the values are logically per-provider configuration that tenant admins should own.
+
+The same pattern will recur for any future provider type that needs config beyond a single credential token (e.g., custom API base URLs, model overrides, region routing).
+
+---
+
+## Design
+
+### Add `config` field to the Provider declaration schema
+
+Extend the Provider YAML schema (defined in `agent-sandbox-config.spec.md`) with a `config` map:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `name` | string | yes | Unique provider name within the namespace. |
+| `type` | string | yes | ACP credential type (e.g., `vertex`, `github`, `anthropic`). |
+| `secret` | string | yes | Name of a Kubernetes Secret holding the credential. |
+| `config` | map[string]string | conditionally required | Provider-specific configuration key-value pairs. Passed to the OpenShell gateway as provider config. Required keys depend on provider type — see [Required Config Keys by Provider Type](#required-config-keys-by-provider-type). |
+
+#### Vertex provider example (before)
+
+```yaml
+kind: Provider
+name: vertex-prod
+type: vertex
+secret: vertex-sa-key
+```
+
+The secret contains only `token` (the SA key JSON). Project ID and region come from control plane env vars.
+
+#### Vertex provider example (after)
+
+```yaml
+kind: Provider
+name: vertex-prod
+type: vertex
+secret: vertex-sa-key
+config:
+  VERTEX_AI_PROJECT_ID: my-gcp-project
+  VERTEX_AI_REGION: us-central1
+```
+
+All three values are now declared together. The provider is fully self-describing.
+
+### Config resolution: declaration only, no global fallback
+
+The control plane SHALL read provider config exclusively from the provider declaration's `config` map. There is no fallback to control plane env vars.
+
+The global env vars `ANTHROPIC_VERTEX_PROJECT_ID` and `CLOUD_ML_REGION` SHALL be removed from the control plane config. The `VertexProjectID` and `VertexRegion` fields SHALL be removed from the control plane's config struct. Provider config is the provider's responsibility, not the control plane's.
+
+### Required Config Keys by Provider Type
+
+| Provider Type | Required Keys |
+|---------------|--------------|
+| `vertex` | `VERTEX_AI_PROJECT_ID`, `VERTEX_AI_REGION` |
+
+Provider types not listed here have no required config keys. The table will be extended as new provider types gain config requirements.
+
+### Validation
+
+The control plane SHALL validate provider config at reconcile time:
+
+- **Vertex providers:** Both `VERTEX_AI_PROJECT_ID` and `VERTEX_AI_REGION` MUST be present in the declaration's `config` map. If either is absent, the reconciler SHALL return an error and set the session status to failed with a message indicating the missing config key(s).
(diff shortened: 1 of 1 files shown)


specs/platform/provider-scoped-config.spec.md
```diff
@@ -0,0 +1,173 @@
...
+**File:** `components/ambient-control-plane/internal/reconciler/` (provider declaration struct)
+
+Add `Config map[string]string` to the provider declaration struct. Parse it from the YAML alongside `name`, `type`, and `secret`.
+
+### 2. Provider config resolution
+
+**File:** `components/ambient-control-plane/internal/openshell/provider_mapping.go`
+
+Replace `ProviderConfig` to read directly from the declaration config map. The function no longer takes global env-var-sourced parameters:
+
```
bsquizz, line 108: **Incomplete change surface — SDK + API server not addressed.**

The "provider declaration struct" referenced here is `types.Provider` in `components/ambient-sdk/go-sdk/types/provider.go` (line 13), which is **auto-generated from the OpenAPI spec** (`components/ambient-api-server/openapi/openapi.yaml`). It currently has no `Config` field.

Adding `Config` to providers requires:
1. Update the OpenAPI spec to add the `config` field to the Provider schema
2. Update the API server's provider handler/persistence (possible DB migration if provider config is a new column)
3. Regenerate the SDK types (`make generate` or equivalent)
4. *Then* the control plane reconciler gets the field for free via `provDecl.Config`

The spec should add a **step 0** before the current step 1 that covers the API server + SDK side. Without it, `provDecl.Config` won't exist at compile time.

(The good news: the OpenShell proto already has `Config map[string]string` on the `Provider` message at `datamodel.pb.go:312`, so the gateway side is ready.)

specs/platform/provider-scoped-config.spec.md
```diff
@@ -0,0 +1,173 @@
...
+**File:** `components/ambient-control-plane/internal/reconciler/kube_reconciler.go` (~line 1309)
+
+Pass the declaration config directly:
+
+```go
+// Before:
+Config: openshell.ProviderConfig(provType, r.cfg.VertexProjectID, r.cfg.VertexRegion),
+
+// After:
+Config: openshell.ProviderConfig(provType, provDecl.Config),
```
bsquizz, line 140: **Missing two additional call sites in `kube_reconciler.go`.**

The spec correctly identifies the `ProviderConfig` call at line ~1309, but there are two more active code paths that inject `VertexProjectID`/`VertexRegion` directly into container env vars:

1. **Line ~1733-1734** (`buildSandboxEnv`):
   ```go
   env["ANTHROPIC_VERTEX_PROJECT_ID"] = r.cfg.VertexProjectID
   env["CLOUD_ML_REGION"] = r.cfg.VertexRegion
   ```
   This is currently behind `if false && r.cfg.VertexEnabled` (dead code), but should be cleaned up as part of this change.

2. **Line ~2823-2824** (container spec builder):
   ```go
   envVar("ANTHROPIC_VERTEX_PROJECT_ID", r.cfg.VertexProjectID),
   envVar("CLOUD_ML_REGION", r.cfg.VertexRegion),
   ```
   This is **live code** gated by `r.cfg.VertexEnabled`. It injects these as container env vars directly — a separate path from the OpenShell provider config flow at line 1309.

Also missing: the `KubeReconcilerConfig` struct at lines 88-89 has its own `VertexProjectID` and `VertexRegion` fields (separate from `config.go`'s struct) that also need removal.

The spec should enumerate all four sites to avoid leaving stale references behind.

specs/platform/provider-scoped-config.spec.md
```diff
@@ -0,0 +1,173 @@
...
+Add `config` with `VERTEX_AI_PROJECT_ID` and `VERTEX_AI_REGION` to demonstrate the full declaration.
+
+### 7. CLI (openshell provider create)
+
+The `--config` flags already exist on the CLI (`openshell provider create --config KEY=VALUE`). No CLI changes are needed — the config values just need to be persisted into the provider declaration ConfigMap rather than being ignored or only passed ephemerally.
+
+---
+
+## Migration
+
```
bsquizz, line 160: **Migration step missing: implicit `CLOUD_ML_REGION` default of `"global"` is being removed.**

The current `config.go` (line 103) sets:
```go
VertexRegion: envOrDefault("CLOUD_ML_REGION", "global"),
```

Any deployment that does not explicitly set `CLOUD_ML_REGION` today gets `"global"` as the default silently. After this change, those deployments must add `VERTEX_AI_REGION: global` to their provider declaration's `config` map — or reconciliation will fail with a missing-key error.

The migration section should call out this implicit default so operators don't get surprised. Something like:

> **Note:** If your deployment does not currently set `CLOUD_ML_REGION`, the control plane was defaulting to `"global"`. You must now explicitly set `VERTEX_AI_REGION: global` in the provider config.

specs/platform/provider-scoped-config.spec.md
```diff
@@ -0,0 +1,173 @@
...
+**File:** `components/ambient-control-plane/internal/config/config.go`
+
+Remove `VertexProjectID` and `VertexRegion` fields from the config struct. Remove the `ANTHROPIC_VERTEX_PROJECT_ID` and `CLOUD_ML_REGION` env var reads.
+
+**Files:** `components/manifests/base/ambient-control-plane-service.yml`, overlay patches, templates
+
+Remove `ANTHROPIC_VERTEX_PROJECT_ID` and `CLOUD_ML_REGION` env var entries and their `operator-config` ConfigMap references.
+
+### 4. Reconciler call site
+
```
bsquizz, line 130: **Relationship between `VertexEnabled`/`VertexCredentialsPath` and this change is unclear.**

The spec removes `VertexProjectID` and `VertexRegion` from the config struct, but `VertexEnabled` (line 31) and `VertexCredentialsPath` (line 34) in `config.go` survive. These are used at `kube_reconciler.go:2821` to gate an env-var injection code path that sets `ANTHROPIC_VERTEX_PROJECT_ID`, `CLOUD_ML_REGION`, and `GOOGLE_APPLICATION_CREDENTIALS` directly on containers.

Should this legacy Vertex env-var injection path also be removed (since OpenShell provider config now handles it), or does it serve a different purpose? If it stays, it'll conflict with the provider-scoped config — the container would get two sources of truth for the same values.

I'd suggest the spec explicitly addresses the fate of `VertexEnabled` and the associated env-var injection blocks, even if the decision is "leave them for now, clean up later."

specs/platform/provider-scoped-config.spec.md
```diff
@@ -0,0 +1,173 @@
...
+---
+
+## Changes Required
+
+### 1. Provider declaration parsing
+
+**File:** `components/ambient-control-plane/internal/reconciler/` (provider declaration struct)
+
+Add `Config map[string]string` to the provider declaration struct. Parse it from the YAML alongside `name`, `type`, and `secret`.
```
bsquizz, line 101: **Clarify where `config` is persisted — "ConfigMap" vs API-server resource.**

This says config "lives in the ConfigMap alongside the provider declaration." But in the codebase, providers are API-server-managed resources fetched via `sdk.Providers().List()` (see `kube_reconciler.go:1265`), backed by PostgreSQL — not ConfigMaps.

The Provider SDK type (`components/ambient-sdk/go-sdk/types/provider.go:13`) is auto-generated from the OpenAPI spec, and the reconciler accesses `provDecl.Secret`, `provDecl.Type`, etc. from these API objects.

If `config` is a new field on the API server's Provider resource, the spec should say so — and note that it requires an OpenAPI schema + possible database migration. If it's a separate ConfigMap sidecar, the spec should describe how the reconciler joins the two.

bsquizz (commented):
🤖 **Amber Analysis**

**Overall:** Strong spec — the problem statement is clear, the design is sound, and the direction is correct. The OpenShell proto already has `Config map[string]string` on the `Provider` message, confirming the gateway is ready for this. However, the "Changes Required" section has gaps that would block implementation.

**Confidence:** 90% — verified all claims against the current codebase.

**Key findings (5 inline comments):**

1. **Missing API-server + SDK changes.** The spec jumps straight to control plane code, but providers are API-server resources (PostgreSQL-backed, fetched via SDK). Adding `Config` requires OpenAPI spec update → SDK regen → possible DB migration *before* the control plane can use `provDecl.Config`. This is the biggest gap.

2. **Two additional `VertexProjectID`/`VertexRegion` call sites not addressed.** The spec covers line ~1309 but misses the live code path at ~2823 and the dead-but-present path at ~1733. Also misses the `KubeReconcilerConfig` struct fields at lines 88-89.

3. **Implicit `CLOUD_ML_REGION` default of `"global"` lost silently.** Migration section should warn operators who relied on the default.

4. **`VertexEnabled` and `VertexCredentialsPath` fate unclear.** These survive the change but gate a code path that injects the same env vars being removed — potential dual source of truth.

5. **"ConfigMap" storage claim conflicts with actual data flow.** Providers are API-server resources, not ConfigMaps.

None of these are design objections — the approach is right. These are implementation surface gaps that would cause confusion or compilation failures during the build phase.

